PR #14, opened 2018-05-21
Makes it so you can use multiple parameters to microk8s.enable and disable

marcoceppi:
```
microk8s.enable dns dashboard
microk8s.disable dns dashboard
```


ktsakalozos:
LGTM +1

PR #15, opened 2018-05-21
Remove exec from actions so actions can be queued

ktsakalozos:
exec will replace the executing process not leaving the option for another process to spawn. This will not work if we want to queue actions.


ktsakalozos:
Self-merging this to fix my previous merge.

PR #19, opened 2018-05-21
Add config action to dump client config

johnsca:
```
microk8s.config [-l]
```

This converts the endpoint to use the IP of the default network interface by default, with the option to drop back to using the loopback address.  Otherwise, this is the same as (but shorter than):

```
microk8s.kubectl config view --raw
```


ktsakalozos:
Thanks @johnsca LGTM +1

PR #24, opened 2018-05-22
Create an alias for microk8s.kubectl

### LINKED ISSUE #20: Add microk8s.kubectl as an alias for kubectl
This would make it so if `kubectl` wasn't installed `microk8s.kubectl` would become become linked to `/snap/bin/kubectl`

This would really help streamline adoption

marcoceppi:
This makes the UX easier for constant users microk8s it will allow users to, if they opt, alias the command

```
snap alias microk8s kubectl
```

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -35,6 +35,7 @@ apps:
     daemon: simple
   kubectl:
     command: kubectl --kubeconfig=$SNAP/client.config
+    aliases: [kubectl]
   enable:
     command: microk8s-enable.wrapper
   disable:


johnsca:
I don't think this is necessary.  You can already create a manual alias with the current snap:

```
ubuntu@conjure-up:~$ kubectl get nodes

Command 'kubectl' not found, but can be installed with:

sudo snap install kubectl

ubuntu@conjure-up:~$ sudo snap install microk8s --beta --classic
microk8s (beta) v1.10.3 from 'kjackal' installed
ubuntu@conjure-up:~$ sudo snap alias microk8s.kubectl kubectl
Added:
  - microk8s.kubectl as kubectl
ubuntu@conjure-up:~$ kubectl get nodes
NAME         STATUS    ROLES     AGE       VERSION
conjure-up   Ready     <none>    3m        v1.10.3
```

johnsca:
I do think it might be worth considering whether we want to post on the forum to request an auto-alias be created, though we should ask about what the user experience would be if they already have kubectl on their system from something else, whether that would block the install of microk8s (like having an alias for microk8s.kubectl does for `snap install kubectl`) or if it would just skip the alias.

johnsca:
I see I missed the reply on the issue, as well as the doc PR.  So it seems like this is moot, so I'm going to close it.  Feel free to re-open if needed.

Issue #33, opened 2018-05-30
microk8s.disable doesn't shut down k8s processes - v1.10.3


bkcsfi:
I would like to be able to completely shutdown microk8s without uninstalling it, but microk8s.disable leaves kubelet and other processes running. 

If I kill them, they restart

ubuntu 18.04 LTS

```
Name      Version    Rev   Tracking  Developer  Notes
core      16-2.32.8  4650  stable    canonical  core
microk8s  v1.10.3    55    beta      canonical  classic

# microk8s.disable
# sleep 20
# ps x | grep microk | colrm 80

 2046 ?        Ssl  258:16 /snap/microk8s/55/kube-apiserver --v=4 --insecure-bi
 2053 ?        Ssl  110:32 /snap/microk8s/55/etcd --data-dir=/var/snap/microk8s
 2063 ?        Ssl   29:56 /snap/microk8s/55/usr/bin/dockerd -H unix:///var/sna
 2560 ?        Ssl   10:44 containerd -l unix:///var/snap/microk8s/common/var/r
25494 ?        Ssl    0:16 /snap/microk8s/55/kubelet --kubeconfig=/snap/microk8

```

marcoceppi:
```
microk8s.disable
```

Is for disabling addons like `dns`, or `dashboard`

To disable `microk8s` you'll need to use `snap` command to do that

```
snap disable microk8s
```

Then to enable

```
snap enable microk8s
```

marcoceppi:
I'll update the readme to reflect this. We'll also need to create something like `microk8s.io/docs` to detail these things further

PR #34, opened 2018-05-30
Patch 1

bkcsfi:
Refs #33


bkcsfi:
canceling this MR, created new one from update-readme branch instead of master

PR #35, opened 2018-05-30
Update README.md

bkcsfi:
Refs #33 - this mr based on update-readme branch instead of master


marcoceppi:
Beat me to it! Thanks for the quick turn around on this patch

PR #39, opened 2018-06-01
Update admission-plugins based on upstream recommendations for 1.10

marcoceppi:
Based on https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#is-there-a-recommended-set-of-admission-controllers-to-use


ktsakalozos:
LGTM +1 although some of them might not make sense in a single user setup.

PR #42, opened 2018-06-05
Adding addon tests and testing in trusty/travis

ktsakalozos:
A basic set of test, one for each addon we have. We can call the "validators" on a running microk8s by either enabling/disabling the addons or expecting the addons to be already there. The later case will be used when working on testing the upgrade paths.

A few fixes that were made while testing on travis (trusty).
-  util-linux are needed by kubenet
- sed needed to be installed
- gawk should be the one we package. 

Also adding a reset command to clean pods. Should be used before removing snap. Fixes: https://github.com/juju-solutions/microk8s/issues/23

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ script:
   - sudo apt update
   - sudo apt install --yes snapd
   - sudo snap install *.snap --classic --dangerous
-  - sleep 10
   - sudo pip install -U pytest
   - pwd
   - ps -ef
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,11 @@ script:
   - sudo apt update
   - sudo apt install --yes snapd
   - sudo snap install *.snap --classic --dangerous
+  - sleep 10
+  - sudo pip install -U pytest
   - pwd
+  - ps -ef
   - ls -l
   - ./tests/smoke-test.sh
+  - (cd tests; pytest -s test-addons.py)
   - sudo snap remove microk8s
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,5 @@ script:
   - ls -l
   - ./tests/smoke-test.sh
   - (cd tests; pytest -s test-addons.py)
+  - sudo microk8s.reset
   - sudo snap remove microk8s
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ snap enable microk8s
 
 ### Removing microk8s
 
-To remove microk8s we recommend you first disable all addons and stop all hosted pods. microk8s.reset can assist with that.
+Before removing microk8s, use `microk8s.reset` to stop all running pods.
 
 ```
 microk8s.reset
--- a/README.md
+++ b/README.md
@@ -80,6 +80,14 @@ microk8s can be restarted later with the snap command
 snap enable microk8s
 ```
 
+### Removing microk8s
+
+To remove microk8s we recommend you first disable all addons and stop all hosted pods. microk8s.reset can assist with that.
+
+```
+microk8s.reset
+snap remove microk8s
+```
 
 ### Configuring microk8s services
 The following systemd services will be running in your system:
--- a/microk8s-resources/actions/disable.dns.sh
+++ b/microk8s-resources/actions/disable.dns.sh
@@ -2,17 +2,18 @@
 
 set -e
 
-## remove an option inside the config file.
 skip_opt_in_config() {
+    # remove an option inside the config file.
+    # argument $1 is the option to be removed
+    # argument $2 is the configuration file under $SNAP_DATA/args
     opt="--$1"
     config_file="$SNAP_DATA/args/$2"
-    sudo AWKLIBPATH="${SNAP}/usr/lib/x86_64-linux-gnu/gawk" AWKPATH="${SNAP}/usr/share/awk/" \
-      "${SNAP}/usr/bin/gawk" -i inplace '!/^'"$opt"'/ {print $0}' "${config_file}"
+    sudo "${SNAP}/bin/sed" -i '/'"$opt"'/d' "${config_file}"
 }
 
 echo "Disabling DNS"
 echo "Reconfiguring kubelet"
-#TODO(kjackal): do not hardcode the info below. Get it from the yaml
+
 skip_opt_in_config "cluster-domain" kubelet
 skip_opt_in_config "cluster-dns" kubelet
 sudo systemctl restart snap.${SNAP_NAME}.daemon-kubelet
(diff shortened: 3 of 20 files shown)


.travis.yml
```diff
@@ -9,8 +9,12 @@ script:
   - sudo apt update
   - sudo apt install --yes snapd
   - sudo snap install *.snap --classic --dangerous
+  - sleep 10
```
tvansteenburgh: Why?

README.md
```diff
@@ -80,6 +80,14 @@ microk8s can be restarted later with the snap command
 snap enable microk8s
 ```
 
+### Removing microk8s
+
+To remove microk8s we recommend you first disable all addons and stop all hosted pods. microk8s.reset can assist with that.
```
tvansteenburgh: "Before removing microk8s, use `microk8s.reset` to stop all running pods."

I understand why we need this right now, but it would be much nicer if we could get a pre-remove hook for snaps and then put this logic there, so the user doesn't have to know or think about it.

microk8s-resources/actions/disable.dns.sh
```diff
@@ -6,7 +6,8 @@ set -e
 skip_opt_in_config() {
```
tvansteenburgh: Please add a brief comment explaining what this function does.

microk8s-resources/actions/disable.dns.sh
```diff
@@ -6,7 +6,8 @@ set -e
...
     opt="--$1"
     config_file="$SNAP_DATA/args/$2"
-    sudo gawk -i inplace '!/^'"$opt"'/ {print $0}' "${config_file}"
+    sudo AWKLIBPATH="${SNAP}/usr/lib/x86_64-linux-gnu/gawk" AWKPATH="${SNAP}/usr/share/awk/" \
```
hyperbolic2346: Heard a rumor that sed can eat the newline too.

snap/hooks/remove
```diff
@@ -9,7 +12,7 @@ mount | grep ${SNAP_COMMON}/var/*/docker | cut -d ' ' -f 3 | xargs umount
 #TODO(kjackal): Make sure this works everywhere we want
 if [ -f /etc/apparmor.d/docker ]; then
   echo "Updating docker-default profile"
-  gawk -i inplace '!/^  signal \(receive\) peer=snap.microk8s.daemon-docker,$/ {print}' /etc/apparmor.d/docker
+  "${SNAP}/usr/bin/gawk" -i inplace '!/^  signal \(receive\) peer=snap.microk8s.daemon-docker,$/ {print}' /etc/apparmor.d/docker
```
hyperbolic2346: investigate replacing with sed

tests/test-addons.py
```diff
@@ -0,0 +1,53 @@
...
+
+class TestAddons(object):
+
+    def test_dns(self):
+        """
+        Sets up DNS addon and validates it works.
+
+        """
+        print("Enabling DNS")
+        microk8s_enable("dns")
```
hyperbolic2346: Is there a way to not disrupt the cluster after running tests? If I have a microk8s with dns and dashboard enabled and then run tests I will have both off at the end.

```
original_dns = microk8s_is_enabled("dns")
microk8s_enable("dns")
...
if original_dns == false:
  microk8s_disable("dns")
```

tests/utils.py
```diff
@@ -0,0 +1,105 @@
...
+def microk8s_enable(addon):
+    """
+    Disable an addon
+
+    Args:
+        addon: name of the addon
+
+    """
+    cmd = '/snap/bin/microk8s.enable {}'.format(addon)
+    time.sleep(10)
```
hyperbolic2346: why are we sleeping here? We haven't run the command yet.

tests/validators.py
```diff
@@ -0,0 +1,57 @@
...
+    kubectl("delete -f {}".format(manifest))
+
+
+def validate_dashboard():
+    """
+    Validate the dashboard addon by looking at the grafana URL.
+    """
+    wait_for_pod_state("", "kube-system", "running", label="k8s-app=influxGrafana")
+    cluster_info = kubectl("cluster-info")
+    grafana_url = "http://127.0.0.1:8080/api/v1/namespaces/kube-system/services/monitoring-grafana/proxy"
```
hyperbolic2346: If we change the API server's port this 8080 will need to change.

README.md
ktsakalozos: We brought this to the snapcraft forum: https://forum.snapcraft.io/t/cleanup-before-services-go-down/5802

.travis.yml
ktsakalozos: Removed. Waiting should be done inside the smoke-test script.

tests/test-addons.py
ktsakalozos: I think this should be addressed when we offer the microk8s.status https://github.com/juju-solutions/microk8s/issues/37 . These tests are to be run in a "controlled environment". I would rather know if dns did not come up instead of skipping the test.

tests/utils.py
ktsakalozos: Removed!

tests/utils.py
```diff
@@ -0,0 +1,105 @@
...
+def microk8s_disable(addon):
+    """
+    Enable an addon
+
+    Args:
+        addon: name of the addon
+
+    """
+    cmd = '/snap/bin/microk8s.disable {}'.format(addon)
+    time.sleep(10)
```
tvansteenburgh: Remove?

tests/validators.py
```diff
@@ -0,0 +1,57 @@
...
+    assert grafana_url in cluster_info
+    attempt = 5
+    while attempt >= 0:
+        resp = requests.get(grafana_url)
+        if resp.status_code == 200:
+            break
+        time.sleep(20)
```
tvansteenburgh: I understand sleeping here, but shouldn't it be more like 1 second instead of 20?!

snap/hooks/remove
ktsakalozos: Done

microk8s-resources/actions/disable.dns.sh
ktsakalozos: Fixed :)

tests/utils.py
ktsakalozos: Yes! This too

tests/validators.py
ktsakalozos: Changes to 50 attempts, one every 2 secs

README.md
tvansteenburgh: Recommend changing this sentence to "Before removing microk8s, use `microk8s.reset` to stop all running pods."

microk8s-resources/actions/disable.dns.sh
ktsakalozos: Done

ktsakalozos:
Merging this to get the ingress tests going.

PR #45, opened 2018-06-07
Remove the etcd socket file when etcd restart

ktsakalozos:
Should fix: https://github.com/juju-solutions/microk8s/issues/44


ktsakalozos:
Merging this. @wallyworld the fix should be on edge in a few hours.

PR #49, opened 2018-06-13
Replace nfs-provisioner with hostpath-provisioner

ktsakalozos:
nfs-proviosner needs nfs-coomon which is rather hard to put in a snap package. Instead we moved to hostpath-provioner. In this PR we also improve tests on storage.

Should fix: https://github.com/juju-solutions/microk8s/issues/47

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 /snap/*
 !/snap/hooks/
 *.snap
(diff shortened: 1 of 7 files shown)


ktsakalozos:
Thanks for the reviews, merging.

PR #51, opened 2018-06-23
Upgrade test linked to travis

ktsakalozos:
This PR adds a test to verify that the upgrade of a snap is not broken.
We first deploy the snap from UPGRADE_MICROK8S_FROM channel. Then we go through the validators to see which ones apply for that charm. We then upgrade to UPGRADE_MICROK8S_TO and we need the same validators to be passing.

The upgrade path that is tested is the one on the edge channel (see travis.yaml) and eventually we will need to run the same upgrade path validation on the cross channel promotion (ie release a revision tested on edge to beta).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,4 @@ script:
   - (cd tests; pytest -s test-addons.py)
   - sudo microk8s.reset
   - sudo snap remove microk8s
-  - UPGRADE_MICROK8S_FROM=edge UPGRADE_MICROK8S_TO=`pwd`/*.snap pytest -s ./tests/test-upgrade.py
+  - UPGRADE_MICROK8S_FROM=edge UPGRADE_MICROK8S_TO=`pwd`/`ls microk8s*.snap` pytest -s ./tests/test-upgrade.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,4 @@ script:
   - (cd tests; pytest -s test-addons.py)
   - sudo microk8s.reset
   - sudo snap remove microk8s
+  - UPGRADE_MICROK8S_FROM=edge UPGRADE_MICROK8S_TO=`pwd`/*.snap pytest -s ./tests/test-upgrade.py
--- a/tests/test-upgrade.py
+++ b/tests/test-upgrade.py
@@ -1,0 +1,82 @@
+import os
+import time
+from validators import validate_dns, validate_dashboard, validate_storage, validate_ingress
+from subprocess import check_call
+from utils import microk8s_enable, wait_for_pod_state, microk8s_disable, wait_for_installation
+
+upgrade_from = os.environ.get('UPGRADE_MICROK8S_FROM', 'beta')
+# Have UPGRADE_MICROK8S_TO point to a file to upgrade to that file
+upgrade_to = os.environ.get('UPGRADE_MICROK8S_TO', 'edge')
+
+
+class TestUpgrade(object):
+    """
+    Validates a microk8s upgrade path
+    """
+
+    def test_upgrade(self):
+        """
+        Deploy, probe, upgrade, validate nothing broke.
+
+        """
+        cmd = "sudo snap install microk8s --classic --{}".format(upgrade_from).split()
+        check_call(cmd)
+        wait_for_installation()
+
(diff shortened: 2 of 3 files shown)


tests/test-upgrade.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Validates a microk8s upgrade path
+    """
+
+    def test_upgrade(self):
+        """
+        Deploy, probe, upgrade, validate nothing broke.
+
+        """
+        cmd = "sudo snap install microk8s --classic --{}".format(upgrade_from).split()
```
hyperbolic2346: I find it useful to output information about the upgrade_from and upgrade_to here so you know based on the log what it tried to do. Maybe:
```
Testing upgrade from candidate to edge
```

.travis.yml
```diff
@@ -17,3 +17,4 @@ script:
   - (cd tests; pytest -s test-addons.py)
   - sudo microk8s.reset
   - sudo snap remove microk8s
+  - UPGRADE_MICROK8S_FROM=edge UPGRADE_MICROK8S_TO=`pwd`/*.snap pytest -s ./tests/test-upgrade.py
```
hyperbolic2346: Are we really starting with edge and upgrading to something else?

ktsakalozos: Yes! Here is why. As soon as we merge a change the builders will trigger and push a new snap in to edge. So we need to make sure the snap we created (`UPGRADE_MICROK8S_TO=pwd/*.snap`) is backwards compatible and will not break what is already available on edge (`UPGRADE_MICROK8S_FROM=edge`).

hyperbolic2346: ok, so edge to freshly built not in the store snap. Makes sense to me, but the tests don't like *.snap, I think you need to get that resolved to filenames or something.

tests/test-upgrade.py
```diff
@@ -0,0 +1,76 @@
...
+        wait_for_installation()
+
+        # Run through the validators and
+        # select those that were valid for the original snap
+        test_matrix = {}
+        try:
+            microk8s_enable("dns")
+            wait_for_pod_state("", "kube-system", "running", label="k8s-app=kube-dns")
+            validate_dns()
+            test_matrix['dns'] = validate_dns
```
johnsca: It seems like these are running the validate functions twice, once here before adding them to the matrix and then once below when looping through the matrix.  I think the first line (`validate_dns()`) should be removed.

tests/test-upgrade.py
```diff
@@ -0,0 +1,76 @@
...
+        except:
+            print('Will not test dns')
```
johnsca: If `microk8s_enable('dns')` fails, that seems like it should be considered a failure and not a skip.   If the intent is to allow for a particular add-on being removed in the future, maybe the condition for doing the check should instead use the new `--help` option and check for the addon in the list?

tests/test-upgrade.py
ktsakalozos: The first run of the validation functions is to select those valid for the original snap. If they are not valid for the original snap there is no way to break them during an upgrade so there is no point in testing them after the upgrade.

Let me give you an example. The storage addon is available in edge but not in beta. When we decide to release a revision to beta all beta deployments will upgrade_from beta to the revision that is currently in edge. What we want to do is to snap install the beta snap, run through the validations and see which ones apply, upgrade_to edge and see if those valid validators still succeed. In the case of 'storage' the validation will fail in the beta snap so the upgrade will not break it either way and after upgrading to edge there is no point in testing the storage validation.

tests/test-upgrade.py
ktsakalozos: As mentioned in the previous comment we skip the test and we do not add it in the `test_matrix` because we are not interested in testing it after the upgrade.

Something else, the validation tests do not have to map to the addons we have. We may have a validation that deploys microbot and checks it is still up after an upgrade. I am mostly concerned with releasing to beta. Edge backwards compatibility is also nice, but it is beta where our users live at the moment.

Also all validations are already checked against the current snap here: https://github.com/juju-solutions/microk8s/blob/master/.travis.yml#L17

ktsakalozos:
Thank you for the reviews.

PR #55, opened 2018-06-26
Rename context from juju-context to microk8s

johnsca:
Having the default context name be `juju-context` is confusing. Renaming it to `microk8s` makes it easier to integrate with other contexts or refer to when using things like ksonnet.


johnsca:
+1 to blocking until we have a test that covers the upgrade.

ktsakalozos:
Rebasing to run the upgrade test

ktsakalozos:
I feel much better now.

LGTM +1

PR #71, opened 2018-07-18
A private registry addon

ktsakalozos:
A registry addon and exposing of docker namespaced (microk8s.docker).

The registry is insecure listening on port 32000 of the localhost.


README.md
```diff
@@ -66,6 +66,7 @@ You can find the addon manifests and/or scripts under `${SNAP}/actions/`, with `
 - **storage**: Create a default storage class. This storage class makes use of the hostpath-provisioner pointing to a directory on the host. Persistent volumes are created under `${SNAP_COMMON}/default-storage`. Upon disabling this addon you will be asked if you want to delete the persistent volumes created.
 - **ingress**: Create an ingress controller.
 - **gpu**: Expose GPU(s) to microk8s by enabling the nvidia-docker runtime and nvidia-device-plugin-daemonset. Requires NVIDIA drivers to already be installed on the host system.
+- **registry**: Deploy a docker private registry with a cluster IP of `10.152.183.254` on port `5000`. If you intend to use this private registry we recommend you enable it as soon as posible so you do not risk a cluster IP conflict. The storage addon will be enabled as part of this addon.
```
tvansteenburgh: This seems kind of lame. Why don't we instead make it a NodePort service and port 5000? Then it can be accessed at `localhost:5000`, which is easier to remember than an IP, and it avoids the possibility of the IP being taken already.

I would also like to include a short example here of how to put an image in the local registry, and how to reference it in a pod spec (i.e. the docker pull, docker tag, docker push steps).

ktsakalozos:
Thanks. Merging after the rebase.

PR #79, opened 2018-07-25
Document microk8s restart when master IP changes

ktsakalozos:
Based on https://github.com/kubernetes/kubeadm/issues/338 and
https://github.com/kubernetes/kubernetes/issues/34043 there is no easy way to update the IP of the master. Lets propose a microk8s restart for now.


ktsakalozos:
Thank you.

PR #80, opened 2018-07-25
Fix travis flakiness with more frequent cluster resets

ktsakalozos:
Our tests enable and disable dns multiple times. For some reason k8s freaks out and keeps a "Terminating" dns-pod entry for too long. Our tests wait for that dns pod to be deleted and travis sees no output for 10 minutes and kills the tests.

With this PR we make sure that at the end of each test we do a proper job in cleaning up all pods so that the next test will spawn dns-pod without k8s freaking out.


ktsakalozos:
Thanks!

PR #82, opened 2018-08-01
Add GPU addon

ktsakalozos:
Do `microk8s.enable gpu` on a host with NVIDIA loaded.

The nvidia docker container will be enabled and the nvidia daemon set will come up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ You can find the addon manifests and/or scripts under `${SNAP}/actions/`, with `
 - **dashboard**: Deploy kubernetes dashboard as well as grafana and influxdb. To access grafana point your browser to the url reported by `microk8s.kubectl cluster-info`.
 - **storage**: Create a default storage class. This storage class makes use of the hostpath-provisioner pointing to a directory on the host. Persistent volumes are created under `${SNAP_COMMON}/default-storage`. Upon disabling this addon you will be asked if you want to delete the persistent volumes created.
 - **ingress**: Create an ingress controller.
-- **gpu**: Deploy the NVIDIA docker as well as the daemonset plugin that would allow you to access your GPU from within kubernetes. This addon requires you to have a) NVIDIA drivers installed and loaded on your host and b) kube DNS enabled.
+- **gpu**: Expose GPU(s) to microk8s by enabling the nvidia-docker runtime and nvidia-device-plugin-daemonset. Requires NVIDIA drivers to already be installed on the host system.
 
 ### Stopping and Restarting microk8s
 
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -65,6 +65,7 @@ You can find the addon manifests and/or scripts under `${SNAP}/actions/`, with `
 - **dashboard**: Deploy kubernetes dashboard as well as grafana and influxdb. To access grafana point your browser to the url reported by `microk8s.kubectl cluster-info`.
 - **storage**: Create a default storage class. This storage class makes use of the hostpath-provisioner pointing to a directory on the host. Persistent volumes are created under `${SNAP_COMMON}/default-storage`. Upon disabling this addon you will be asked if you want to delete the persistent volumes created.
 - **ingress**: Create an ingress controller.
+- **gpu**: Deploy the NVIDIA docker as well as the daemonset plugin that would allow you to access your GPU from within kubernetes. This addon requires you to have a) NVIDIA drivers installed and loaded on your host and b) kube DNS enabled.
```
tvansteenburgh: "Expose GPU(s) to microk8s by enabling the nvidia-docker runtime and nvidia-device-plugin-daemonset. Requires NVIDIA drivers to already be installed on the host system."

(It would be nice if we could automatically enable dns instead of requiring the user to do that first.)

tests/test-addons.py
```diff
@@ -69,3 +75,24 @@ def test_ingress(self):
...
+        try:
+            print("Enabling gpu")
+            gpu_enable_outcome = microk8s_enable("gpu")
+        except CalledProcessError:
+            # Failed to enable gpu. Skip the test.
+            print("Disabling DNS")
+            microk8s_disable("dns")
+            return
+        validate_gpu()
+        print("Disbale gpu")
```
tvansteenburgh: Disable

microk8s-resources/actions/enable.gpu.sh
```diff
@@ -13,12 +13,11 @@ else
   exit 1
 fi
 
-echo "Enabling DNS"
-/snap/bin/microk8s.enable dns
-
 refresh_opt_in_config "default-runtime" "nvidia" dockerd
 sudo systemctl restart snap.${SNAP_NAME}.daemon-docker
-sleep 5
+sleep 10
```
tvansteenburgh: Could we instead poll the status of the daemon until it's started, or we hit a reasonable timeout? Just setting it to 10s seems bad.

ktsakalozos:
Thanks @tvansteenburgh merging this.

PR #83, opened 2018-08-03
Istio addon

ktsakalozos:
Grab the release of Istio indicated by the $ISTIO_VERSION variable.

Introduce the istioctl command.


build-scripts/fetch-binaries.sh
```diff
@@ -28,4 +28,16 @@ echo $KUBE_VERSION > $KUBE_SNAP_BINS/version
   mkdir -p cni
   curl -LO https://github.com/containernetworking/plugins/releases/download/${CNI_VERSION}/cni-plugins-$KUBE_ARCH-${CNI_VERSION}.tgz
   tar -zxvf cni-plugins-$KUBE_ARCH-${CNI_VERSION}.tgz -C cni
+
+  ISTIO_ERSION=$(echo $ISTIO_VERSION | sed 's/v//g')
```
battlemidget: ISTIO_ERSION, is that a typo?

ktsakalozos: Nope :) Version is "v1.0.0" but we need only the "1.0.0" for the url below. So that is the ISTIO_VERSION without the V

marcoceppi: I think Adam means the variable name `ISTIO_ERSION`

tests/test-live-addons.py
```diff
@@ -37,3 +43,10 @@ def test_ingress(self):
 
         """
         validate_ingress()
+
+    def test_istio(self):
+        """
+        Validates storage works.
```
tvansteenburgh: Validate Istio works.

README.md
```diff
@@ -66,6 +66,7 @@ You can find the addon manifests and/or scripts under `${SNAP}/actions/`, with `
 - **storage**: Create a default storage class. This storage class makes use of the hostpath-provisioner pointing to a directory on the host. Persistent volumes are created under `${SNAP_COMMON}/default-storage`. Upon disabling this addon you will be asked if you want to delete the persistent volumes created.
 - **ingress**: Create an ingress controller.
 - **gpu**: Expose GPU(s) to microk8s by enabling the nvidia-docker runtime and nvidia-device-plugin-daemonset. Requires NVIDIA drivers to already be installed on the host system.
+- **istio**: Deploy the core Istio services. You can use the `microk8s.istioctl` command to mange your deployments.
```
tvansteenburgh: Which ones are the "core" services? What exactly is getting deployed when I enable this? Can we hyperlink "Istio" to the Istio website or GH repo for people who want to know more about it?

microk8s-resources/actions/enable.istio.sh
```diff
@@ -0,0 +1,30 @@
...
+  sudo touch "$SNAP_USER_COMMON/istio.lock"
+fi
+
+refresh_opt_in_config "allow-privileged" "true" kubelet
+refresh_opt_in_config "allow-privileged" "true" kube-apiserver
+sudo systemctl restart snap.${SNAP_NAME}.daemon-apiserver
+sudo systemctl restart snap.${SNAP_NAME}.daemon-kubelet
+
+
+echo "Istio is coming up"
```
tvansteenburgh: "Istio is starting"

microk8s-resources/actions/enable.istio.sh
```diff
@@ -0,0 +1,30 @@
...
+set -e
+
+source $SNAP/actions/common/utils.sh
+
+echo "Enabling Istio"
+
+# pod/servicegraph will start failing without dns
+"$SNAP/microk8s-enable.wrapper" dns
+
+read -p "Enforce mutual TLS authentication between sidecars? (Y/N): " confirm
```
tvansteenburgh: Maybe we should just assume Y and not prompt for this (i.e., always use mutual TLS). It would be simpler and not require user interaction. Any downside?

microk8s-resources/actions/disable.istio.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e
+
+source $SNAP/actions/common/utils.sh
+
+echo "Disabling Istio"
+
+skip_opt_in_config "allow-privileged" kubelet
+skip_opt_in_config "allow-privileged" kube-apiserver
```
tvansteenburgh: The downside to turning off allow-privileged here is that the user may have turned them on manually, before they installed Istio, in which case we are now inadvertently reverting their configuration change.

README.md
ktsakalozos: I picked up "core" from here: https://istio.io/docs/setup/kubernetes/quick-start/#installation-steps right now I honestly do not know if there non-core services.

microk8s-resources/actions/enable.istio.sh
tvansteenburgh: Okay. Well "N" is the safer option if the user doesn't know the difference, so that should be the default. It would also be nice to say something like "(if unsure, choose N)", and maybe even provide a link (somewhere) to the options where they can read about the differences.

ktsakalozos:
Thanks, merging.

PR #100, opened 2018-09-03
Test on other distributions

ktsakalozos:
With this PR we add a test that would start an lxc container, run an sistribution specific `install-deps.sh` script and then run the tests and checn the upgrade path.

We should wire this to our CI to automate the releases to beta.

### DIFF
--- a/tests/test-distro.sh
+++ b/tests/test-distro.sh
@@ -5,7 +5,7 @@ if echo "$*" | grep -q -- 'help'; then
     echo "Usage: $prog LXC-IMAGE ORIGINAL-CHANNEL UPGRADE-WITH-CHANNEL"
     echo ""
     echo "Example: $prog ubuntu:18.04 beta edge"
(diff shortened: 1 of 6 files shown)


tests/test-distro.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+if echo "$*" | grep -q -- 'help'; then
+    prog=$(basename -s.wrapper "$0")
+    echo "Usage: $prog LXC-IMAGE ORIGINAL-CHANNEL UPGRADE-WITH-CHANNEL"
+    echo ""
+    echo "Example: $prog ubuntu:18.04 beta edge"
+    echo "Use Ubuntu 18.04 for running our tests on."
```
tvansteenburgh: s/on//

ktsakalozos:
Thank you, merging this.

Issue #53, opened 2018-06-24
Feature request: Add support for metrics server


sbile:
Please consider adding support for metrics server so that horizontal pod autoscaling works.

PR #102, opened 2018-09-04
Metrics server

ktsakalozos:
This is the metrics server addon.

Fixes https://github.com/ubuntu/microk8s/issues/53


tests/test-live-addons.py
```diff
@@ -67,3 +68,9 @@ def test_forward(self):
         """
         validate_forward()
 
+    def test_metrics_server(self):
+        """
+        Validates port forward works.
```
tvansteenburgh: s/port forward/metrics server/

tests/test-upgrade.py
```diff
@@ -91,6 +92,14 @@ def test_upgrade(self):
         except:
             print('Will not test port forward')
 
+        try:
+            enable = microk8s_enable("metrics-server")
+            assert "Nothing to do for" not in enable
+            validate_metrics_server()
+            test_matrix['metrics_server'] = validate_metrics_server
+        except:
+            print('Will not the metrics server')
```
tvansteenburgh: s/the/test/

ktsakalozos:
The metrics server is not enabled by default. It is enabled as a normal addon so I added it in the list of the addons in the README (which I should have done in the first place but forgot).

ktsakalozos:
Thank you for the review.

Issue #101, opened 2018-09-04
permission denied creating a pod


wallyworld:
https://pastebin.ubuntu.com/p/8g523hTqb3/

Deploying a pod I get various permission denied errors and the pod stays in ContainerCreating.

microk8s             v1.11.2               167   edge      canonical✓    classic

ktsakalozos:
Hi @wallyworld ,

Can you hellp me reproduce this? Are you on xenial on lxc? Can you share the manifests you are using? Also could you show me the output of `journalctl -u snap.microk8s.daemon-apiserver.service`, `journalctl -u snap.microk8s.daemon-kubelet.service` and `journalctl -u snap.microk8s.daemon-docker.service`

Thank you for reporting this.

ktsakalozos:
Could you also share your `dmesg`? I wonder if there are any apparmor denials there.

wallyworld:
Ah the dmesg llog showed apparmor denials. The journelctl logs were full of errors I think related to the apparmor root cause. All I have done is install docker snap and microk8s snap so it seems there's something misconfigured somewhere?

[11183.920576] audit: type=1400 audit(1536111210.327:1708): apparmor="DENIED" operation="ptrace" profile="docker-default" pid=1308 comm="kubelet" requested_mask="tracedby" denied_mask="tracedby" peer="snap.microk8s.daemon-kubelet"
[11184.027986] audit: type=1400 audit(1536111210.435:1709): apparmor="DENIED" operation="signal" profile="docker-default" pid=2746 comm="docker-containe" requested_mask="receive" denied_mask="receive" signal=kill peer="snap.microk8s.daemon-docker"

wallyworld:
I thought I'd try to remove microk8s and start again.

$ sudo snap remove microk8s
error: cannot perform the following tasks:
- Remove data for snap "microk8s" (167) (remove /var/snap/microk8s/common/var/lib/docker/aufs: device or resource busy)

ktsakalozos:
When docker spawns its containers an apparmor profile is applied. This profile is normally found under `/etc/apparmor.d/docker`. This default-docker profile is almost right for microk8s but not perfect. We need to make sure the dockerd packaged inside microk8s can send signals to the spawned docker containers so that they can be (for example) killed. To do so whenever snap.microk8s.daemon-docker starts
[...]

PR #107, opened 2018-09-11
Reset and docker restart will load the right docker-default profile

ktsakalozos:
This mitigates https://github.com/ubuntu/microk8s/issues/101

We reload the right docker-default profile upon dockerd restart and microk8s.reset.

Also documenting the issue with multiple dockerds and AppArmor.

### DIFF
--- a/docs/dockerd.md
+++ b/docs/dockerd.md
@@ -1,0 +1,14 @@
+# Dockerd in microk8s
+
+The docker daemon used by microk8s is listening on `unix:///var/snap/microk8s/docker.sock`. You can access it with the `microk8s.docker` command. To skip the `microk8s` prefix we suggest you employ a snap alias:
+```
+sudo snap alias microk8s.docker docker
+docker ps
+```
+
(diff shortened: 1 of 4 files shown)


docs/dockerd.md
```diff
@@ -0,0 +1,14 @@
+# Dockerd in microk8s
+
+The docker daemon used by microk8s is listening on `unix:///var/snap/microk8s/docker.sock`. You can access it with the `microk8s.docker` command. To skip the `microk8s` prefix we suggest you employ a snap alias:
+```
+sudo snap alias microk8s.docker docker
+docker ps
+```
+
+When AppArmor is enabled all docker daemons running in a system will apply the same `docker-default` profile on running containers. Each daemon makes sure that it is the only process manging the docker containers (eg send start stop signals). Effectively this allowes only one dockerd running on any host. Therefore, you have to make sure no other dockerd is running on your sytem along with microk8s.
```
tvansteenburgh: s/manging/managing/
e.g., sending start and stop signals

ktsakalozos:
Thank you for the review

Issue #119, opened 2018-09-26
Kube-proxy is failing to start because it cannot update conntrack hashsize


ktsakalozos:
This happens when we run microk8s inside LXD inside GCE.

```
Sep 26 17:32:16 machine-518 microk8s.daemon-proxy[8435]: I0926 17:32:16.153690    8435 conntrack.go:98] Set sysctl 'net/netfilter/nf_conntrack_max' to 131072
Sep 26 17:32:16 machine-518 microk8s.daemon-proxy[8435]: I0926 17:32:16.153770    8435 conntrack.go:52] Setting nf_conntrack_max to 131072
Sep 26 17:32:16 machine-518 microk8s.daemon-proxy[8435]: I0926 17:32:16.176814    8435 conntrack.go:83] Setting conntrack hashsize to 32768
Sep 26 17:32:16 machine-518 microk8s.daemon-proxy[8435]: F0926 17:32:16.176899    8435 server.go:361] write /sys/module/nf_conntrack/parameters/hashsize: operation not supported
```

I do not fully understand why this happens, needs more investigation.
Possibly related: https://github.com/kubernetes/kubernetes/issues/24295

We should at least detect this issue in `microk8s.inspect` and suggest the use of `--conntrack-max-per-core=0` in `/var/snap/microk8s/current/args/kube-proxy`.

PR #120, opened 2018-09-26
Kube proxy not touching conntrack hashsize in virtual envs

ktsakalozos:
Our CI cross-distro tests on LXC on top of GCE instances are failing with kube-proxy not starting after a `micrk8s.enable dns`.

This is a workaround (or not?) where the we tell kube-proxy not to update the conntrack hashsize during tests. The issue https://github.com/ubuntu/microk8s/issues/119 is here to further investigate this behavior.

### DIFF
--- a/tests/test-upgrade.py
+++ b/tests/test-upgrade.py
@@ -10,7 +10,7 @@ from validators import (
     validate_forward,
     validate_metrics_server
 )
-from subprocess import check_call, CalledProcessError
+from subprocess import check_call, CalledProcessError, call
 from utils import microk8s_enable, wait_for_pod_state, microk8s_disable, wait_for_installation
 
 upgrade_from = os.environ.get('UPGRADE_MICROK8S_FROM', 'beta')
@@ -33,7 +33,7 @@ class TestUpgrade(object):
         cmd = "sudo snap install microk8s --classic --channel={}".format(upgrade_from).split()
         check_call(cmd)
         wait_for_installation()
-        if in_virtual_env():
+        if is_container():
             # In some setups (eg LXC on GCE) the hashsize nf_conntrack file under
             # sys is marked as rw but any update on it is failing causing kube-proxy
             # to fail.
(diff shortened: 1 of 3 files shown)


tests/test-upgrade.py
```diff
@@ -33,6 +33,13 @@ def test_upgrade(self):
         cmd = "sudo snap install microk8s --classic --channel={}".format(upgrade_from).split()
         check_call(cmd)
         wait_for_installation()
+        if in_virtual_env():
+            # In some setups (eg LXC on GCE) the hashsize nf_conntrack file under
+            # sys is marked as rw but any update on it is failing causing kube-proxy
+            # to fail.
+            here = os.path.dirname(os.path.abspath(__file__))
+            apply_patch = os.path.join(here, "patch-kube-proxy.sh")
+            check_call("sudo {}".format(apply_patch).split())
```
kwmonroe: Do we need this here?  It seems like we would have already done this in `test-distro.sh`, and applying this patch twice would append `--conntrack-max-per-core=0` to kube-proxy args multiple times.

tests/test-upgrade.py
```diff
@@ -115,8 +122,18 @@ def test_upgrade(self):
...
+
+
+def in_virtual_env():
+    '''
+    Returns: True if the deployment is in a VM/container.
+
+    '''
+    try:
+        check_call("sudo grep -E (lxc|hypervisor) /proc/1/environ /proc/cpuinfo".split())
+        return True
```
kwmonroe: Do we want to return True for all virt environments?  ie, VMs like kvm, qemu, vmware in addition to containers like lxc, docker, and rkt?

We may be better off using `systemd-detect-virt` instead of relying on `/proc`:
https://www.freedesktop.org/software/systemd/man/systemd-detect-virt.html

You can call it to detect *all* virt like this (rc = 0 means it's a virtual env):
```
subprocess.call(['systemd-detect-virt']) == 0
```

Or just containers like this:
```
subprocess.call(['systemd-detect-virt', '--container']) == 0
```

tests/test-upgrade.py
ktsakalozos: You are right that it seems we are reapplying the patch. The real issue it that `test-distro.sh` should also be removing microk8s (not only resetting it) so that the upgrade test will really do the upgrade.

ktsakalozos:
Just got a green test run on our CI. Merging this. Thanks @kwmonroe

PR #124, opened 2018-10-03
Use script to trigger pytest. CI slaves need this.

ktsakalozos:
Seems CI slave do not have a tty and script will create one for us.

https://github.com/lxc/lxd/issues/1724#issuecomment-194416774


ktsakalozos:
Self-merging this. The travis error is because we do not have the 1.11 branch yet.

https://travis-ci.org/ubuntu/microk8s/builds/436549218

PR #126, opened 2018-10-03
remember why we're using 'script' to launch pytest

kwmonroe:
Add a comment in case we forget why `script` was needed here.


ktsakalozos:
Thanks

PR #127, opened 2018-10-04
Have 1.11 branch to build from 1.11 k8s

ktsakalozos:
This is part of the branch creation automation we need to have. A card is already created.


ktsakalozos:
Thanks @kwmonroe

PR #139, opened 2018-10-18
Correct docker socket path and add DOCKER_HOST env variables

pesimon:


ktsakalozos:
Thank you for the fixes @pesimon !

LGTM +1

Issue #160, opened 2018-11-02
Inspect performance


AdamIsrael:
I've noticed that the performance of `microk8s.inspect` degrades over time. It appeared that most of the time was spent inspecting `snap.microk8s.daemon-apiserver`.

Stepping through `check_service`, I found most of the time was spent dumping the journalctl logs. In my case, it generated an 785M log with 3.3+ million lines.

Is it really necessary to capture the full log? If not, what would a sane limit be?

ktsakalozos:
Fixed via https://github.com/ubuntu/microk8s/pull/161

Indeed the apiserver was configured with aggressive logging. Log verbosity is recenty tuned to the default values.

Your PR addresses the issue in a better way as it puts an upper limit to the log lines we will grab in inspect. Thank you.

PR #161, opened 2018-11-02
Add a default limit to journalctl

AdamIsrael:
In response to issue #160, this PR adds a default limit of 100,00 entries to dump for the inspect report.


ktsakalozos:
Thank you @AdamIsrael, this will make the inspect performance much more stable.

PR #170, opened 2018-11-09
Enable digitalSignature key usage for CA

lhotari:
- fix "KeyUsage does not allow digital signatures" error when
  connecting to api manager with Java based https client


ktsakalozos:
Thank you for the PR @lhotari . You should see your fix on the edge channel of the latest and 1.12 channels within the day (`snap install --classic --edge`).

Cheers

Issue #162, opened 2018-11-02
feature request: microk8s.[start|stop]


kwmonroe:
The current mix of `enable|disable` is confusing. We have these that affect _addons_:
```
microk8s.disable foo
microk8s.enable foo
```

And these that affect the microk8s _service(s)_:
```
sudo snap disable microk8s
sudo snap enable microk8s
```

This leads to a UX like this:
```bash
## /me wants to save my battery, so i turn off microk8s
ubuntu@bionic:~$ sudo snap disable microk8s
microk8s disabled

## time has passed and I want to interact with k8s again
ubuntu@bionic:~$ microk8s.enable

Command 'microk8s.enable' not found, but can be installed with:

sudo snap install microk8s

## /me grumbles because i confused snap vs microk8s enable

## worse, /me forgets that i disabled microk8s and do what the output says
ubuntu@bionic:~$ sudo snap install microk8s --classic
snap "microk8s" is already installed, see 'snap help refresh'
```

I would like to have control over the running services within microk8s without resorting to snap actions. I propose `microk8s.[start|stop]`, which would do the right thing to relevant system services.

If `microk8s.stop` has been run, future calls to `microk8s.*` should notify the user appropriately, as in:
```bash
ubuntu@bionic:~$ microk8s.stop
microk8s stopped

ubuntu@bionic:~$ microk8s.enable dashboard
microk8s is not running, try microk8s.start
```

ktsakalozos:
Thank you for opening the issue @kwmonroe. To add to the confusion there is also the `snap start/stop microk8s` option.

I am a big +1 for making microk8s aware of its state also for another reason. During installation and a disable/enable cycle `microk8s.kubectl` is not be able to reach the apiserver and throws out potentially confusing messages. Since we will work into improving the UX with the stop/start commands we should also make sure we should let know the user when microk8s is starting and of course return meaningful return codes (https://github.com/ubuntu/microk8s/issues/115). Look how ugly the sleep command is in the installation instructions of microk8s https://github.com/geoand/sb-fmp-microk8s .

PR #172, opened 2018-11-09
Add start stop commands

ktsakalozos:
Fixes: https://github.com/ubuntu/microk8s/issues/162

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -1,5 +1,14 @@
 #!/usr/bin/env bash
 
+exit_if_stopped() {
+  # test if the snap is marked as stopped
+  if [ -e ${SNAP_DATA}/var/lock/stopped.lock ]
+  then
+    echo "microk8s is not running, try microk8s.start"
+    exit 1
+  fi
+}
+
 
 refresh_opt_in_config() {
     # add or replace an option inside the config file.
--- a/microk8s-resources/wrappers/microk8s-start.wrapper
+++ b/microk8s-resources/wrappers/microk8s-start.wrapper
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -eu
+
+export PATH="$SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH"
+
+sudo snap start ${SNAP_NAME} --enable
+if [ -e ${SNAP_DATA}/var/lock/stopped.lock ]
+then
+  # Mark the api server as starting
+  sudo rm ${SNAP_DATA}/var/lock/stopped.lock &> /dev/null
+fi
--- a/microk8s-resources/wrappers/microk8s-start.wrapper
+++ b/microk8s-resources/wrappers/microk8s-start.wrapper
@@ -4,9 +4,14 @@ set -eu
 
 export PATH="$SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH"
 
-sudo snap start ${SNAP_NAME} --enable
-if [ -e ${SNAP_DATA}/var/lock/stopped.lock ]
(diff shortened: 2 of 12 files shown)


microk8s-resources/wrappers/microk8s-start.wrapper
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -eu
+
+export PATH="$SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH"
+
+sudo snap start ${SNAP_NAME} --enable
```
kwmonroe: Can we check the return code here and potentially offer help?  Something like,
```bash
if rc is 0
  if lock exists
    remove lock
  fi
else
  ## maybe pkill or otherwise stop any half-started services?
  echo 'Failed to start microk8s. See snap logs or journalctl or ...'
  exit 1
fi
```
I know we're `set -e`, so a failed `snap start` would exit before the lock removal, but I think self-documenting would be useful here.

microk8s-resources/wrappers/microk8s-stop.wrapper
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -eu
+
+export PATH="$SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH"
+
+sudo snap stop ${SNAP_NAME} --disable
```
kwmonroe: As with `start`, can we point the user somewhere useful if `stop` returns non-zero?

microk8s-resources/wrappers/run-with-config-args
```diff
@@ -42,6 +42,12 @@ then
       sleep 6
     done
   fi
+  if [ -e ${SNAP_DATA}/var/lock/stopped.lock ]
+  then
+    # Mark the api server as starting. This is needed incase you
+    # microk8s.stop and then snap start microk8s
+    rm -rf ${SNAP_DATA}/var/lock/stopped.lock &> /dev/null
```
kwmonroe: Nice corner case handling, but the `-r` is unnecessary.

microk8s-resources/actions/common/utils.sh
```diff
@@ -1,5 +1,14 @@
 #!/usr/bin/env bash
 
+exit_if_stopped() {
+  # test if the snap is marked as stopped
+  if [ -e ${SNAP_DATA}/var/lock/stopped.lock ]
+  then
+    echo "microk8s is not running, try microk8s.start"
+    exit 1
+  fi
+}
```
kwmonroe: @ktsakalozos while we're in here, what do you think about trying to address a bit of #174 in this function?  I was thinking we expand it like this:
```bash
if -e stopped.lock
    echo "microk8s is not running, try microk8s.start"
    exit 1
else
  if check_apiserver: ## apiserver port unreachable; systemclt status apiserver != running; apiserver process not running; etc:
    while retry < 5
      echo "waiting for microk8s api server to start"
      sleep 2
      if check_apiserver == good
        return
      else
        retry++
    done
    echo "microk8s api server did not start, see journalctl"
    exit 1
  fi
fi
```
Since this PR adds the `exit_if_stopped` check to all the relevant wrappers, i think it makes sense to use this util as a "stopped or not running" gate.

ktsakalozos: I am already on it @kwmonroe but not in this PR. Have a look here https://github.com/ubuntu/microk8s/pull/178/files#diff-0f01327987334a699c82fa97496f4895R10 . With the `microk8s.status` command you will be able to get the status (running or not), and also wait for microk8s to get ready (with a timeout or wait forever). :)

microk8s-resources/wrappers/microk8s-stop.wrapper
ktsakalozos: Done

PR #179, opened 2018-11-13
Fix Docker docker path in docs

darabos:
Well at least on my system (`bento/ubuntu-18.04` in Vagrant) this is the path the socket ends up at.

Thanks for microk8s, it's great!


ktsakalozos:
Same here :)

+1 LGTM

Thank you for this PR and for using microk8s.

PR #181, opened 2018-11-14
 Add proxy setting for pip fix

ktsakalozos:
This changes are needed for pip install not to fail in our CI.

We make sure pip uses the proxy passed.


ktsakalozos:
@battlemidget these are the changes for the latest and 1.12 tracks. Need to do the same for 1.11 and 1.10.

PR #185, opened 2018-11-15
Increase timeouts, and http retries

ktsakalozos:
Add retries when talking to xip.io and the snapstore.

Longer waits while enabling addons.


ktsakalozos:
Merging this to move on with testing.

PR #186, opened 2018-11-15
Increase timeouts, and http retries

ktsakalozos:


ktsakalozos:
Merging this to move on with testing.

PR #190, opened 2018-11-16
Better dns testing

ktsakalozos:
Move dns test along with dashboard testing becasue we are running short on time on travis.

Fix dns testing


ktsakalozos:
With this we actually test dns.

PR #193, opened 2018-11-19
Faster nvidia test abort

ktsakalozos:
We increased the timeouts so our CI will not timeout, but travis fails if the tests run for more than 2 hours.

This PR will cause gpu tests to be aborted quickly in case no nvidia drivers are loaded.


ktsakalozos:
Merging this as part of the on-going CI tuning. Please review at your own pace.

PR #204, opened 2018-11-23
Add building from source doc

evilnick:
This adds the 'building from source' documentation as a standalone page.

* rewritten as steps
* added link to Snapcraft docs

@ktsakalozos my intention here is to add some standalone pages (one PR at a time) and then adapt the readme to take them into account.

The readme and the final website docs will be slightly different (we don't have side navigation in the readme) but not too much. Making the changes here should improve the docs in this repo and make it easier to sync with the published docs online

### DIFF
--- a/docs/build_from_source.md
+++ b/docs/build_from_source.md
@@ -58,7 +58,7 @@ After a standard build, run the following:
 ```bash
 cp -r parts/microk8s/build/build/kube_bins .
 export KUBE_SNAP_BINS=$PWD/kube_bins/v1.12.2/
- snapcraft
+snapcraft
  ```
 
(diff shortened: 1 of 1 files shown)


ktsakalozos:
Should we remove the build instructions from the main README and add a link to this work?

evilnick:
@ktsakalozos I'd rather land the separate pieces and then rework the readme - there are a few more pages to go, so it'll be easier just to rewrite the readme once

ktsakalozos:
Fair enough

ktsakalozos:
LGTM +1

ktsakalozos:
> there are a few more pages to go

Could you give us an idea on what is coming?

evilnick:
The other parts from that outline doc ;)
I'll do the add-ons page next

PR #208, opened 2018-11-24
Call dpkg from the snap

ktsakalozos:
This is a fix for the distros that do not have dpkg. Since we are using it to get the architecture we have to have it in our snap.


ktsakalozos:
Merging this so we can move on with a release.

PR #246, opened 2018-12-14
Update nvidia runtime to match dockerd upgrade

ktsakalozos:
Tested on a AWS p2 machine


ktsakalozos:
Tested on AWS and local. Merging this to fix edge. Please review when you have time.

PR #248, opened 2018-12-14
Update nvidia runtime to match dockerd upgrade.

ktsakalozos:


ktsakalozos:
Going to merge this because I do not want a faulty build to hit edge. I am confident this PR is fine because it passed the 1.12 and master branches.

The reason why we do not have enough time to run the tests on travis is because we increased the timeouts to have our ci passing.

PR #254, opened 2018-12-19
Added automatic bash completion for microk8s.docker microk8s.kubectl …

assafmendelson:
…and microk8s.istioctl

Note: the docker bash completion currently gets the latest from the docker repository (while istioctl and kubectl are from the actual binaries). It would be better if in the future we can figure out the exact version and match the completion to that.
For now this should be good enough

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -193,36 +193,32 @@ parts:
         echo "Preparing istio"
         cp $KUBE_SNAP_BINS/istioctl .
         cp -r $KUBE_SNAP_BINS/istio-yaml ./actions/istio
+        # Add bash completion for istioctl. This is located here to make sure it is only added when istio is added.
+        if [ -f istioctl ]; then
+          ./istioctl collateral completion --bash -o .
+          sed -i  "s/complete -o default -F __start_istioctl istioctl/complete -o default -F __start_istioctl microk8s.istioctl/g" istioctl.bash
+          sed -i  "s/complete -o default -F __start_istioctl istioctl/complete -o default -o nospace -F __start_istioctl microk8s.istioctl/g" istioctl.bash
+        else
+           echo "NO istioctl FOUND!!!!!!!!!!!!!!! ABORTING istioctl bash completion!!!!!!!!!!!!"
+           exit -1
+        fi
       fi
 
       echo "Creating inspect hook"
       cp $KUBE_SNAP_ROOT/scripts/inspect.sh .
 
-      # create bash completion script for microk8s.docker, microk8s.kubectl and microk8s.istio
-      # create docker completion. At this stage, the latest from the docker repository is used as a basis, it should
-      # probably be replaced with something else which would bring the completion matching the specific version inside
-      # microk8s
+      # Add bash completion for microk8s.docker
+      cp $SNAPCRAFT_STAGE/usr/share/bash-completion/completions/docker docker.bash
+      sed  -i "s/complete -F _docker docker docker.exe dockerd dockerd.exe/complete -F _docker microk8s.docker/g" docker.bash
 
-      curl -L https://raw.githubusercontent.com/docker/cli/master/contrib/completion/bash/docker | sed "s/complete -F _docker docker docker.exe dockerd dockerd.exe/complete -F _docker microk8s.docker/g" > docker.bash
-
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -195,6 +198,31 @@ parts:
...
+      # probably be replaced with something else which would bring the completion matching the specific version inside
+      # microk8s
+
+      curl -L https://raw.githubusercontent.com/docker/cli/master/contrib/completion/bash/docker | sed "s/complete -F _docker docker docker.exe dockerd dockerd.exe/complete -F _docker microk8s.docker/g" > docker.bash
+
+      # Assume kubectl and istioctl are already copied to the current directory
+      if [ -f kubectl ]; then
+        ./kubectl completion bash | sed "s/complete -o default -F __start_kubectl kubectl/complete -o default -F __start_kubectl microk8s.kubectl/g" | sed "s/complete -o default -o nospace -F __start_kubectl kubectl/complete -o default -o nospace -F __start_kubectl kubectl/g" > kubectl.bash
+      else
+         echo "NO kubectl FOUND!!!!!!!!!!!!!!! ABORTING kubectl bash completion!!!!!!!!!!!!"
```
ktsakalozos: We do not want to accidentally ship with bash completion missing so we would better fail the snapcraft build instead here.

snapcraft.yaml
```diff
@@ -195,6 +198,31 @@ parts:
       echo "Creating inspect hook"
       cp $KUBE_SNAP_ROOT/scripts/inspect.sh .
 
+      # create bash completion script for microk8s.docker, microk8s.kubectl and microk8s.istio
+      # create docker completion. At this stage, the latest from the docker repository is used as a basis, it should
+      # probably be replaced with something else which would bring the completion matching the specific version inside
+      # microk8s
+
+      curl -L https://raw.githubusercontent.com/docker/cli/master/contrib/completion/bash/docker | sed "s/complete -F _docker docker docker.exe dockerd dockerd.exe/complete -F _docker microk8s.docker/g" > docker.bash
```
ktsakalozos: We should pin the docker.io package we pull from the deb repository https://github.com/ubuntu/microk8s/blob/master/snapcraft.yaml#L93 . The reason is that for GPU support we have to cherry-pick the correct nvidia debs (eg https://github.com/ubuntu/microk8s/blob/master/snapcraft.yaml#L223). Pinning the docker.io package would allow us to find the exact bash completion script.

assafmendelson: I agree (I figured we could do it later on). 

In any case, is there a common path to where the docker package is installed in both cases? When installing from the regular docker.io package, the file is located in parts/docker/install/usr/share/bash-completion/completions/docker. I am not sure where it would be located in the nvidia case though.

snapcraft.yaml
```diff
@@ -190,11 +193,32 @@ parts:
         echo "Preparing istio"
         cp $KUBE_SNAP_BINS/istioctl .
         cp -r $KUBE_SNAP_BINS/istio-yaml ./actions/istio
+        # Add bash completion for istioctl. This is located here to make sure it is only added when istio is added.
+        if [ -f istioctl ]; then
```
ktsakalozos: Could we also skip the file existence checks? If any executable (`istioctl` or `kubectl`) is missing calling it will fail stopping snapcraft build. This is the reason we have `set -eu` at the top.

assafmendelson: Fixed

snapcraft.yaml
```diff
@@ -190,11 +193,22 @@ parts:
         echo "Preparing istio"
         cp $KUBE_SNAP_BINS/istioctl .
         cp -r $KUBE_SNAP_BINS/istio-yaml ./actions/istio
+        # Add bash completion for istioctl. This is located here to make sure it is only added when istio is added.
```
ktsakalozos: I believe we also need a `touch istioctl.bash` for non-amd64 architectures here   
https://github.com/ubuntu/microk8s/pull/254/files#diff-184032a532406b07009403e26f4fc62fR188 . In non-amd64 architectures the `microk8s.istioctl` command still exists but the wrapper tells you that istio is not available for the current architecture.

ktsakalozos:
Thank you @assafmendelson. The patch should be available within a day on the edge channel!

PR #262, opened 2018-12-23
Adjust default backend targetPort for ingress.yaml

davefinster:
The containerPort in the default-http-backend deployment is defined as 8080, but the service definition is pointing to 80. When I deployed this, I kept getting unreachable upstream errors from nginx. Making this change corrected the problem, which given the container definition, makes sense.


ktsakalozos:
Thank you @davefinster

PR #267, opened 2018-12-29
fix dockerd sets iptables FORWARD policy to DROP

### LINKED ISSUE #266: microk8s is sets iptables policy FORWARD to DROP on reboot
I created a fresh install of ubuntu 18.04 minimal on a vmware esxi 6.7u1 host. i checked `iptables -S` and got this output (truncated):
```
-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT
```
This is consistent, no matter how often i reboot. Then i installed microk8s and all looks fine:
```
robert@k:~$ sudo snap install microk8s --edge --classic
microk8s (edge) v1.13.1 from Canonical✓ installed
robert@k:~$ sudo iptables -S
-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT
<trimmed>
```
However, after a reboot:
```
robert@k:~$ sudo iptables -S
[sudo] password for robert: 
-P INPUT ACCEPT
-P FORWARD DROP
-P OUTPUT ACCEPT
<trimmed>
```
I've attached the `microk8s.inspect` tarball as
[inspection-report-20181229_224350.tar.gz](https://github.com/ubuntu/microk8s/files/2716771/inspection-report-20181229_224350.tar.gz)

mnbf9rca:
fixes #266 so that policy is persistent across reboots

### DIFF
--- a/microk8s-resources/default-args/dockerd
+++ b/microk8s-resources/default-args/dockerd
@@ -4,3 +4,4 @@
 --graph ${SNAP_COMMON}/var/lib/docker
 --pidfile ${SNAP_COMMON}/docker-pid
 --config-file=${SNAP_DATA}/args/docker-daemon.json
+--iptables=false


gbevan:
hi, for reference, this is how i fixed it using the DOCKER USER chain instead of tweaking the FORWARD DROP rule - https://gist.github.com/gbevan/8a0a786cfc2728cd2998f868b0ff5b72  there is a link in there to the original article that I found and followed.

hacst:
I can't judge if this PR is the right approach but I think I also stumbled over this behavior resulting in my KVM setup breaking. Due to this iptables change my virtual bridge for KVM stopped working and I had to manually allow forward for the bridge device so KVM VMs could use the bridge device again. The bridge device (br0) was created manually and had nothing to do with docker or microk8s. It would be great if microk8s could work in a way that doesn't affect configuration unrelated to it.

planetbeing:
It appears Docker will only set the default FORWARD policy to DROP if before launching, `net.ipv4.ip_forward` is not enabled and it had to enable it. (https://github.com/docker/libnetwork/blob/master/drivers/bridge/setup_ip_forwarding.go)

Editing `/etc/sysctl.conf` to uncomment the `net.ipv4.ip_forward=1` line in Ubuntu appears to stop this behavior and perhaps be a more elegant way of addressing this issue.

planetbeing:
It also may be instructive to read why the policy was changed by Docker in the first place: https://github.com/moby/moby/issues/14041 and see if it applies here. If it does, it may not be a good idea for microk8s to advise users to set the policy in the first place and instead add additional iptables rules analogous to what was added for the docker0 interface for the cbr0 interface to get Kubernetes pod networking to work.

ktsakalozos:
> I had luck getting networking to work with the following iptables/systemd unit:
> https://gist.github.com/planetbeing/7c8f919a4dfbacf7681145a402884dbd

I like this. It sets iptable rules on only the interface we create (`cbr0`). It should also work through our transition to containerd. We could feature flag these few iptable rules and have them applied everytime we start dockerd https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/run-docker-with-args#L31 . Your thoughts?

ktsakalozos:
Closing this PR due to inactivity.

Issue #43, opened 2018-06-06
Check availability of port 8080 in snap install hook


tvansteenburgh:
If 8080 isn't available, increment and try again up to some reasonable limit. In other words, find an available port for the apiserver to use instead of just quietly failing if 8080 is already in use.

odino:
Also, 8080 is quite of a commonly used port :) it'd be great if microk8s would have an env variable / config where users could specify on what port to bind.

ktsakalozos:
Yes, that is true.

We cannot pass parameters during snap install. What we could do is to select an open port from a range during the installation (eg first available port in [8080, 8090]) and/or have a microk8s.init command where you would be asked this kind of questions (similar to what for example lxd is doing with lxd init).

odino:
@ktsakalozos I also like the way kubectl manages things, where you can do a `config set / get` to update parameters -- so something like `microk8s.config set APISERVER_PORT 9876` would be nice :)

ktsakalozos:
This is interesting. This configuration approach might extend to addons as well, for example we could do `microk8s.config set REGISTRY_SIZE 80G` to set the size of the registry in the respective addon.

tvansteenburgh:
Here's a quick workaround for running the api-server on a different port:
```
API_PORT=8081
echo "--insecure-port=$API_PORT" | sudo tee -a /var/snap/microk8s/current/args/kube-apiserver 
sudo systemctl restart snap.microk8s.daemon-apiserver.service 
```
You can then run `microk8s.inspect` to confirm all services are running properly.

johnraz:
The above procedure will also make `microk8s.status` to return the cluster as not running because it will use `kubectl` with the wrong kubeconfig too.

Otherwise it seems to work nicely.

fhemberger:
Default secure port for the Kubernetes is `6443`. Assuming `6080` as insecure port for consistency, why not chose `60080` and `60443` for microk8s to avoid collision with an existing Kubernetes API endpoint?

Those ports aren't likely to be used otherwise.

tristanlins:
This is unfortunately a blocker for me :-(
Tomcat uses port 8080 by default and I certainly will not reconfigure all my development tomcats or spring boot apps ...
The workaround of @tvansteenburgh is unfortunately only a crutch. So I'll stay with Minikube for the time being, in the hope that a reasonable solution will be found here soon ;-)

How about a microk8s.config tool that helps to reconfigure?
[...]

PR #269, opened 2018-12-30
WIP: Make k8s API port configurable

fhemberger:
Based on the discussions in #43, I started working on making the Kubernetes API port configurable.

I looked up ways to do this:
https://docs.snapcraft.io/configuration-in-snaps/510
https://github.com/nextcloud/nextcloud-snap/commit/7ddf70aa7df2eef7aaee444ffb20bd19532f99c8

but I'm not sure if this is the right approach. I'd prefer to set the port right on installation of the snap, if that's even possible.

**This is very much WIP. I'm also not particularly familiar with snap, so any help is appreciated**:

- [ ] Variable substitution in config files isn't working yet
- [ ] After changing of the port, what's the best way to restart all affected components?
- [ ] `kubectl` config needs to be updated (with variable substitution)

It would be nice if someone could give me a hand with this.


ktsakalozos:
Thank you for this PR @fhemberger awesome news. Here are some thoughts on how we could implement this feature.

Let's create an environment file under `${SNAP_DATA}/etc/microk8s.environment` . This file will be our "database". We will grow this file as we add more configuration options. `microk8s.environment` may eventually look like:

```
KUBE_API_INSECURE_PORT=8080
KUBE_API_SECURE_PORT=6443
KUBE_API_INSECURE_INTERFACE=127.0.0.1
....
```

The above file is intentionally like this so that bash scripts would just source it in order to know about the environment variables set.

The configure hook is called upon microk8s installation and whenever a snap variable changes [1]. This hook would do two things:
 - Update the `microk8s.environment`
 - Update any yaml and arguments files

Here is how the configure hook would look like (pseudocode):
```
source ${SNAP_DATA}/etc/microk8s.environment
NEW_KUBE_API_INSECURE_PORT="$(snapctl get k8s.api-insecure-port)"
if ! [ -z "${NEW_KUBE_API_INSECURE_PORT}" ] && [ "${NEW_KUBE_API_INSECURE_PORT}" different from "${KUBE_API_INSECURE_PORT}"] ; then
  render_arguments_files
  render_kubeconfig
  sudo snap stop ${SNAP_NAME}
  sudo snap start ${SNAP_NAME}
  # Or only restart the affected services
fi
 ```
 
 You see there are two calls, `render_arguments_files` and `render_kubeconfig`, to update the `${SNAP_DATA}/args/*` files and the kubeconfig used from the clients (`client.config`) to access MicroK8s. We could have template files to produced all args and kubeconfig files. For example, for the kube-controller-manager we would have a `kube-controller-manager.template`:
 ```
--master='http://127.0.0.1:${K8S_API_PORT}'
--service-account-private-key-file=${SNAP_DATA}/certs/serviceaccount.key
--root-ca-file=${SNAP_DATA}/certs/ca.crt
 ```
 
These template files should be kept under `${SNAP_DATA}` (eg `${SNAP_DATA}/var/templates`) so as to be editable by the user in case he want to tune them to his needs.

The case of kubeconfig is a bit more involving as we will need to move it from `${SNAP}` to `${SNAP_DATA}`.
 
[1] https://docs.snapcraft.io/supported-snap-hooks/3795

fhemberger:
Hi, thanks for your feedback! Unfortunately, the year started with lots of client work for me, so I don't know when I find the time to work on a new draft. So if anybody wants to chime in, I'd be more than happy.

PR #279, opened 2019-01-15
Use netstat to get the default interface name

slurms:
The script that runs detection of new IP addresses fails under Ubuntu 18.04 with the message:

```
microk8s.daemon-apiserver-kicker[29730]: link: error fetching interface information: Device not found
```

In one place, `netstat` is used, and the other two `ip` is. `netstat` appears to work, so use that instead.

### DIFF
--- a/microk8s-resources/wrappers/apiservice-kicker
+++ b/microk8s-resources/wrappers/apiservice-kicker
@@ -22,7 +22,7 @@ do
        systemctl is-active --quiet snap.microk8s.daemon-apiserver.service
     then
         # Get current IP
-        DEFAULT_INTERFACE="$(ip route | gawk '/default/ { print $5 }' | head -1)"
+        DEFAULT_INTERFACE="$(netstat -rn | grep '^0.0.0.0' | awk '{print $NF}' | head -1)"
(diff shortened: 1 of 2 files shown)


slurms:
Hi @ktsakalozos 

No worries. I've uploaded the output of `microk8s.inspect` to this dropbox [link](https://www.dropbox.com/s/8l6s6cxc4kudrzd/inspection-report-20190131_143455.tar.gz?dl=0).

The output of `ip route` is:
```bash
nick@farnsworth:~$ ip route
default dev ppp0 scope link 
150.101.32.44 dev ppp0 proto kernel scope link src 114.198.20.120 
172.17.0.0/16 dev docker0 proto kernel scope link src 172.17.0.1 linkdown 
192.168.1.0/24 dev lan proto kernel scope link src 192.168.1.1 
```

which means the output of `ip route | gawk '/default/ { print $5 }' | head -1` ends up being:
```bash
nick@farnsworth:~$ ip route | gawk '/default/ { print $5 }' | head -1
link
```

I have no problems with `netstat -rn | grep '^0.0.0.0' | awk '{print $NF}' | head -1` which gives (correctly):
```
nick@farnsworth:~$ netstat -rn | grep '^0.0.0.0' | awk '{print $NF}' | head -1
ppp0
```

This is a system running Ubuntu 18.04 I believe.

ktsakalozos:
LGTM +1 thank you for the PR. It should be released with the next point release of kubernetes.

PR #283, opened 2019-01-18
Added user guideline with PATH for snap and snap core dependency

xgenvn:
I'm on Debian 9, and these changes just to make sure that the dependency as well as PATH for snap are set correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@
 Snaps are frequently updated to match each release of Kubernetes. The quickest way to get started is to install directly from the snap store. You can install MicroK8s and let it update to the latest stable upstream Kubernetes release with:
 
 ```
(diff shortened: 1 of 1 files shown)


ktsakalozos:
Hi @xgenvn,

Thank you for the PR. I am afraid I will have to reject these changes. Here is why.

`snap install microk8s --classic` will install the core snap if it is missing. The reason you had to do a `snap install core` is reported as a bug here https://forum.snapcraft.io/t/snapd-not-apt-installing-properly-on-debian-9/8240/7 . I also hear the snapd package for Debian has been updated and it has to propagate through the Debian release process to reach the end users.

Thank you. We appreciate your time and effort.

PR #288, opened 2019-01-22
Add troubleshooting section about log collectors

yasinuslu:
I'm migrating my application from docker-compose to kubernetes. While i was doing that i wanted to add efk stack to the mix. I spent some hours before i realized that i was trying to collect logs from an irrelevant docker daemon :)

I wanted to add this to troubleshooting section so that other people can save this time

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ microk8s.stop
 microk8s.start
 ```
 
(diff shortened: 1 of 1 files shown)


ktsakalozos:
Thank you for the PR @yasinuslu

Issue #280, opened 2019-01-16
Unclear how to reinstall


neerolyte:
Please run `microk8s.inspect` and attach the generated tarball to this issue.

Skipped, nothing's running, I'd like to reinstall but can't.

The removal scripts for the snap are failing so I can't reinstall a corrupt environment:

```
$ snap remove microk8s 
error: cannot perform the following tasks:
- Stop snap "microk8s" services ([--root / is-enabled snap.microk8s.daemon-proxy.service] failed with exit status 1: Failed to get unit file state for snap.microk8s.daemon-proxy.service: No such file or directory
)
- Remove data for snap "microk8s" (335) (remove /var/snap/microk8s/common/var/lib/docker/btrfs/subvolumes/69dd65045e5790f7c3b3e63de92b7bb46e919211e06ab32bdfcfc32f9315ac51: operation not permitted)
$ sudo snap install microk8s --classic 
snap "microk8s" is already installed, see 'snap help refresh'
$ sudo snap refresh microk8s 
error: cannot refresh "microk8s": refreshing disabled snap "microk8s" not supported
```

ktsakalozos:
Hi @neerolyte

Can you try this:
```
sudo snap enable microk8s
# wait a few seconds and then
sudo snap remove microk8s
```
I suspect the issue you are facing is probably an unclean removal of the snap that caused it to go into a "disabled" state.

Thank you.

neerolyte:
```
$ sudo snap enable microk8s; date
microk8s enabled
Thu 17 Jan 08:54:50 AEDT 2019
$ date; sudo snap remove microk8s
Thu 17 Jan 08:56:00 AEDT 2019
error: cannot perform the following tasks:
- Stop snap "microk8s" services ([--root / is-enabled snap.microk8s.daemon-kubelet.service] failed with exit status 1: Failed to get unit file state for snap.microk8s.daemon-kubelet.service: No such file or directory
)
- Remove data for snap "microk8s" (335) (remove /var/snap/microk8s/common/var/lib/docker/btrfs/subvolumes/69dd65045e5790f7c3b3e63de92b7bb46e919211e06ab32bdfcfc32f9315ac51: operation not permitted)
```

I should probably clarify that I've tried running a bunch of the snap commands in the hope that something would help already (including `snap enable`).

neerolyte:
I think something in the (snap supplied?) docker that it was using was not able to handle the btrfs storage driver I just tried cleaning up the subvols manually:

```
[...]

PR #296, opened 2019-01-28
fix uninstall failure when Btrfs is used as the storage driver

satoru-takeuchi:
`snap remove microk8s` fails as follows if the storage driver is Btrfs.

```
$ sudo snap remove microk8s
error: cannot perform the following tasks:
- Remove data for snap "microk8s" (383) (remove /var/snap/microk8s/common/var/lib/docker/btrfs/subvolumes/<id>: operation not permitted)
```

It's because Btrfs subvolumes can't be deleted by rmdir() syscall.
Although this problem disappear if the linux kernel version is
v4.18 or more, it's better to fix it for older kernel users.

### DIFF
--- a/snap/hooks/remove
+++ b/snap/hooks/remove
@@ -9,6 +9,13 @@ snapctl stop ${SNAP_NAME}.daemon-docker 2>&1 || true
 (cat /proc/mounts | grep ${SNAP_COMMON}/var/lib/docker | cut -d ' ' -f 2 | xargs umount) || true
 (cat /proc/mounts | grep ${SNAP_COMMON}/var/run/docker | cut -d ' ' -f 2 | xargs umount) || true
 
+BTRFS_SUBVOL_DIR=${SNAP_COMMON}/var/lib/docker/btrfs/subvolumes/
+if [ -d ${BTRFS_SUBVOL_DIR} ] ; then
+  for snap in $(ls ${BTRFS_SUBVOL_DIR}) ; do
+    btrfs sub del ${BTRFS_SUBVOL_DIR}/${snap} 2>&1 || true
+  done
+fi
+
 #TODO(kjackal): Make sure this works everywhere we want
 if [ -f /etc/apparmor.d/docker ]; then
   echo "Updating docker-default profile"


ktsakalozos:
Hi @satoru-takeuchi 

Can you please show me how you setup MicroK8s with btrfs so I can reproduce your setup? We try to keep MicroK8s self-contained meaning that we want binaries such as `btrfs` to be shipped with the snap. 

Thank you for the PR

satoru-takeuchi:
Thank you for your response, @ktsakalozos

I setup microk8s just by the following command with btrfs root filesystem.

```
$ mount | grep btrfs
/dev/sda1 on / type btrfs (rw,relatime,ssd,space_cache,subvolid=257,subvol=/@) 
/dev/sda1 on /home type btrfs (rw,relatime,ssd,space_cache,subvolid=258,subvol=/@home)
$ sudo snap install microk8s --classic
microk8s v1.13.2 from Canonical✓ installed
$ microk8s.docker info | grep btrfs
Storage Driver: btrfs
WARNING: No swap limit support
```

However, I guess you're not a Btrfs user. To make Btrfs as a docker' storage driver
from non-btrfs users, the following operations are needed.

https://docs.docker.com/storage/storagedriver/btrfs-driver/

As for microk8s, we should replace "/var/lib/docker" with "/var/snap/microk8s/common/var/lib/docker/".

satoru-takeuchi:
> In https://github.com/ubuntu/microk8s/pull/296/files#diff-5e0254f4e4d4ae7c6097bbc7feda139aR15 how 
> do we know btrfs is on our path?

I didn't consider that, sorry.

> If we were to use the full path how would we know it it the same across all distros.

I see. So I should use just "btrfs" rather than full path, .../btrfs).

> Would it work if we put btrfs in the snap? We need to add it in this list:
> https://github.com/ubuntu/microk8s/blob/master/snapcraft.yaml#L114

I have one question about it.

How can I install btrfs command by setting stage-packages by distribution
independent way?

For example, btrfs commandis included in "btrfs-tools" package on Ubuntu16.04 and
"btrfs-progs" package on Ubuntu 18.04. However, as I read on the snapcraft's spec,
packages written in stage-packages are installed by distribution's package manager.
If my understanding is correct, when I add "btrfs-tools", it doesn't work on Ubuntu
18.04 and if I add "btrfs-progs", it doen't work on Ubuntu 16.04.

https://docs.snapcraft.io/snapcraft-yaml-reference/4276

ktsakalozos:
The MicroK8s snap is based on Ubuntu 16.04 (we may move to 18.04 at some point in the future). This means that is you add "btrfs-tools" in the staging packages btrfs will be in the snap under `${SNAP}`. So in your patch you can access the btrfs client using the full path `${SNAP}/...path inside the snap.../btrfs`. We should also make sure docker is indeed using the btrfs tools shipped with the snap. We do use the right paths in https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/run-docker-with-args but lets double check. Also an FYI, we will be swapping dockerd with containerd in future releases.

satoru-takeuchi:
Thank you for giving me the detailed explanation. I'll fix my PR.

rirl:
Sorry for a lame question here, but I see this problem on Ubuntu 18.04.02 (server).
is there some manual work-around someone can suggest to allow me to remove the dangling microk8s subvolumes?  Thanks for any input.

df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            5.9G     0  5.9G   0% /dev
tmpfs           1.2G  1.7M  1.2G   1% /run
/dev/sda3       132G  5.5G  125G   5% /
tmpfs           5.9G   76M  5.8G   2% /dev/shm
tmpfs           5.0M  4.0K  5.0M   1% /run/lock
tmpfs           5.9G     0  5.9G   0% /sys/fs/cgroup
/dev/loop5      2.3M  2.3M     0 100% /snap/icey-lynx/1
/dev/loop0      3.8M  3.8M     0 100% /snap/gnome-system-monitor/57
/dev/loop1       92M   92M     0 100% /snap/core/6531
/dev/loop2       36M   36M     0 100% /snap/gtk-common-themes/1198
/dev/loop3      141M  141M     0 100% /snap/gnome-3-26-1604/82
/dev/loop6       91M   91M     0 100% /snap/core/6350
tmpfs           1.2G   12K  1.2G   1% /run/user/1000

sudo snap remove microk8s
error: cannot perform the following tasks:
- Remove data for snap "microk8s" (412) (remove /var/snap/microk8s/common/var/lib/docker/btrfs/subvolumes/1d206346bc6cd81c1aaf16da9698fd5fdc2dfaed5af3a1d606d8f0c78451fae6: operation not permitted)

microk8s.status

Command 'microk8s.status' not found, but can be installed with:

sudo snap install microk8s

satoru-takeuchi:
@rirl 

> is there some manual work-around someone can suggest to allow me to remove the dangling microk8s subvolumes? Thanks for any input.

yes, you can remove them as follows.

https://github.com/ubuntu/microk8s/issues/280#issuecomment-454961142

satoru-takeuchi:
I don't use Btrfs in the host of microk8s anymore. So close it.

Issue #301, opened 2019-02-01
Ingress on arm64 is failing


ktsakalozos:
With 
```
NGINX Ingress controller
  Release:    0.11.0
  Build:      git-9d0ce7c
  Repository: https://github.com/kubernetes/ingress-nginx
unknown flag: --publish-status-address
```

Aworkaround for now is to `microk8s.kubectl edit daemonset.apps/nginx-ingress-microk8s-controller` and remove the unknown flag: `--publish-status-address`

PR #307, opened 2019-02-06
Fix ingress on arm64

ktsakalozos:
Fixes: https://github.com/ubuntu/microk8s/issues/301


ktsakalozos:
In addition to the ingress failure on arm this PR also disables addons and respective tests of fluentd and Jaeger that are not valid for arm64.

PR #321, opened 2019-02-13
fix typo

Ohadbasan:


ktsakalozos:
Hi @Ohadbasan,

Do you see why the build is failing https://travis-ci.org/ubuntu/microk8s/builds/492799022#L2250 ?

Also for the record, this is not a typo. ISTIO_ERSION holds the ISTIO_VERSION without the "v". For example lets say ISTIO_VERSION="v1.0.5", ISTIO_ERSION="1.0.5". But since this seems to confuse people I do not object to your PR.

Ohadbasan:
I see the failures. I am not sure its related to the PR.
seems like the the micro8s_reset is trying to access a file that doesn't exist.
am I wrong?

ktsakalozos:
The build of the snap is failing on travis. Are you able to build the snap locally without any issues?

The error:
```
/bin/sh: 35: build-scripts/fetch-binaries.sh: Bad substitution
```

Ohadbasan:
I am using mac os and a gnu/linux distro is mandatory for snap building (afaik)
I have just validated the syntactic validity of this changes on my machine (using bash 4.4)
if this script is invoked using dash or a vanilla posix shell, it might be the cause of the issue but I can not reproduce it on my mac machine.

PR #323, opened 2019-02-15
Containerd and security fixes for 1.14 release

ktsakalozos:
This PR includes both the security fixes and the move to containerd.


magne:
I've tested the PR, and it seems to have some error in the `snapcraft.yml` file. When I try to build using `snapcraft cleanbuild`, I get the following:
```
Creating snapcraft-taciturnly-xanthous-amee
Starting snapcraft-taciturnly-xanthous-amee
Waiting for a network connection...
Network connection established
```
after that it just sits there, no CPU or network activity.

I had to use the docker command `docker run -v $(pwd):$(pwd) -w $(pwd) snapcore/snapcraft:stable sh -c "apt update && snapcraft"` taken from the Travis file to build it.

I've written an addon that will enable/disable RBAC based on this PR, and it seems to work. Authorization needs kubectl to connect to the api server through the secure port to work, which is why this PR is important. The kubectl config that is used out-of-the box still uses the insecure port, I would suggest that you switch to the secure port as part of this PR. Preferably, the insecure port should be disabled, but I realize that this will break backward compatibility.

Maybe we could have some command that will secure the cluster?

BTW, the `microk8s.inspect` command shows failing daemon-docker (naturally).

ktsakalozos:
Hi @magne,

We really appreciate your feedback.

The issue with the `snapcraft cleanbuild` is rather strange. `cleanbuild` spawns an lxc container and does the build inside there. It seems this lxc provisioning step is waiting for user input. I do not know why but opened a support topic on the snapcraft forum.
https://forum.snapcraft.io/t/snapcraft-cleanbuild-waiting-for-input/10053

In respect to `microk8s.kubectl` using the secure port, I have no objections. The insecure port is deprecated and will be removed in the near future from upstream Kubernetes, so we might as well do the transition now that we have backwards incompatible changes. So, in the last commit, `microk8s.kubectl` is using arguments found in `/var/snap/microk8s/curent/args/kubectl`. New installations will use a kubeconfig file poitning to the secure port. Old MicroK8s installations upgraded will keep using a kubeconfig pointing to the insecure port. This means that in your PR you can detect the present mode (secure/insecure) and provide a warning saying something like, "RBAC is enabled, but your microk8s.kubectl is using the insecure port to access the cluster."

Thank you for spotting the `microk8s.inspect` problem. Should be fixed in the latest snap pushed on the `1.13/edge/secure-containerd` channel. Give it a try:

```
snap install microk8s --classic --channel=1.13/edge/secure-containerd
```
Or just refresh to get the latest (434) revision.

Many thanks.

waquidvp:
Could we update to containerd version v1.2.4 and the corresponding runc so that the fix for CVE-2019-5736 is in it.

ktsakalozos:
Hi @waquidvp, thank you for bringing the CVE to my attention. How much do you want to be involved in this update? The pull request should be as simple as updating the tags from where we pull containerd and runc https://github.com/ubuntu/microk8s/blob/feature/containerd-access/build-scripts/prepare-env.sh#L19 . After this update I will build and release the snap for testing.

I would appreciate any help I could get during the testing phase. Would you also be interested in submitting the PR?

Thanks

waquidvp:
I will work on this PR if you don't mind...

waquidvp:
When I build locally to test my changes, should I be updating `KUBE_VERSION` to something like `1.14.0-beta.1` since this PR is aimed to release the `1.14.0` Kubernetes version. I also noticed that in this PR, `KUBE_VERSION` is still set at `1.13.4`, sould this be changed as beta releases for 1.14.0 start to come out, and I am guessing will be changed to `1.14.0` once that is released and ready.

waquidvp:
I also would like to try out a newer version of the prometheus-operator to see if it solves some of the problems I have been having with it. Do you think that should go in part of this PR and target it for the 1.14.0 release?

ktsakalozos:
> should I be updating `KUBE_VERSION` to something like `1.14.0`

Our CI sets this to the right value as soon at it branches the code. So yes, set it to `1.14.0` so it does not conflict with the master branch upon merge.


> try out a newer version of the prometheus-operator

I would prefer to have a separate PR.

ktsakalozos:
@waquidvp when you build MicroK8s localy make sure you do so with `snapcraft cleanbuild` that uses LXD under the hood.

waquidvp:
So I have been using `snapcraft cleanbuild` for building and testing, but when doing something like `KUBE_VERSION=1.14.0-beta.1 snapcraft cleanbuild` it doesn't use the env variable that I just set. I am guessing this is something to do with the env variables not being passed into the lxc. Is there any way to pass env vars to the lxd running snapcraft.

ktsakalozos:
@waquidvp to build 1.14.0-beta I had to:

```
lxc launch ubuntu:16.04 --ephemeral test-build
lxc exec test-build -- snap install snapcraft --classic
lxc exec test-build -- apt update
lxc exec test-build -- git clone https://github.com/ubuntu/microk8s project
lxc exec test-build -- sh -c "cd project && KUBE_VERSION=1.14.0-beta.1 snapcraft"
```
Then to get the `.snap` file out of the container:
```
lxc file pull test-build/root/microk8s-small/microk8s_v1.14.0-beta.1_amd64.snap .
```

waquidvp:
I have created a new PR so that we can continue our discussion about the containerd update there: https://github.com/ubuntu/microk8s/pull/350

waquidvp:
When debugging issues with containerd, the `ctr` command is useful, but it will be more useful to have `crictl` aswell, I could create a new PR to add it but should it be scoped into this PR?

ktsakalozos:
> I could create a new PR to add it but should it be scoped into this PR?

Adding `crictl` would have been awesome. If you submit a PR it will be included in this branch. Many thanks.

docs/ports.md
```diff
@@ -0,0 +1,58 @@
...
+
+Service | Socket
+--- | ---
+docker | unix:///var/snap/microk8s/current/docker.sock
+etcd | unix://etcd.socket:2379
+
+
+## Authentication and Authorization
+
+Upon a new deployment MicroK8s creates a new CA, a signed server certificate and a service account key file. These files are stored under `/var/microk8s/current/certs`. Kubelet and the API server are aware of the same CA and so the signed server certificate is used by the API server to authenticate with kubelet (`--kubelet-client-certificate`). Clients talking to the secure port of the API server (`6443`) have to also be aware of the CA (`certificate-authority-data` in user kubeconfig).
```
waquidvp: Here it refers to `6443` as the port for the secure API server, but this doesn't seem to work for me. So further investigation shows that the secure port is actually `16443`, should this be updated?

ktsakalozos: Thank you for spotting this. Fixed.

ktsakalozos:
Here is goes!

PR #328, opened 2019-02-21
Use the correct ConfigMap for nginx ingress

keshavdv:
By default, the ingress installation seems to use a non-existing ConfigMap and having another one that exists but isn't used is confusing. This switches the controller to use the one that microk8s creates.


ktsakalozos:
Thank you fro the PR @keshavdv. Would you mind rebasing so we can get a green build from travis.

keshavdv:
Done!

ktsakalozos:
Thank you! The fix should be available in 1.13.5+

PR #350, opened 2019-03-08
Use newer containerd and runc versions

waquidvp:
This change uses a newer containerd and runc mainly so that we have the fix for CVE-2019-5736.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ sudo: required
 services: [docker]
 
 script:
-  - docker run -v $(pwd):$(pwd) -w $(pwd) snapcore/snapcraft:stable sh -c "apt update && snapcraft"
+  - docker run -v $(pwd):$(pwd) -w $(pwd) snapcore/snapcraft:beta sh -c "apt update && snapcraft"
   - sudo apt update
   - sudo apt install --yes snapd
   - sudo snap install *.snap --classic --dangerous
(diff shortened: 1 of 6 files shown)


waquidvp:
I have been trying to get this change to work for about 2 days now and have had no luck, I am creating this PR so that someone else could potentially help me out.

waquidvp:
I would also like to point out that I was using `KUBE_VERSION=1.14.0-beta.1` when building.

waquidvp:
This could be to do with some configuration change that is required somewhere in `default-args` for one of the components, I am not sure.

waquidvp:
Update: After rebuilding with these changes but without setting `KUBE_VERSION` (so the latest stable `v1.13.4`), I got some services running (dns and prometheus). I then removed this snap and rebuilt again but now with `KUBE_VERSION` set to `v1.14.0-beta.1` and installed it on the same machine, and now runnning the same services (dns and prometheus) seems to work. I will try to install and run this on a new and fresh machine to see if I get the same problem as described here: https://github.com/ubuntu/microk8s/pull/350#issuecomment-470917197.

waquidvp:
I have now tested a build with `KUBE_VERSION=v1.14.0-beta.1` using the method described here: https://github.com/ubuntu/microk8s/pull/323#issuecomment-470826769. I have tried it on many machines (enabled/disabled many services) and seems to work. I have merged master into this branch to get the newer snapcraft docker image so that the CI can build and continue on to testing.

waquidvp:
On a side note, I think what I was experiencing here https://github.com/ubuntu/microk8s/pull/350#issuecomment-470917197 was to do with not using `snapcraft cleanbuild`.

ktsakalozos:
Thank you @waquidvp . It looks great.

Issue #157, opened 2018-11-01
Add configuration options (size, location) to microk8s.enable registry.


carmine:
At present, the registry is constrained to 20Gb and to a specific path in the local environment. It'd be great if both were configurable. The use case:

- Using the registry in anger, you'll exceed 20GB quickly, particularly for ML images
- If microk8s/registry is running in a VM that is ephemeral (may be destroyed), it'd be great to pass in a mounted location so that the registry contents will survive from vm to vm.

tvansteenburgh:
@ktsakalozos What are your thoughts on this?

ktsakalozos:
There have been some private conversations with @carmine on this topic.

We have two asks here. First how to add parameters in our addons. Second how to reuse storage volumes.

I do not have a good and clean answer for the second ask. I am not sure if we should handle this with the host-provisioner or follow the upstream solutions.

Here are my thoughts on adding parameters to addons. Imagine that you have a properties ini file that would look like this:

```
[registry]
size=20GB
claim=bar

[storage]
dir=${snap_data}/default-storage
```
Could we change the default registry properties with something like:
```
microk8s.configure registry --size=30GB
microk8s.enable registry
```

or 

```
microk8s.create myregistryconf --from registry
microk8s.configure myregistryconf --size=30GB
```

That would add a
```
[myregistryconf]
size=30GB
claim=bar
```

and then do 

```
microk8s.enable registry --with myregistryconf
```

This apprach would be generic for all addons we have.

carmine:
I think the ability to provide a conf would be great, but also the ability to specify key settings in the enable. Downside to this suggestion is that the "key settings" list could grow. 

If this is more of a storage issue .. then I would start with storage. Something like:

microk8s.enable storage --name class_name --size [X] --location /some/path --default true

From there, I could then do something similar with registry, specifying the storage class and pvc size.

PR #352, opened 2019-03-09
 Feature - linkerd addon.

balchua:
This feature allows enabling linkerd from microk8s.  Linkerd version is 2.2.0


build-scripts/fetch-binaries.sh
```diff
@@ -42,4 +42,15 @@ echo $KUBE_VERSION > $KUBE_SNAP_BINS/version
     mv istio-${ISTIO_ERSION}/install/kubernetes/istio-demo-auth.yaml ./istio-yaml/
     mv istio-${ISTIO_ERSION}/install/kubernetes/istio-demo.yaml ./istio-yaml/
   fi
+
+  #Linkerd on Linux
+  if [ "$KUBE_ARCH" = "amd64" ]
```
waquidvp: Could you not reuse the if statement used in line 32?
```suggestion
  # Istio and Linkerd is only released on amd64
  if [ "$KUBE_ARCH" = "amd64" ]
```

balchua: I can combine istio and linkerd in this case.

waquidvp:
Since fluentd is for amd64 arch, you probably should remove the enable/disable scripts from the actions in the case of arm64, this is done for example with istio at https://github.com/ubuntu/microk8s/blob/master/snapcraft.yaml#L189

waquidvp:
Also would be useful to configure the `LINKERD_VERSION` with an env var. Have it default to `2.2.0` but let it be configurable, here is an example with istio: https://github.com/ubuntu/microk8s/blob/master/build-scripts/prepare-env.sh#L16, and also print it here: https://github.com/ubuntu/microk8s/blob/master/build-scripts/prepare-env.sh#L38

balchua:
Suddenly i get this error at the end of travis build

`Download snap "microk8s" (455) from channel "edge" (received an unexpected http response code (503) when trying to download https://fastly.cdn.snapcraft.io/download-origin/fastly/EaXqgt1lyCaxKaQCU349mlodBkDCXRcg_455.snap?token=1552197600_54dbb3e4cc3838fa6f4fc07e812ec7113e35d15e)`

ktsakalozos:
That was a transient error @balchua. Seems ok now.

balchua:
Thanks!

build-scripts/fetch-binaries.sh
```diff
@@ -41,5 +41,14 @@ echo $KUBE_VERSION > $KUBE_SNAP_BINS/version
     mv istio-${ISTIO_ERSION}/install/kubernetes/helm/istio/templates/crds.yaml ./istio-yaml/
     mv istio-${ISTIO_ERSION}/install/kubernetes/istio-demo-auth.yaml ./istio-yaml/
     mv istio-${ISTIO_ERSION}/install/kubernetes/istio-demo.yaml ./istio-yaml/
+    
+    #Linkerd on Linux    
+    echo "Fetching linkerd binary."
```
ktsakalozos: Can we move the downloading of the linkerd binary in the wrapper (`microk8s-resources/wrappers/microk8s-linkerd.wrapper`). The wrapper would check that the binary is available and if not it would try to download it in $SNAP_DATA/bin/.

The reason for this ask is that downloading and packaging binaries inside the MicroK8s snap does not scale. The snap will grow in size too much too quickly. One of the strong points of MicroK8s is that it deploys fast and this deployment time is dominated by the time needed to download the snap from the snap store. Addons is a convenient extension point. We welcome everyone to write addon scripts but we should not force everyone to pay the penalty of downloading binaries that may never be used.

I know that this is not what we do in the case of istio. We have to change the istio wrapper too to comply with the approach above (https://github.com/ubuntu/microk8s/issues/363).

ktsakalozos:
Thank you for the PR @balchua. Looks great, solid work! There is only one comment I would like addressed before accepting it.

@waquidvp thank you for the review.

balchua:
Thanks.  @ktsakalozos good idea.  I will move the downloading of linkerd binary at the time of enabling the addon.
I have some question though, current version of linkerd allows the user to auto inject the proxy very much like what istio does today.
I was thinking of providing some user option to enable the auto proxy inject.
What do you think?

ktsakalozos:
Thanks @balchua, the enable command sounds fine for grabbing the linkerd binary.

How would the user experience look like in the case of this extra option. For istio we ask the user during the `microk8s.enable istio` action. The drawback with this approach is that you cannot include `microk8s.enable istio` easily in scripts. We are considering the option of adding configuration parameters (https://github.com/ubuntu/microk8s/issues/157#issuecomment-437310716) but we are open to any suggestions.

balchua:
I was thinking of adding arguments to enable command.  But at the same time not too many knobs to turn on/off.

ktsakalozos:
Arguments on the enable command may not work well as the users are able to queue the enabling of more than one addons, eg `microk8s.enable dns dashboard linkerd`. How would these arguments look like in this context?

balchua:
Just realized that users can enable multiple addons in one go.  
In this context it would look like `microk8s.enable linkerd --auto-inject-on dns dashboard`. But then the user experience will be strange..

balchua:
Was just thinking out loud.  What if the args are specified `microk8s.enable linkerd:auto-inject=on;other-opts=xx dns dashboard`.

ktsakalozos:
> microk8s.enable linkerd:auto-inject=on;other-opts=xx dns dashboard

This look good. Please put the arguments parsing in `utils.sh` (https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/common/utils.sh) so we can use it in both the enable and disable wrappers. Thank you.

balchua:
Linkerd addon with optional argument to pass during enable.  it looks like this now `microk8s.enable linkerd:--proxy-auto-inject`  On reset, `$SNAP_DATA/bin` is deleted.  In the future, addon binaries will automatically be cleaned up upon reset.  When disabling the addon, the linkerd binary is also deleted.

README.md
```diff
@@ -92,6 +92,7 @@ With `microk8s.status` you can see the list of available addons and which ones a
 - **prometheus**: Deploy the [Prometheus Operator](https://github.com/coreos/prometheus-operator) v0.25.
 - **fluentd**: Deploy [Elasticsearch-Kibana-Fluentd](https://kubernetes.io/docs/tasks/debug-application-cluster/logging-elasticsearch-kibana/) logging and monitoring solution.
 - **jaeger**: Deploy the [Jaeger Operator](https://github.com/jaegertracing/jaeger-operator) v1.8.2 in the "simplest" configuration.
+- **linkerd**: Deploy linkerd2 [Linkerd](https://linkerd.io/2/overview/) service mesh.  By default proxy auto inject is not enabled. , To enable auto proxy injection, simply use `microk8s.enable linkerd:--proxy-auto-inject`.
```
ktsakalozos: Can you please remove the extra ",".

Can we drop the "--" in `linkerd:--proxy-auto-inject`

build-scripts/fetch-binaries.sh
```diff
@@ -41,5 +41,7 @@ echo $KUBE_VERSION > $KUBE_SNAP_BINS/version
     mv istio-${ISTIO_ERSION}/install/kubernetes/helm/istio/templates/crds.yaml ./istio-yaml/
     mv istio-${ISTIO_ERSION}/install/kubernetes/istio-demo-auth.yaml ./istio-yaml/
     mv istio-${ISTIO_ERSION}/install/kubernetes/istio-demo.yaml ./istio-yaml/
+
```
ktsakalozos: Can we not add empty lines?

build-scripts/fetch-binaries.sh
```diff
@@ -41,5 +41,7 @@ echo $KUBE_VERSION > $KUBE_SNAP_BINS/version
     mv istio-${ISTIO_ERSION}/install/kubernetes/helm/istio/templates/crds.yaml ./istio-yaml/
     mv istio-${ISTIO_ERSION}/install/kubernetes/istio-demo-auth.yaml ./istio-yaml/
     mv istio-${ISTIO_ERSION}/install/kubernetes/istio-demo.yaml ./istio-yaml/
+    
   fi
+
```
ktsakalozos: Same here

build-scripts/prepare-env.sh
```diff
@@ -20,6 +20,7 @@ export CONTAINERD_COMMIT="${CONTAINERD_COMMIT:-bb71b10fd8f58240ca47fbb579b9d1028
 # Release v1.0.0~rc6 with CVE-2019-5736 fix
 export RUNC_COMMIT="${RUNC_COMMIT:-2b18fe1d885ee5083ef9f0838fee39b62d653e30}"
 
+
```
ktsakalozos: Same here

microk8s-resources/actions/common/utils.sh
```diff
@@ -73,6 +73,33 @@ use_manifest() {
...
+    # Extracts the addon arguments.
+    # Example: enable linkerd:proxy-auto-inject=on;other-args=xyz
+    # Parameter:
+    #   $1 the addon arguments in array
+    # Returns:
+    #   add-on arguments array
+    local IFS=':'
+    read -ra ADD_ON <<< "$1"
+    local IFS=';'
+    read -ra ARGUMENTS <<< "${ADD_ON[1]}"
```
ktsakalozos: This does not work for more than one arguments, see for example:
```
$ microk8s.enable linkerd:arg1;arg2
Enabling Linkerd2
Enabling DNS
Applying manifest
service/kube-dns unchanged
serviceaccount/kube-dns unchanged
configmap/kube-dns unchanged
deployment.extensions/kube-dns configured
Restarting kubelet
DNS is enabled
namespace/linkerd unchanged
serviceaccount/linkerd-controller unchanged
clusterrole.rbac.authorization.k8s.io/linkerd-linkerd-controller unchanged
clusterrolebinding.rbac.authorization.k8s.io/linkerd-linkerd-controller unchanged
serviceaccount/linkerd-prometheus unchanged
clusterrole.rbac.authorization.k8s.io/linkerd-linkerd-prometheus unchanged
clusterrolebinding.rbac.authorization.k8s.io/linkerd-linkerd-prometheus unchanged
service/linkerd-controller-api unchanged
service/linkerd-proxy-api unchanged
deployment.extensions/linkerd-controller configured
customresourcedefinition.apiextensions.k8s.io/serviceprofiles.linkerd.io unchanged
serviceaccount/linkerd-web unchanged
service/linkerd-web unchanged
deployment.extensions/linkerd-web configured
service/linkerd-prometheus unchanged
deployment.extensions/linkerd-prometheus configured
configmap/linkerd-prometheus-config unchanged
serviceaccount/linkerd-grafana unchanged
service/linkerd-grafana unchanged
deployment.extensions/linkerd-grafana configured
configmap/linkerd-grafana-config unchanged
Linkerd is starting

Command 'arg2' not found, did you mean:

  command 'args' from deb ucommon-utils

Try: sudo apt install <deb name>


```

microk8s-resources/actions/enable.linkerd.sh
```diff
@@ -0,0 +1,28 @@
...
+  sudo chmod uo+x "$SNAP_DATA/bin/linkerd"
+
+fi 
+
+echo "Enabling Linkerd2"
+
+# pod/servicegraph will start failing without dns
+"$SNAP/microk8s-enable.wrapper" dns
+
+"$SNAP_DATA/bin/linkerd" install "$*" | "$SNAP/kubectl" "--kubeconfig=$SNAP_DATA/credentials/client.config" apply -f -
```
ktsakalozos: Could we be more precise in what arguments we allow? Seems we are passing all arguments passed but this may have bad implications.

snapcraft.yaml
```diff
@@ -251,6 +256,7 @@ parts:
         ./istioctl collateral completion --bash -o .
         sed -i  "s/complete -o default -F __start_istioctl istioctl/complete -o default -F __start_istioctl microk8s.istioctl/g" istioctl.bash
         sed -i  "s/complete -o default -F __start_istioctl istioctl/complete -o default -o nospace -F __start_istioctl microk8s.istioctl/g" istioctl.bash
+
```
ktsakalozos: Lets not add empty lines.

README.md
balchua: My thoughts here was to allow the same arguments as what the original linkerd arguments allow without too much limitations, hence the "--".

build-scripts/fetch-binaries.sh
balchua: No problem with this one.

microk8s-resources/actions/common/utils.sh
balchua: Yes this one, the issue here is that the shell treated ; as different commands.  For this kind of scenario it has to be enclosed in quotes.

microk8s-resources/actions/enable.linkerd.sh
balchua: My initial thought was to allow only certain args to be passed, but then i realized that certain addons, in this case linkerd, allows several more args that can be passed while installing linkerd.  Such as mutual tls on the data plane... And maybe some other args in the future...

balchua:
@ktsakalozos, i am trying to work out the items you highlighted especially the one limiting the arguments passed to linkerd.  I would like to allow the user to have as much as possible similar experience as using the native `linkerd install` command during linkerd enable..  that's why i didn't try to limit the args passed during enable.  Still some work to do here by removing `--` on the args and implicitly add them inside the wrapper.  This will force `linkerd`  to disallow arguments unknown to it. What do you think? Thanks.

balchua:
Another thing i noticed is that when using `$SNAP_DATA` the user needs to have permission to it, hence it requires sudo.  What about using `$SNAP_USER_DATA/bin` to keep user binaries?

ktsakalozos:
> This will force `linkerd` to disallow arguments unknown to it. What do you think?

Sounds good to me.

> What about using `$SNAP_USER_DATA/bin` to keep addon binaries?

I do not know about that. What if there are many users on the machine?

balchua:
The documentation wasn't very clear about multiple users.

From the [documentation](https://docs.snapcraft.io/environment-variables/7983) 
`SNAP_USER_DATA`
Directory for user data.

This directory is backed up and restored across snap refresh and snap revert operations.

Typical value: `/home/zyga/snap/hello-world/27`

The final number there is `$SNAP_REVISION`

I guess it wouldn't work for other users in the same system.

ktsakalozos:
Thank you for this PR @balchua and for your patience!

Issue #261, opened 2018-12-22
microk8s.reset does not delete CRDS


dmarkey:
Pretty minor but CRDs that are defined aren't deleted when doing a reset- pretty minor

PR #355, opened 2019-03-13
fix: delete CRDs on reset

miguelgarcia:
This change removes installed CRDs when microk8s.reset is executed. I find it useful, particularly for people playing with CRDs and Custom Operators, like myself.

See #261


ktsakalozos:
LGTM +1. Thank you for the PR @miguelgarcia. It should be available with the next patch release.

Issue #370, opened 2019-03-22
/snap/microk8s/442/usr/bin/gawk: error while loading shared libraries: libreadline.so.6


khteh:
Please run `microk8s.inspect` and attach the generated tarball to this issue.
[inspection-report-20190322_115821.tar.gz](https://github.com/ubuntu/microk8s/files/2994873/inspection-report-20190322_115821.tar.gz)

We appreciate your feedback. Thank you for using microk8s.
After I have done `snap install microk8s --channel=1.14/beta --classic` and try to change default port, this is what I get:

`/snap/microk8s/442/usr/bin/gawk: error while loading shared libraries: libreadline.so.6: cannot open shared object file: No such file or directory`

ktsakalozos:
Can you please tell me what exactly did you do to change the default port? Please be as descriptive as possible so I can reproduce the error.

ktsakalozos:
`microk8s.config` throws this. I am able to reproduce it. Should have a fix for this soon. Thank you for reporting it here.

Issue #371, opened 2019-03-22
Do some packages use libs from the host?


ktsakalozos:
The following packages are in a snapcraft part marked with `no-patchelf`. We need to investigate if this allows them to escape the "confinement" we have:
    - net-tools
    - util-linux
    - zfsutils-linux
    - socat
    - iproute2
    - dpkg

PR #372, opened 2019-03-22
Fix packages that need patchelf

ktsakalozos:
These libraries if pot patchelfed use libraries from the host and may fail if the used libs are missing. 

Fixes: https://github.com/ubuntu/microk8s/issues/370

We need to investigate if the rest of the libraries need to be patchelfed: https://github.com/ubuntu/microk8s/issues/371


ktsakalozos:
Merging this to get new snaps.

PR #398, opened 2019-04-01
Add caching of juju images if available.

Veebers:
If Juju is installed when Microk8s is installed or refreshed cache the operator images that Juju uses for Caas.

There is similar work going into the Juju snap that checks for Microk8s and caches if found, as well as work going into Juju that will treat Microk8s as a built in k8s provider (much like how LXD is if installed on the system).

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -117,3 +117,11 @@ produce_server_cert() {
     openssl req -new -key ${SNAP_DATA}/certs/server.key -out ${SNAP_DATA}/certs/server.csr -config ${SNAP_DATA}/certs/csr.conf
     openssl x509 -req -in ${SNAP_DATA}/certs/server.csr -CA ${SNAP_DATA}/certs/ca.crt -CAkey ${SNAP_DATA}/certs/ca.key -CAcreateserial -out ${SNAP_DATA}/certs/server.crt -days 100000 -extensions v3_ext -extfile ${SNAP_DATA}/certs/csr.conf
 }
+
+maybe_cache_juju_operator_images() {
+    if [ ! -f /snap/bin/juju.cache-images ]; then
+        echo "No supported version of Juju installed."
+        exit 0
+    fi
+    /snap/bin/juju.cache-images
+}
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -99,3 +99,10 @@ then
     systemctl restart snap.${SNAP_NAME}.daemon-apiserver.service
     systemctl restart snap.${SNAP_NAME}.daemon-proxy.service
 fi
+
+# Cache Juju image if juju is present and this is an install/refresh
+if [ ! -e ${SNAP_DATA}/.cache-juju-image ]
+then
+   maybe_cache_juju_operator_images >> /tmp/mk8s.log
+   touch ${SNAP_DATA}/.cache-juju-image
+fi
--- a/snap/hooks/install
+++ b/snap/hooks/install
@@ -32,3 +32,4 @@ chmod 644 ${SNAP_DATA}/credentials/client.config
 $SNAP/bin/sed -i 's/PASSWORD/'"${admin_token}"'/g' ${SNAP_DATA}/credentials/client.config
 ca_data=$(cat ${SNAP_DATA}/certs/ca.crt | ${SNAP}/usr/bin/base64 -w 0)
 $SNAP/bin/sed -i 's/CADATA/'"${ca_data}"'/g' ${SNAP_DATA}/credentials/client.config
+
--- a/snap/hooks/pre-refresh
+++ b/snap/hooks/pre-refresh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+set -eux
+
+# Ensure juju images get cached if needed on refresh
+rm -fr ${SNAP_DATA}/.cache-juju-image


microk8s-resources/actions/common/utils.sh
```diff
@@ -117,3 +117,11 @@ produce_server_cert() {
     openssl req -new -key ${SNAP_DATA}/certs/server.key -out ${SNAP_DATA}/certs/server.csr -config ${SNAP_DATA}/certs/csr.conf
     openssl x509 -req -in ${SNAP_DATA}/certs/server.csr -CA ${SNAP_DATA}/certs/ca.crt -CAkey ${SNAP_DATA}/certs/ca.key -CAcreateserial -out ${SNAP_DATA}/certs/server.crt -days 100000 -extensions v3_ext -extfile ${SNAP_DATA}/certs/csr.conf
 }
+
+maybe_cache_juju_operator_images() {
+    if [ ! -f /snap/bin/juju.cache-images ]; then
```
ktsakalozos: I remember I read somewhere that `/snap` is configurable, but I cannot find a link now. Is there an env variable we could use or could you please point me to some snapcraft doc saying `/snap/` is not configurable?

microk8s-resources/actions/common/utils.sh
```diff
@@ -117,3 +117,11 @@ produce_server_cert() {
...
+        echo "No supported version of Juju installed."
+        exit 0
+    fi
+    /snap/bin/juju.cache-images
```
ktsakalozos: What is the cost of this? How much data are we downloading at this point?

snap/hooks/configure
```diff
@@ -99,3 +99,10 @@ then
     systemctl restart snap.${SNAP_NAME}.daemon-apiserver.service
     systemctl restart snap.${SNAP_NAME}.daemon-proxy.service
 fi
+
+# Cache Juju image if juju is present and this is an install/refresh
+if [ ! -e ${SNAP_DATA}/.cache-juju-image ]
+then
+   maybe_cache_juju_operator_images >> /tmp/mk8s.log
```
ktsakalozos: Could we place the log in side "$SNAP_DATA" or "$SNAP_COMMON"? It will make our life easier if we decide to go with strict confinement?

microk8s-resources/actions/common/utils.sh
ktsakalozos: The entire MicroK8s is 200MB. We would like to keep this number low as it dominates the installation time.

ktsakalozos:
Thank you for the PR @Veebers , where can I read more on the design of this effort?

microk8s-resources/actions/common/utils.sh
Veebers: oh, good point. I'll double check for an env var to use.

snap/hooks/configure
Veebers: d'oh, this is left over cruft for when I was testing/debugging. I could log that to a file in DATA/COMMON if wanted though.

microk8s-resources/actions/common/utils.sh
Veebers: It's about 200MB for the 2 images. I can understand the apprehension adding to the install size. Let me find some shareable details around why we want to do this :-)

ktsakalozos: This doubles the deployment time. Obviously not good.

Why do we flip the switch of caching during installation and don't we flip the switch with a `microk8s.enable juju-integration`. This would be the same as with any other addon, I am also pretty sure we already expect a few addons to be available (eg dns storage) so it is not surprising for the user.

snap/hooks/configure
ktsakalozos: I am fine with not logging it.

microk8s-resources/actions/common/utils.sh
ktsakalozos: There is also the option of an auto connect interface. I am afraid this approach of inferring the existence of juju through the /snap/bin may not work when we move to strict confinement.

microk8s-resources/actions/common/utils.sh
chipaca: FWIW classic snaps won't install unless `/snap/` exists, either as a directory or a symlink.

ktsakalozos: This is safe then.

microk8s-resources/actions/common/utils.sh
ktsakalozos: Could this `juju.cache-images` be run on the background?

microk8s-resources/actions/common/utils.sh
```diff
@@ -117,3 +117,11 @@ produce_server_cert() {
     openssl req -new -key ${SNAP_DATA}/certs/server.key -out ${SNAP_DATA}/certs/server.csr -config ${SNAP_DATA}/certs/csr.conf
     openssl x509 -req -in ${SNAP_DATA}/certs/server.csr -CA ${SNAP_DATA}/certs/ca.crt -CAkey ${SNAP_DATA}/certs/ca.key -CAcreateserial -out ${SNAP_DATA}/certs/server.crt -days 100000 -extensions v3_ext -extfile ${SNAP_DATA}/certs/csr.conf
 }
+
+maybe_cache_juju_operator_images() {
+    if [ ! -f /snap/bin/juju.cache-images ]; then
+        echo "No supported version of Juju installed."
+        exit 0
```
ktsakalozos: You will exit everything here?! What if you are in the snap install process?

ktsakalozos:
Closing this PR, I guess it will be addressed in other ways.

Issue #402, opened 2019-04-02
Install error with utils.sh


falsandtru:
I succeeded to install microk8s several months ago but now I can't.

OS: Ubuntu 18.04

```
$ sudo snap install microk8s --classic
2019-04-02T10:16:17+09:00 INFO Waiting for restart...
error: cannot perform the following tasks:
- Run install hook of "microk8s" snap if present (run hook "install": 
-----
+ export LD_LIBRARY_PATH=/snap/microk8s/492/lib:/snap/microk8s/492/usr/lib:/snap/microk8s/492/lib/x86_64-linux-gnu:/snap/microk8s/492/usr/lib/x86_64-linux-gnu
+ LD_LIBRARY_PATH=/snap/microk8s/492/lib:/snap/microk8s/492/usr/lib:/snap/microk8s/492/lib/x86_64-linux-gnu:/snap/microk8s/492/usr/lib/x86_64-linux-gnu
+ export PATH=/snap/microk8s/492/usr/sbin:/snap/microk8s/492/usr/bin:/snap/microk8s/492/sbin:/snap/microk8s/492/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/usr/bin:/usr/local/bin
+ PATH=/snap/microk8s/492/usr/sbin:/snap/microk8s/492/usr/bin:/snap/microk8s/492/sbin:/snap/microk8s/492/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/usr/bin:/usr/local/bin
+ source /snap/microk8s/492/actions/common/utils.sh
+ cp -r /snap/microk8s/492/default-args /var/snap/microk8s/492/args
+ mkdir /var/snap/microk8s/492/certs
+ openssl genrsa -out /var/snap/microk8s/492/certs/serviceaccount.key 2048
Generating RSA private key, 2048 bit long modulus
................................+++
......................................................+++
unable to write 'random state'
e is 65537 (0x010001)
+ openssl genrsa -out /var/snap/microk8s/492/certs/ca.key 2048
Generating RSA private key, 2048 bit long modulus
...............................+++
....+++
unable to write 'random state'
e is 65537 (0x010001)
[...]

Issue #400, opened 2019-04-01
Ubuntu 18.04 installation issue


rafaelsoaresbr:
Please run `microk8s.inspect` and attach the generated tarball to this issue.
N/A
We appreciate your feedback. Thank you for using microk8s.

I'm trying to install microk8s with
```
sudo snap install microk8s --classic
```
on a fresh installation of Ubuntu 18.04 (command-line install) and got the following error:
```
error: cannot perform the following tasks:
- Run install hook of "microk8s" snap if present (run hook "install": 
-----
+ export LD_LIBRARY_PATH=/snap/microk8s/492/lib:/snap/microk8s/492/usr/lib:/snap/microk8s/492/lib/x86_64-linux-gnu:/snap/microk8s/492/usr/lib/x86_64-linux-gnu
+ LD_LIBRARY_PATH=/snap/microk8s/492/lib:/snap/microk8s/492/usr/lib:/snap/microk8s/492/lib/x86_64-linux-gnu:/snap/microk8s/492/usr/lib/x86_64-linux-gnu
+ export PATH=/snap/microk8s/492/usr/sbin:/snap/microk8s/492/usr/bin:/snap/microk8s/492/sbin:/snap/microk8s/492/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/usr/bin:/usr/local/bin
+ PATH=/snap/microk8s/492/usr/sbin:/snap/microk8s/492/usr/bin:/snap/microk8s/492/sbin:/snap/microk8s/492/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/usr/bin:/usr/local/bin
+ source /snap/microk8s/492/actions/common/utils.sh
+ cp -r /snap/microk8s/492/default-args /var/snap/microk8s/492/args
+ mkdir /var/snap/microk8s/492/certs
+ openssl genrsa -out /var/snap/microk8s/492/certs/serviceaccount.key 2048
Generating RSA private key, 2048 bit long modulus
............+++
.................................................+++
unable to write 'random state'
e is 65537 (0x010001)
+ openssl genrsa -out /var/snap/microk8s/492/certs/ca.key 2048
Generating RSA private key, 2048 bit long modulus
...........................+++
...............+++
unable to write 'random state'
e is 65537 (0x010001)
[...]

Issue #395, opened 2019-03-29
Unable to install last version (492)


franxois:
Hello, I'm unable to install last stable version of microk8s on my ubuntu 18.10.
The installation process finish with :

```
Error Loading request extension section req_ext
140600890205248:error:220A4076:X509 V3 routines:a2i_GENERAL_NAME:bad ip address:../crypto/x509v3/v3_alt.c:457:value=adr:192.168.1.19                                           
140600890205248:error:22098080:X509 V3 routines:X509V3_EXT_nconf:error in extension:../crypto/x509v3/v3_conf.c:47:name=subjectAltName, value=@alt_names 
```

I allready had microk8s but I was unable to upgrade from microk8s 1.14 (483) to 492. `snap refresh`  was giving me the same error. I decided to remove and reinstall mikrok8s stable but I had the same error. 

Using channel 1.13/stable is working.

ktsakalozos:
Hi @franxois 

During deployment we try to identify the machine's IP so we can issue the right certificates. The IP discovery is done by these two lines of code: https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/common/utils.sh#L97 and the certificate issue is here: https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/common/utils.sh#L117

Could you please share the output of  `netstat -rn` and `ifconfig`?

Thanks

diegolakatos:
Hello, I'm having the same issue with version 492 and channel 1.13/stable is working.

```
netstat -rn                                               
Tabela de Roteamento IP do Kernel
Destino         Roteador        MáscaraGen.    Opções   MSS Janela  irtt Iface
0.0.0.0         192.168.1.1     0.0.0.0         UG        0 0          0 wlp3s0
169.254.0.0     0.0.0.0         255.255.0.0     U         0 0          0 docker0
172.17.0.0      0.0.0.0         255.255.0.0     U         0 0          0 docker0
192.168.1.0     0.0.0.0         255.255.255.0   U         0 0          0 wlp3s0
```
```
ifconfig
docker0: flags=4099<UP,BROADCAST,MULTICAST>  mtu 1500
        inet 172.17.0.1  netmask 255.255.0.0  broadcast 172.17.255.255
        inet6 fe80::42:8cff:fe81:21e0  prefixlen 64  scopeid 0x20<link>
        ether 02:42:8c:81:21:e0  txqueuelen 0  (Ethernet)
[...]

PR #404, opened 2019-04-02
Use ip and not ifconfig for IP detection

ktsakalozos:
Seems ifconfig output is locale sensitive so we use ip to extract the IP.

We also improve the case where no IP is detected.

Fixes: https://github.com/ubuntu/microk8s/issues/402 https://github.com/ubuntu/microk8s/issues/395 https://github.com/ubuntu/microk8s/issues/400


falsandtru:
When can I get a new release from snap? I don't know the release cycle of snap packages.

ktsakalozos:
@falsandtru the commit just produced a snap available on edge. Would you please give it a try with:
```
sudo snap install microk8s --edge --classic
```

Thanks

falsandtru:
I could refresh, thanks!

Issue #253, opened 2018-12-18
Elasticsearch stable Helm charts max file descriptors 


novakov-alexey:
Stable Helm Elasticsearch charts https://github.com/helm/charts/tree/master/stable/elasticsearch
fail on Elasticsearch nodes startup:

```
[2018-12-18T22:11:25,113][INFO ][o.e.b.BootstrapChecks    ] [es-elasticsearch-master-0] bound or publishing to a non-loopback address, enforcing bootstrap checks
ERROR: [1] bootstrap checks failed
[1]: max file descriptors [4096] for elasticsearch process is too low, increase to at least [65536]
```

microk8s.inspect output:

```
Inspecting services
  Service snap.microk8s.daemon-docker is running
  Service snap.microk8s.daemon-apiserver is running
  Service snap.microk8s.daemon-proxy is running
  Service snap.microk8s.daemon-kubelet is running
  Service snap.microk8s.daemon-scheduler is running
  Service snap.microk8s.daemon-controller-manager is running
  Service snap.microk8s.daemon-etcd is running
  Copy service arguments to the final report tarball
Inspecting AppArmor configuration
Gathering system info
  Copy network configuration to the final report tarball
  Copy processes list to the final report tarball
  Copy snap list to the final report tarball
  Inspect kubernetes cluster

 WARNING:  IPtables FORWARD policy is DROP. Consider enabling traffic forwarding with: sudo iptables -P FORWARD ACCEPT
Building the report tarball
  Report tarball is at /var/snap/microk8s/340/inspection-report-20181218_231630.tar.gz
```

On host:
```
cat /etc/sysctl.conf
vm.max_map_count = 262144
fs.file-max = 1634038
```

ktsakalozos:
Hi @novakov-alexey 

Are you able to set an environment variable `discovery.type` to `single-node` as suggested https://stackoverflow.com/questions/42300463/elasticsearch-5-x-bootstrap-checks-failing and we do in this PR https://github.com/ubuntu/microk8s/pull/217/files#diff-a74758cc33ad9bbd62f6a3d900aa3ee1R95

Thank you for trying out MicroK8s

novakov-alexey:
Hi @ktsakalozos 

Thanks for quick reply. 

Setting `discovery.type` to `single-node` is more or less workaround. I want to start more than one Elasticsearch node on the Microk8s for testing purposes. When using single-node config, Elasticsearch won't build proper cluster 😞 

[...]

PR #413, opened 2019-04-05
Set the maximum number of open file descriptors to 65k for containers

ktsakalozos:
Fixes https://github.com/ubuntu/microk8s/issues/253


ktsakalozos:
Got confirmation from the users that this fix works. So merging this.

Issue #415, opened 2019-04-05
failed to access from outside VM: Unable to connect to the server: x509: certificate is valid for 127.0.0.1...


iwanbk:
Please run `microk8s.inspect` and attach the generated tarball to this issue.

[inspection-report-20190405_143043.tar.gz](https://github.com/ubuntu/microk8s/files/3048374/inspection-report-20190405_143043.tar.gz)

Hi, as the issue subject said, i failed to access microk8s installed on my VM, the host OS is Mac.

I tried to install it by getting the config using `microk8s.config` and copy it to the host machine.

and i got this
```
ibk:~
$ KUBECONFIG=~/microk8s_kubeconfig kubectl get pods
Unable to connect to the server: x509: certificate is valid for 127.0.0.1, 10.152.183.1, 10.0.2.15, not 192.168.33.10
```

I'm aware about #110, but it is my local network, so it should be safe.

The solution seems to be to regenerate the certificate, but i'm not sure how to do it on microk8s.

ktsakalozos:
Hi @iwanbk, could you try the `microk8s.config -l` (note the `-l` flag) to produce the config. If this is no good for you, could you try the `--edge` channel (`sudo snap install microk8s --classic --edge`)? The issue you are reporting is already addressed and is released to edge, we will soon release it to the rest of the channels.

iwanbk:
> microk8s.config -l 

didn't help

> sudo snap install microk8s --classic --edge

didn't work as well.

FYI, i created my ubuntu 18.04 with vagrant, and it has two eth* interfaces.
I disabled the local one, so the only one left is the "public" interfaces.

And then i installed the `--edge` channel and it worked!

iwanbk:
>  and it worked!

note that i only did small test:
- create pod
- kubectl get pods

ktsakalozos:
@iwanbk the code that figures out the ip is here: https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/common/utils.sh#L98

Could you share the output of `ip -o -4 addr list` and `netstat -rn`? I wonder if we are figuring out the wrong IP.

karlmutch:
I am seeing something very similar when using VirtualBox on an OSX machine to deploy the Linux host and then install microk8s from there.  When using AWS EC2 instances it seems fine.  In my case using client-go calls to send commands to a running pod causes the issue but it comes and goes at random seemingly without reinstalling.

```
[...]

PR #422, opened 2019-04-09
Include all host IPs in server cert

niklasholm:
Suggested fix for issue #415.

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -104,6 +104,16 @@ get_default_ip() {
     fi
 }
 
+get_ips() {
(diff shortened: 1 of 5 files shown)


microk8s-resources/actions/common/utils.sh
```diff
@@ -104,6 +104,16 @@ get_default_ip() {
     fi
 }
 
+get_ips() {
+    local IP_ADDR="$($SNAP/bin/hostname -I)"
```
ktsakalozos: I think `hostname` is not under  `$SNAP/bin/hostname`. Please add it under https://github.com/ubuntu/microk8s/blob/master/snapcraft.yaml#L132

PR #433, opened 2019-04-17
Take over the containerd management

ktsakalozos:
Makes MicroK8s play well next to containerd. Issue reported in https://github.com/ubuntu/microk8s/issues/430

PR #438, opened 2019-04-19
 Take over containerd and handle in pod execution on 1.11

ktsakalozos:


ktsakalozos:
Merging this fix. It is mainly a back port of #433

PR #439, opened 2019-04-19
Take over containerd and handle in pod execution on 1.10

ktsakalozos:


ktsakalozos:
Merging this since 1.10 seems to be broken causing the upgrade tests to fail.

Issue #421, opened 2019-04-09
Cannot connect to microk8s from another machine


soapergem:
I just ran a fresh install of microk8s with this command:

```
sudo snap install microk8s --classic --channel=1.14/stable
```

I can access everything while I'm SSH'ed into the box this is running on, with commands much like this:

```
microk8s.kubectl get pods
```

However when I set my `~/.kube/config` file - on another machine on the same network - to this:

```
apiVersion: v1
clusters:
- cluster:
    server: http://192.168.1.123:8080
  name: microk8s-cluster
contexts:
- context:
    cluster: microk8s-cluster
    user: admin
  name: microk8s
current-context: microk8s
kind: Config
preferences: {}
users:
- name: admin
  user:
    username: admin
```

And then try to run this command:

```
kubectl get pods
```

I get this error message:

```
The connection to the server 192.168.1.123:8080 was refused - did you specify the right host or port?
```

In fact, when I run `curl localhost:8080` while on the box, I see valid K8S output, but if I try running `curl 192.168.1.123:8080` from outside the box, even that returns:

```
curl: (7) Failed to connect to 192.168.1.123 port 8080: Connection refused
```

I've tried a couple of things to get this working, first of which was allowing access to 8080 on the firewall with this command:

```
sudo ufw allow 8080
sudo ufw enable
```

Unfortunately that doesn't solve the problem. When I run `netstat -an | grep "LISTEN "` I can see an entry for port 8080 as follows:

```
tcp        0      0 127.0.0.1:8080          0.0.0.0:*               LISTEN
```

But because it says `127.0.0.1` instead of `0.0.0.0`, this would explain why it's not working off the box. Is there some `microk8s.*` command I don't know about, to open up access?

ktsakalozos:
In v1.14 we restricted the insecure access to MicroK8s from port 8080 to local users only.

You have two options:
- Open insecure port 8080 to everyone. To do this you need to edit `/var/snap/microk8s/current/args/kube-apiserver` and set the `--insecure-bind-address=127.0.0.1` to `0.0.0.0`. Then restart MicroK8s with `microk8s.stop` and `microk8s.start`.

- Use the secure port 16443. Due to a recent bug fix you need to get MicroK8s from edge `sudo snap install microk8s --channel=1.14/edge --classic`. This bug fix will soon reach stable. Then use the config you get with `microk8s.config` to reach the cluster.

soapergem:
[...]

PR #441, opened 2019-04-20
Feature/modifiable csr

balchua:
In relation to the issue https://github.com/ubuntu/microk8s/issues/421 this PR makes a copy of csr.conf.template to `$SNAP_DATA/certs`, so users can add additional IPs.


ktsakalozos:
Thank you for the PR @balchua +1

pixelshaded:
Using 1.14/edge snap, I cannot get my edits to the template to work. Once you edit the template, how do you get microk8s to generate a new certificate from the template? I've stopped, started, and even reset the cluster with no luck. I always get the same list of hosts from the certificate error.

balchua:
@pixelshaded you're right, i couldn't find anywhere that regenerates the csr other than when the apiserver-kicker determines there is a change in the ip.  
Is the ip you are adding to the template not showing up when you do a `hostname -I`?

balchua:
@pixelshaded i may have found something, do you see a file `external_ip.txt` in `/var/snap/microk8s/current`?  If yes can you also add your ip address into that? And do a `microk8s.stop` then start? I think this may trigger the apiserver-kicker to recreate the server certs based on the `csr.conf.template`.

ktsakalozos:
@pixelshaded @balchua indeed this fix/feature needs a bit more thinking. The user experience is not good. 

@pixelshaded here is another way to update the csr file.
 - update `/var/snap/microk8s/current/certs/csr.conf.template` based on your needs
 - `rm  /var/snap/microk8s/current/certs/csr.conf`
 - `sudo snap set microk8s csr-refresh=true`

pixelshaded:
@balchua I'm not sure the external_ip.txt will work for my use case because I want to add a domain and not an IP. @ktsakalozos your suggestion worked for updating the file.

balchua:
@pixelshaded @ksakalozos thanks! Yeah we need to think about this a little bit more than just copying the template to the SNAP_DATA.

ktsakalozos:
@balchua what if we kept a copy of the template currently used under `$SNAP_COMMON/var/run`. When MicroK8s starts we detect if there are any changes on the template since last boot.

balchua:
@ktsakalozos the template in `$SNAP_COMMON/var/run` is the last used template and using this file and compare it with `$SNAP_DATA/certs/csr.conf.template`?

That can be a way, but why not render the `csr.conf.template` and compare that with `csr.conf`.  Unless there is something else im missing.  

Either way ,about the comparison, do we compare the data or a simple checksum / diff comparison will suffice?

ktsakalozos:
I believe you are right we could split the fuction https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/common/utils.sh#L145 in a `render_server_csr` and `produce_server_cert` so we could compare the current `csr.conf` to a rendered candidate.

Now that I am looking at the code it makes sense to have the apiservice kicker (https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/apiservice-kicker)  detect changes in the same way it already does for the IPs. So the if statement in https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/apiservice-kicker#L29 may look like "if (rendered.csr.conf differs from ${SNAP_DATA}/certs/csr.conf) then re-create certs". This should actually detect both IP and user changes, so we should be fine.

A checksum or diff should be enough, we will need to add the utility in the `snapcraft.yaml` to use it.

balchua:
Let me create an issue to keep track of this.

Issue #382, opened 2019-03-26
docker disappeared


g00nix:
Docker disappeared from microk8s:

```
root@kube:~# microk8s.docker
microk8s.docker: command not found
root@kube:~# ls /snap/bin/  
microk8s.config  microk8s.ctr  microk8s.disable  microk8s.enable  microk8s.inspect  microk8s.istioctl  microk8s.kubectl  microk8s.reset  microk8s.start  microk8s.status  microk8s.stop
root@kube:~# cat /etc/issue
Ubuntu 18.04.2 LTS \n \l
```

It it still going to be used in the project? Is there an alternative to inspect what kube is doing in the background?

If this was a planned change, is there documentation/release notes?

mattiasarro:
They have [replaced](https://github.com/ubuntu/microk8s/commit/7c3311a6d24a0eae482dc09e86471ead41aaa19e) docker daemon with containerd.

For me, this is a problem, since I was relying on exposing the Docker daemon on a TCP port. I did this by adding `-H tcp://0.0.0.0:$PORT` to `/var/snap/microk8s/current/args/dockerd`, which meant I could execute docker commands from outside the machine (with `docker -H tcp://$HOST:$PORT`).

Now I don't really know if I can connect to the containerd daemon the same way, and if so, how.

gun1x:
> For me, this is a problem

I can't say it's a "problem" for me. It was just a surprise. We were using the that docker to also upload images into the registry, but we just went for `apt install docker` for that scope.

Well, if that is the case, I guess it's up to the devs if they close the ticket or not.

ktsakalozos:
Hi,

Indeed in the 1.14 release contanerd replaced dockerd. We gave the heads up on this change in the on this topic https://github.com/ubuntu/microk8s/issues/382#issuecomment-476713382 and on the #microk8s channel at https://k8s.slack.com/ some time ago. For those who cannot do the transition dockerd is available from the 1.13 channel:
```
snap install microk8s --classic --channel=1.13/stable
```

One of the reasons why we moved ahead with containerd is that a few users wanted to deploy MicroK8s next to a local dockerd and that was causing unexpected issues. As @gun1x mentions this should not be a problem anymore. 

@mattiasarro, I am not aware of how you are using dockerd. Could you provide more info so we can offer some suggestions?

Apologies for any inconvenience we may have cause.

termie:
[...]

PR #453, opened 2019-05-04
Include Reg into microk8s

etheleon:
As suggested in this [thread](https://github.com/ubuntu/microk8s/issues/382#issuecomment-488608144), [reg](https://github.com/genuinetools/reg) will be a good addition to enable users to easily interact with local registry


etheleon:
@ktsakalozos , the tests are failing any idea why?

microk8s-resources/actions/enable.reg.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -e
+
+source $SNAP/actions/common/utils.sh
+
+read -ra ARGUMENTS <<< "$1"
+argz=("${ARGUMENTS[@]/#/--}")
```
ktsakalozos: I do not think you are doing anything with `argz` so lets simplify the code by removing these two lines.

microk8s-resources/actions/disable.reg.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e
+
+source $SNAP/actions/common/utils.sh
+
+echo "Disabling Reg"
```
ktsakalozos: I think a single message should be enough. :)

microk8s-resources/actions/enable.reg.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# check if reg cli is already in the system.  Download if it doesn't exist.
+if [ ! -f "${SNAP_DATA}/bin/reg" ]; then
+  REG_VERSION="${REG_VERSION:-v0.16.0}"
+  echo "Fetching reg version $REG_VERSION."
+  sudo mkdir -p "$SNAP_DATA/bin"
+  sudo "${SNAP}/usr/bin/curl" -L https://github.com/genuinetools/reg/releases/download/${REG_VERSION}/reg-linux-amd64 -o "$SNAP_DATA/bin/reg"
```
ktsakalozos: MicroK8s ships for amd64 adn arm64, so we should use the  $SNAP_ARCH to get the right binary from the releases. Something like this should work: `https://github.com/genuinetools/reg/releases/download/${REG_VERSION}/reg-linux-${SNAP_ARCH}`.

ktsakalozos:
@etheleon thank you for this PR. I left a few comments. It would be great if you could somehow add a test for this addon. Here is how to do that: 
- In https://github.com/ubuntu/microk8s/blob/master/tests/test-addons.py#L65 and https://github.com/ubuntu/microk8s/blob/master/tests/test-upgrade.py#L92 enable `reg` along with the registry.
- In https://github.com/ubuntu/microk8s/blob/master/tests/validators.py#L163 use the `reg` binary to do any operation on the private registry. An ls should return the my-busybox image, right?

If you follow the instruction and build MicroK8s from source https://github.com/ubuntu/microk8s#building-from-source you should see that you do not have a `microk8s.reg` command in your path. The reason is you are missing an entry into `snapcraft.yaml` like this one: https://github.com/ubuntu/microk8s/blob/master/snapcraft.yaml#L66 and the corresponding wrapper script like this one: https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/microk8s-linkerd.wrapper

It would have been better if the fist time you try the `microk8s.req` command the wrapper would download the binary then. This way you would not need the `microk8s.enable reg` step. What do you think?

microk8s-resources/actions/enable.reg.sh
etheleon: Alright!

microk8s-resources/actions/disable.reg.sh
etheleon: Alright!

microk8s-resources/actions/enable.reg.sh
etheleon: Alright!

etheleon:
@ktsakalozos Thanks for the detailed write up on writing the test! Let me get started on it! :)

etheleon:
- snapcraft: add reg as app
- validator: include validator for `reg`
  * tags busybox
  * push to local registry
  * assets presence of image
  * deletes tag
- microk8s.reg wrapper: installs reg if binary not found
- microk8s.enable reg: installs reg (might be duplicate)

ktsakalozos:
@etheleon are you able to see this: https://travis-ci.org/ubuntu/microk8s/builds/534665856?utm_source=github_status&utm_medium=notification  ?

I believe you are missing an import. 
```
       validate_reg()
      NameError: global name 'validate_reg' is not defined
```

ktsakalozos:
@etheleon now we have:
```
>       output = (subprocess
                  .run(['reg', 'ls', '-f', 'localhost:32000'], stdout=subprocess.PIPE)
                  .stdout
                  .decode('utf-8')
                  .split("\n"))
E       AttributeError: 'module' object has no attribute 'run'
```

I should document how to run the tests locally, (they are run with pytest).

tests/test-upgrade.py
```diff
@@ -96,6 +97,15 @@ def test_upgrade(self):
         except:
             print('Will not test registry')
 
+
+        try:
+            enable = microk8s_enable("reg")
```
ktsakalozos: Can you please add a `timeout_insec=30` when enabling the add-on. If we do not set a timeout enabling tries for too long and tests on travis timeout.

ktsakalozos:
Closing this PR due to inactivity.

Issue #286, opened 2019-01-20
kubectl port forward missing libnsl.so


DanielHouston:
[inspection-report-20190120_223636.tar.gz](https://github.com/ubuntu/microk8s/files/2777003/inspection-report-20190120_223636.tar.gz)

Installed via the snap on Fedora 29, after running `microk8s.kubectl port-forward service/abc 8085:8080` and then trying to access the port via the web browser, I get the following:

`portforward.go:391] an error occurred forwarding 8085 -> 8080: error forwarding port 8080 to pod bb350c827195b04cac4520fbb81b9c0f7806f872d01bf97c0b78dd3b7665a56b, uid : exit status 127: /snap/microk8s/354/usr/bin/socat: error while loading shared libraries: libnsl.so.1: cannot open shared object file: No such file or directory`

JimPatterson:
Any progress on this?  I'm still seeing it on Fedora 29 with the latest microk8s (v1.14.1 rev 522).

I think it might be related to #371 and that issue mentions socat, but the fix committed in #372  did not mention socat.

For a simple reproducer, in one terminal run:

```
kubectl run --generator=run-pod/v1 forward-example --image=hjacobs/tiny-docker-http-test --port 8080
kubectl port-forward pods/forward-example 9000:8080
```

and then in another terminal run: `curl http://localhost:9000 ; echo`

In the first terminal you will see:

```
Forwarding from 127.0.0.1:9000 -> 8080
Forwarding from [::1]:9000 -> 8080
Handling connection for 9000
E0506 12:33:14.257000   17353 portforward.go:400] an error occurred forwarding 9000 -> 8080: error forwarding port 8080 to pod 833b6afb5307f47f1352c10f2a968ad54c8867481c3be0dbe1939ca5f775e724, uid : failed to execute portforward in network namespace "/var/run/netns/cni-4a1575b8-86b4-70cd-57dc-e4cd1f4210c7": socat command returns error: exit status 127, stderr: "/snap/microk8s/522/usr/bin/socat: error while loading shared libraries: libnsl.so.1: cannot open shared object file: No such file or directory\n"
```

In case this helps: [inspection-report-20190506_124554.tar.gz](https://github.com/ubuntu/microk8s/files/3148967/inspection-report-20190506_124554.tar.gz)

JimPatterson:
The above PR fixes this on Fedora 29, but I don't know enough about snaps to say that it does not cause some problem somewhere else.

PR #459, opened 2019-05-06
Fix socat needing patchelf (ubuntu/microk8s#286)

JimPatterson:
On Fedora 29, when using port forward, socat failed with:

    error while loading shared libraries: libnsl.so.1:
    cannot open shared object file: No such file or directory


ktsakalozos:
LGTM +1 Thank you @JimPatterson

PR #462, opened 2019-05-07
adds command to make iptables -P FORWARD ACCEPT save

TheNotary:
I just got bit by this the other day.  It looked like a DNS issue that other users had in the past, so I geared up to diagnose DNS issues (2hr because I was quite rusty with docker).  Then I realized that it was a complete network issue after pinging 8.8.8.8 and getting nothing which luckily reminded me of a blog post where a guy lost tons of time diagnosing the issue.  I hope there weren't technical reasons why this package wasn't added to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,10 @@ The [Kubenet](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-stor
 Make sure packets to/from the pod network interface can be forwarded
 to/from the default interface on the host:
 
-`sudo iptables -P FORWARD ACCEPT`
+```
+sudo iptables -P FORWARD ACCEPT
+sudo apt-get install iptables-persistent # this package allows you to persist this iptable rule across reboot
+```
 
 or, if using `ufw`:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -160,7 +160,10 @@ The [Kubenet](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-stor
 Make sure packets to/from the pod network interface can be forwarded
 to/from the default interface on the host:
 
-`sudo iptables -P FORWARD ACCEPT`
+```
+sudo iptables -P FORWARD ACCEPT
+sudo apt-get install iptables-persistent # this package allows you to persist this iptable rule across reboot
```
ktsakalozos: Could we have the "# this package..." line above the `sudo apt-get install iptables-persistent`?

ktsakalozos:
Thank you for this PR @TheNotary looks fine.

Can you also add the same suggestion in the output of `microk8s.inspect`. It will reach more users this way.
https://github.com/ubuntu/microk8s/blob/master/scripts/inspect.sh#L97

TheNotary:
Sure thing!  Do those updates look ok?  In the past I've had BSD users become confused with shell commands beginning with `#` symbols because to them it means "these commands were run as root."

ktsakalozos:
Thank you for the PR @TheNotary. LGTM +1

PR #467, opened 2019-05-10
Add link to Cilium MicroK8s guide

joestringer:
There's a short guide in the Cilium docs about how to bootstrap microk8s with Cilium providing the networking/security as a CNI. Add it to the community links.

### DIFF
--- a/docs/community.md
+++ b/docs/community.md
@@ -6,6 +6,7 @@ Feel free to update the table below by submitting a Pull Request.
 
 | Event/Mention                                          | Author(s)       | Source     | Date          |
 |---------------------------------------------------------------------|--------------|----------|---------------|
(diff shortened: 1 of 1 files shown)


ktsakalozos:
LGTM +1 Thank you @joestringer

ktsakalozos:
I wonder if it would be better to have Cilium as an addon. If you put the installation instructions in a script under https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/ and name it `enable.cilium.sh` you would just need to do a `microk8s.enable cilium`.

joestringer:
@ktsakalozos ah, that sounds like a potentially good approach, I'll look into it.

joestringer:
Here's the PR for the above functionality: https://github.com/ubuntu/microk8s/pull/594

Issue #468, opened 2019-05-13
Cluster gets stuck in SchedulingDisabled state until reboot


tdickman:
My local node sometimes enters the `SchedulingDisabled` state - restarting my node usually fixes the issue, but it's unclear why this is occurring.

```
$ kubectl get node
NAME          STATUS                     ROLES    AGE     VERSION
tom-x1-work   Ready,SchedulingDisabled   <none>   5d22h   v1.14.1
```

The event logs from `kubectl describe node` don't reveal anything interesting:

```
  Normal   Starting                 43m                kube-proxy, tom-x1-work  Starting kube-proxy.
  Normal   Starting                 43m                kubelet, tom-x1-work     Starting kubelet.
  Warning  InvalidDiskCapacity      43m                kubelet, tom-x1-work     invalid capacity 0 on image filesystem
  Normal   NodeAllocatableEnforced  43m                kubelet, tom-x1-work     Updated Node Allocatable limit across pods
  Normal   NodeHasNoDiskPressure    43m                kubelet, tom-x1-work     Node tom-x1-work status is now: NodeHasNoDiskPressure
  Normal   NodeHasSufficientPID     43m                kubelet, tom-x1-work     Node tom-x1-work status is now: NodeHasSufficientPID
  Normal   NodeNotReady             43m                kubelet, tom-x1-work     Node tom-x1-work status is now: NodeNotReady
  Normal   NodeNotSchedulable       43m                kubelet, tom-x1-work     Node tom-x1-work status is now: NodeNotSchedulable
  Normal   NodeHasSufficientMemory  43m                kubelet, tom-x1-work     Node tom-x1-work status is now: NodeHasSufficientMemory
  Normal   NodeReady                43m                kubelet, tom-x1-work     Node tom-x1-work status is now: NodeReady
```

[inspection-report-20190513_094707.tar.gz](https://github.com/ubuntu/microk8s/files/3173179/inspection-report-20190513_094707.tar.gz)

tdickman:
This same issue occurred today - this time I had stopped the cluster the night before, suspended my computer, and then started the cluster again this morning. I went to start an application on my computer 30 minutes later, and saw that the cluster was in `SchedulingDisabled` state.

I restarted my cluster, and it is still in the `SchedulingDisabled` state.

ktsakalozos:
@tdickman with the merged PR the stop command will not drain the node. This means the node will never go into SchedulingDisabled and we will not see this race condition again (I hope).

[...]

PR #469, opened 2019-05-15
Start/stop will not drain/uncordon the node

ktsakalozos:
Fixes: https://github.com/ubuntu/microk8s/issues/468

The user will have to drain the node himself if he wished to right before stopping MicroK8s.


ktsakalozos:
Merging this to get some feedback.

PR #471, opened 2019-05-16
Remove user guide from docs

ktsakalozos:
Keeping only the dev docs and pointing to the official docs for the user guide.


ktsakalozos:
@evilnick @tvansteenburgh here is how the new README would look like: 
https://github.com/ubuntu/microk8s/blob/feature/only-dev-guide/README.md

evilnick:
It would be nice to work the community stuff into the docs too at some point.

ktsakalozos:
Thank you

tdensmore:
I think it was a terrible idea to move documentation OUT of this repo. Most developers refer to the codebase as **the source of truth**.

ktsakalozos:
@tdensmore the user guide docs where moved to https://github.com/canonical-web-and-design/microk8s.io/ as they are being hosted on the microk8s.io site. It seems to be common practice to have the docs in their own project.

For some time we tried to keeps the docs in two places and sync them periodically. This did not work very well. We were often running into issues like styles that did not match, fixes that were accidentally skipped etc. I am sure you understand the benefit of not having duplicate code; it is almost the same thing.

We are open to any suggestions on how to handle this.

Issue #470, opened 2019-05-15
Auto detect changes in csr.conf.template and recreate certs.


balchua:
This is a continuation of the modifiable csr PR. https://github.com/ubuntu/microk8s/pull/441
Related to issue https://github.com/ubuntu/microk8s/issues/421

ktsakalozos:
Thank you @balchua

ikari-pl:
In layman's terms, what do I do to make use of this feature? In one of the previous PRs I found manual steps in teh comments, that included the following:
```
    mcedit /var/snap/microk8s/671/certs/csr.conf.template
    rm /var/snap/microk8s/current/certs/csr.conf
    sudo snap set microk8s csr-refresh=true```
but I noticed that while the extra domain has been added to my cert, the extra IP has not. People are also mentioning a `external_ip.txt` file, that I do not have.

balchua:
@ikari-pl all the user need to do is update the file `/var/snap/microk8s/current/certs/csr.conf.template` with your changes.  Once saved microk8s will automatically update the csr.conf and generate the certs accordingly.  After that it will restart the apiserver.

ktsakalozos:
Hi @ikari-pl updating the template file should be enough to trigger the rendering of the csr.  As you can see in https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/common/utils.sh#L170 IPs are extracted from the system's and places in the csr template replaceing the #MOREIPS tag. If you have your own set of IPs you can remove that tag from the csr template.

ikari-pl:
Thanks for the super fast response. @ktsakalozos @balchua i guess it might be caused by the fact that, if I read the bash correctly, it overwrites IPs 3 and up, generating its own list starting with `IP.3` which is the number I have used :) I was able to work around this now.
Would it also work if I put `IP.anythingelse`?

balchua:
@ikari-pl i think so, its just a way to make sure user dont override the previous key (IP or DNS) value. These DNS  and IP will be added to the certs.

ikari-pl:
Yes, I just wanted to raise the point that just by editing the file I didn't know only numbers lower than 3 will remain, and 3 and up will get overwritten by the autogenerated sections. Maybe the auto ones could start with a high number, like 990? :)

balchua:
Got what you meant.  From the `csr.conf.template` anything below the `#MOREIPS` are user supplied.  From there you can specify and key.

PR #476, opened 2019-05-19
update the csr.conf when csr.conf.template changed.

balchua:
creates a csr.conf.rendered and compare it with csr.conf.
Forces the apiserver-kicker to recreate and reload the certificates when changes are detected.

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -156,7 +156,7 @@ produce_server_cert() {
         echo "changeme" >  "${SNAP_DATA}/certs/csr.conf" 
     fi
 
-    if ! cmp -s "${SNAP_DATA}/certs/csr.conf.rendered" "${SNAP_DATA}/certs/csr.conf"; then
+    if ! "${SNAP}/usr/bin/cmp" -s "${SNAP_DATA}/certs/csr.conf.rendered" "${SNAP_DATA}/certs/csr.conf"; then
       cp ${SNAP_DATA}/certs/csr.conf.rendered ${SNAP_DATA}/certs/csr.conf
       openssl req -new -key ${SNAP_DATA}/certs/server.key -out ${SNAP_DATA}/certs/server.csr -config ${SNAP_DATA}/certs/csr.conf
       openssl x509 -req -in ${SNAP_DATA}/certs/server.csr -CA ${SNAP_DATA}/certs/ca.crt -CAkey ${SNAP_DATA}/certs/ca.key -CAcreateserial -out ${SNAP_DATA}/certs/server.crt -days 100000 -extensions v3_ext -extfile ${SNAP_DATA}/certs/csr.conf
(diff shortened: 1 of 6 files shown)


balchua:
related to https://github.com/ubuntu/microk8s/issues/470 and previous PR https://github.com/ubuntu/microk8s/pull/441

snapcraft.yaml
```diff
@@ -159,6 +159,7 @@ parts:
     - util-linux
     - zfsutils-linux
     - iproute2
+    - diffutils
```
ktsakalozos: Can we move this package in the stage-packages of the containerd part so it is not affacted by the "no-patchelf"?

balchua:
Sure. Let me get to that.

ktsakalozos:
Thank you for the PR @balchua . Please allow for a small delay in reviewing this, we are caught up at kubecon.

microk8s-resources/actions/common/utils.sh
```diff
@@ -143,13 +143,36 @@ get_ips() {
...
+    #  0 if no change
+    #  1 otherwise. 
+
+    render_csr_conf
+    if ! [ -f "${SNAP_DATA}/certs/csr.conf" ];
+    then
+        echo "changeme" >  "${SNAP_DATA}/certs/csr.conf" 
+    fi
+
+    if ! cmp -s "${SNAP_DATA}/certs/csr.conf.rendered" "${SNAP_DATA}/certs/csr.conf"; then
```
ktsakalozos: We would better use cmp with the full path `${SNAP}/usr/bin/cmp`  so we know we always use the one `cmp` we ship.

ktsakalozos: We do not do this for openssl because openssl is a dependency to snapd and it is on the host for sure.

microk8s-resources/actions/common/utils.sh
```diff
@@ -143,13 +143,36 @@ get_ips() {
...
+      echo "0"
+    fi
+
+}
+
+render_csr_conf() {
+    # Render csr.conf.template to csr.conf.rendered
 
-    local IP_ADDR="$1"
+    local IP_ADDR="$(get_ips)"
```
ktsakalozos: Since we do not get the IPs as arguments the for loop below should be updated:
```
        for ADDR in $(echo "$IP_ADDR"); do
            ips+="${sep}IP.$((i++)) = ${ADDR}"
            sep='\n'
        done
```

microk8s-resources/actions/common/utils.sh
balchua: Oh i see. I'll get to that.

microk8s-resources/actions/common/utils.sh
balchua: Aaa i missed this one.  And my workstation only comes with one network so the issue didn't manifest.  I'll work on this one. Thanks!

balchua:
I have updated util.sh. thanks

ktsakalozos:
LGTM +1 thank you @balchua

Issue #457, opened 2019-05-05
Run ingress without default backend


JorritSalverda:
The _ingress_ addon is installed with a default backend by creating a deployment for `default-http-backend` and passing argument `--default-backend-service=$(POD_NAMESPACE)/default-http-backend` to the `nginx-ingress-microk8s-controller` deployment. Nginx can return the 404 itself without the default backend though, so it's a waste of resources.

Best to drop the argument and remove the `default-http-backend` deployment.

ktsakalozos:
Good point @JorritSalverda. Thank you for pointing this out.

It should be as easy as updating the ingress yaml https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/ingress.yaml

Marking this issue as "help needed" in hope that someone will pick this up before us. Thanks.

balchua:
i was looking into this, and just wondering why is the arm64 version is limited to 0.11.
https://github.com/ubuntu/microk8s/blob/8cbb4d30f5850fc29e25d09ee207f2f85f3fb233/microk8s-resources/actions/enable.ingress.sh#L14
```
if [ "${ARCH}" = arm64 ]
then
  TAG="0.11.0"
  EXTRA_ARGS=""
fi
```

But quay.io repo have tags all the way to `0.24.1`.

https://quay.io/repository/kubernetes-ingress-controller/nginx-ingress-controller-arm64?tab=tags

And how does one test an arm64 architecture? with cloud provider?

thanks.

ktsakalozos:
@balchua at the time we introduced the ingress add-on 0.15 was available for amd64 but the arm64 image was failing. I have no objections in testing and upgrading to the latest stable ingress.

To test on arm64 we just grab an Ubuntu machine from AWS (amazon offers arm64). In our CI we use juju to provision one such machine https://github.com/charmed-kubernetes/jenkins/blob/master/jobs/release-microk8s-beta/Jenkinsfile#L30 . Testing is exactly the same as in the case of amd64, try running test-distro https://github.com/ubuntu/microk8s/blob/master/tests/test-distro.sh to test a few distros in lxc containers.

balchua:
@ktsakalozos let me try to get an arm64 machine first. 😁. Have not tried any arm based server before.

balchua:
Closing as it is already fixed by this PR https://github.com/ubuntu/microk8s/pull/662

PR #484, opened 2019-05-27
Remove ingress default backend deployment, service and argument

JorritSalverda:
Fixes https://github.com/ubuntu/microk8s/issues/457

### DIFF
--- a/microk8s-resources/actions/ingress.yaml
+++ b/microk8s-resources/actions/ingress.yaml
@@ -1,58 +1,3 @@
-apiVersion: extensions/v1beta1
-kind: Deployment
-metadata:
-  name: default-http-backend
-  labels:
-    app: default-http-backend
-spec:
-  replicas: 1
-  selector:
-    matchLabels:
-      app: default-http-backend
-  template:
-    metadata:
-      labels:
-        app: default-http-backend
-    spec:
-      terminationGracePeriodSeconds: 60
-      containers:
-      - name: default-http-backend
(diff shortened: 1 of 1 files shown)


JorritSalverda:
Would you like me to add instructions somewhere on how to clean up the old deployment and service? And if so what would be a good place?

balchua:
@JorritSalverda you probably need to also update the arm64 version too.

balchua:
@JorritSalverda there's also a bunch of tests related to ingress in https://github.com/ubuntu/microk8s/blob/8cbb4d30f5850fc29e25d09ee207f2f85f3fb233/tests/validators.py#L79.  currently it checks the default-http-backend pod if its running or not.

ktsakalozos:
> Would you like me to add instructions somewhere on how to clean up the old deployment and service? And if so what would be a good place?

To do a proper cleanup we should keep the backend related parts of the old `ingress.yaml` in an `ingress-backend.yaml` and when we do a `microk8s.disable ingress` delete that just in case we are transitioning from an "old" ingress to a "new" one. The file you will need to patch is https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/disable.ingress.sh#L21

balchua:
If no one mind, I am continuing this PR. https://github.com/ubuntu/microk8s/pull/619

balchua:
@ktsakalozos @JorritSalverda i think we can close this PR, since #619 is merged.

PR #495, opened 2019-06-05
Fix dashboard when rbac is enabled

balchua:
Kubernetes dashboard is not working when RBAC is enabled. 
Upgraded dashboard to v1.10.1 with minimum required `Roles` for the dashboard.

Dashboard v1.10.1 removed the "Skip" button forces users to use the access token or kubeconfig, thus improving security.


ktsakalozos:
@balchua overall I am +1 to this PR. I am a bit skeptical about removing the skip button, but we need to do it.

What would you say if we added a message at the end of `microk8s.enable dashboard` saying:
"
To access the dashboard use the default token retrieved with:

token=$(microk8s.kubectl -n kube-system get secret | grep default-token | cut -d " " -f1)
microk8s.kubectl -n kube-system describe secret $token
"

balchua:
@ktsakalozos.  Thanks for the review.  im good with adding extra messages at the end of the command.

In the new dashboard, the administrator can create different service accounts to give limited accesses to kubernetes objects.

Maybe a message after `microk8s.enable dashboard` with a link to FAQ where we explain how to get access to the new dashboard.? Wdyt?

ktsakalozos:
Merging this PR. Thank you @balchua. Lets continue the discussion over at: https://github.com/ubuntu/microk8s/pull/504

Issue #320, opened 2019-02-13
Make Knative work on MicroK8S


buggtb:
There's something weird going on with MicroK8S DNS stuff that prevents it functioning with Knative.

This is more a note to self than bug report but:

```
root@juju-f46d51-73:~# kubectl get ksvc helloworld-go  --output=custom-columns=NAME:.metadata.name,DOMAIN:.status.domain
NAME            DOMAIN
helloworld-go   <none>

```

Should return a domain and doesn't.

Tried deployments with
https://github.com/knative/docs/blob/master/install/Knative-with-any-k8s.md
and
https://github.com/knative/docs/blob/master/install/Knative-with-Minikube.md

ktsakalozos:
Hi @buggtb ,

I do not get the same result on an aws machine:
```
ubuntu@ip-172-31-12-115:~$ microk8s.kubectl get ksvc helloworld-go  --output=custom-columns=NAME:.metadata.name,DOMAIN:.status.domain
NAME            DOMAIN
helloworld-go   helloworld-go.default.example.com
```

Here is the full set of what I did:
```
sudo snap install microk8s --classic
microk8s.enable istio
microk8s.kubectl label namespace default istio-injection=enabled
curl -L https://github.com/knative/serving/releases/download/v0.3.0/serving.yaml   | sed 's/LoadBalancer/NodePort/'   | microk8s.kubectl apply --filename -
sudo iptables -P FORWARD ACCEPT
microk8s.kubectl apply -f ./service.yaml 
microk8s.kubectl get ksvc helloworld-go  --output=custom-columns=NAME:.metadata.name,DOMAIN:.status.domain
```

The `service.yaml` looks like this:
```
apiVersion: serving.knative.dev/v1alpha1 # Current version of Knative
kind: Service
metadata:
  name: helloworld-go # The name of the app
  namespace: default # The namespace the app will use
spec:
  runLatest:
    configuration:
      revisionTemplate:
        spec:
          container:
            image: gcr.io/knative-samples/helloworld-go # The URL to the image of the app
            env:
              - name: TARGET # The environment variable printed out by the sample app
                value: "Go Sample v1"
```

Here is the final view of my cluster:
```
NAMESPACE         NAME                                           READY   STATUS      RESTARTS   AGE
istio-system      pod/grafana-6b99c54645-229wc                   1/1     Running     0          18m
istio-system      pod/istio-citadel-68c48d4fc8-m86nd             1/1     Running     0          18m
[...]

PR #501, opened 2019-06-10
Added support for Knative (#320)

olatheander:
Added support for enabling/disabling Knative and implicitly also enabling Istio since Knative depends on it.


balchua:
@olatheander i think your test is failing because it is looking for `activator-service`, when it should only be `activator`.

olatheander:
@balchua Thanks for the feedback and you're correct. I have renamed to `activator`so let's see how Travis likes it now.

ktsakalozos:
Thank you for this PR @olatheander . There are a few adjustment I will do in a followup PR.

There is a race condition in https://github.com/ubuntu/microk8s/pull/501/files#diff-7ea263c42dd785375276d6c28ab62355R32 where the second apply may fail because the CRDs are not applied yet. The result is that the enable will fail with: 
```
unable to recognize "/snap/microk8s/x1/actions/knative/release.yaml": no matches for kind "ClusterChannelProvisioner" in version "eventing.knative.dev/v1alpha1"
unable to recognize "/snap/microk8s/x1/actions/knative/release.yaml": no matches for kind "ClusterChannelProvisioner" in version "eventing.knative.dev/v1alpha1"
```

The travis tests are timing out so I will merge the knative and istio tests and have them running in our CI only.

olatheander:
@ktsakalozos Thank you for the merge and the feedback regarding the race-condition. I try and look into it too.

Regarding the time-out of the tests that is something I'm looking into but have not yet a fix for but I have an idea for a solution. There is no particular reason why it should time out, i.e. the time it takes to wire up Istio + Knative is not in that magnitude. If it works out there will be another PR coming up.

ktsakalozos:
@olatheander I pinged you over at the followup PR https://github.com/ubuntu/microk8s/pull/505 , let me know what you think.

PR #511, opened 2019-06-17
allow-privileged not available for 1.15 kubelet

ktsakalozos:


ktsakalozos:
Merging since there is no way around this.

PR #528, opened 2019-07-02
kubeflow addon

ktsakalozos:
This is a way to add kubeflow, Needs tests, reviews etc


balchua:
Im no juju expert here, but im curious to know why `destroy-all-models`?  I thought one can use `remove-application`

```
echo "Removing Kubeflow"
"$SNAP/microk8s-juju.wrapper" destroy-controller $CONTROLLER --destroy-all-models --destroy-storage 
```

ktsakalozos:
@balchua yes, we could remove all applications that make up kubeflow one by one. We could even remove the model. The reason we remove the controller as well is that since there is no kubeflow model/application there is nothing for the controller to tend, it just takes up resources. I would rather hide the interaction with juju completely.

Btw I am very excited about bringing in tools like juju on demand to help setting up infrastructures around MicroK8s. 

@knkski may have a better answer.

microk8s-resources/actions/disable.kubeflow.sh
```diff
@@ -0,0 +1,13 @@
...
+set -eu
+
+source $SNAP/actions/common/utils.sh
+
+
+CONTROLLER="uk8s"
+MODEL="kubeflow"
+
+echo "Removing Kubeflow"
+"$SNAP/microk8s-juju.wrapper" destroy-controller $CONTROLLER --destroy-all-models --destroy-storage || true
```
ktsakalozos: This could be "Destroy the kubeflow model. If there are no more models on this controller, remove the controller as well."

Also not sure about the `|| true` why would we want this?

microk8s-resources/actions/enable.kubeflow.sh
```diff
@@ -0,0 +1,68 @@
...
+
+Congratulations, Kubeflow is now available.
+Run \`microk8s.kubectl proxy\` to be able to access the dashboard at
+
+    http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/#!/overview?namespace=$MODEL
+
+The central dashboard is available at http://$1/
+To tear down Kubeflow and associated infrastructure, run this command:
+
+    microk8s.juju kill-controller $CONTROLLER --destroy-all-models --destroy-storage
```
ktsakalozos: This should be `microk8s.disable kubeflow`, no juju reference.

microk8s-resources/actions/enable.kubeflow.sh
```diff
@@ -0,0 +1,68 @@
...
+
+
+function bootstrap_and_deploy () {
+  "$SNAP/microk8s-juju.wrapper" bootstrap $CLOUD $CONTROLLER
+  "$SNAP/microk8s-juju.wrapper" add-model $MODEL $CLOUD
+  # Uncomment this line to present local disks into microk8s as Persistent Volumes
+  # microk8s.kubectl create -f storage/local-storage.yml || true
+  "$SNAP/microk8s-juju.wrapper" create-storage-pool operator-storage kubernetes storage-class=microk8s-hostpath
+  "$SNAP/microk8s-juju.wrapper" deploy kubeflow
+  # juju-wait is needed?
```
ktsakalozos: @knkski have you seen cases where `juju-wait` is needed? If yes, we may want to re-implement the waiting the juju-wait is doing.

microk8s-resources/actions/enable.kubeflow.sh
```diff
@@ -0,0 +1,68 @@
...
+  "$SNAP/microk8s-enable.wrapper" dns storage dashboard juju
+  for i in {1..5}
+  do
+    "$SNAP/microk8s-status.wrapper" --wait-ready --timeout 60 && break || sleep 10
+  done
+}
+
+
+function bootstrap_and_deploy () {
+  "$SNAP/microk8s-juju.wrapper" bootstrap $CLOUD $CONTROLLER
```
ktsakalozos: Test the existence of the controller and reuse it or fail gracefully. The same for adding the model.

microk8s-resources/actions/enable.kubeflow.sh
```diff
@@ -0,0 +1,68 @@
...
+  "$SNAP/microk8s-juju.wrapper" add-model $MODEL $CLOUD
+  # Uncomment this line to present local disks into microk8s as Persistent Volumes
```
ktsakalozos: We should make this a configuration parameter eg `microk8s.enable kubeflow:local-storage=true`

microk8s-resources/actions/kubeflow/sa.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
```
ktsakalozos: We need instructions (at least a README) describing how these manifests where created.

microk8s-resources/actions/kubeflow/secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mlpipeline-minio-artifact
+type: Opaque
+data:
+  accesskey: bWluaW8=
```
ktsakalozos: Should these be auto-generated during `microk8s.enable`

microk8s-resources/wrappers/microk8s-juju.wrapper
```diff
@@ -0,0 +1,25 @@
...
+set -eu
+
+export PATH="$SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH"
+export LD_LIBRARY_PATH="$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/x86_64-linux-gnu:$SNAP/usr/lib/x86_64-linux-gnu"
+export GOCOOKIES=/var/lib/landscape/juju-homes/1/.go-cookies
+export JUJU_DATA="$SNAP_DATA/juju/share/juju"
+export JUJU_HOME="$SNAP_DATA/juju-home"
+
+if [ ! -f "${SNAP_DATA}/bin/juju" ]; then
+  echo "Juju not available, try enabling kubeflow with 'microk8s.enable kubeflow'"
```
ktsakalozos: That should be `microk8s.enable juju`

microk8s-resources/actions/enable.juju.sh
```diff
@@ -0,0 +1,25 @@
...
+
+source $SNAP/actions/common/utils.sh
+
+
+function get_juju_client () {
+  # check if juju cli is already in the system.  Download if it doesn't exist.
+  if [ ! -f "${SNAP_DATA}/bin/juju" ]; then
+    sudo mkdir -p "$SNAP_DATA/bin"
+    sudo mkdir -p "$SNAP_DATA/tmp"
+    sudo "${SNAP}/usr/bin/curl" -L https://launchpad.net/juju/2.6/2.6.4/+download/juju-2.6.4-centos7.tar.gz -o "$SNAP_DATA/tmp/juju.tar.gz"
```
ktsakalozos: Hopefully the juju people will be able to point us to a smaller tarball. Thanks @wallyworld :)

balchua:
@ktsakalozos thanks for the response.  In my opinion having juju as an independent addon to microk8s is great.  It can allow a more seamless complex addons like you did for kubeflow easier to adapt inside microk8s.  But yeah juju binary is big.

microk8s-resources/actions/enable.kubeflow.sh
```diff
@@ -0,0 +1,68 @@
...
+For more information, see documentation at:
+https://github.com/juju-solutions/bundle-kubeflow/blob/master/README.md
+
+EOF
+
+}
+
+
+function enable_addons () {
+  "$SNAP/microk8s-enable.wrapper" dns storage dashboard juju
```
knkski: This could probably get changed to the bash equivalent of this:
https://github.com/juju-solutions/bundle-kubeflow/blob/master/scripts/deploy-microk8s#L62
(I've since updated the scripts to use Python instead of Bash, but the logic remains the same)

microk8s-resources/actions/disable.kubeflow.sh
knkski: The `|| true` was handy for local development for me, where I might be wanting to tear everything down and not care if one destruction step failed, but I don't think we need it here (same for the `|| true` below).

If we start getting errors from these lines due to say partial initialization that makes things weird, it may be worth adding a `--force` flag that ignores any errors.

microk8s-resources/actions/enable.kubeflow.sh
ktsakalozos: Using a python script to enable kubeflow is fine since we will end up including python in the snap for the clustering work. What is the best way to consume your scripts? I do not see versioned releases so should we copy them over to MicroK8s whenever we want to have an add-on release? Should we pull the scripts whenever the `microk8s.enable kubeflow` runs? Should we have a pip release? Something else?

I would rather not have and maintain a bash recreation of your work. Since you are making a kubeflow installer lets use that directly.

microk8s-resources/actions/disable.kubeflow.sh
wallyworld: I thought we need a -y here also?

microk8s-resources/actions/enable.juju.sh
wallyworld: Yes, we will look to do a k8s only CLI client, hopefully will be approx 12MB

microk8s-resources/actions/enable.kubeflow.sh
wallyworld: +1
juju controllers
will print the current controllers that are running, on which they'll be 1 or none.
juju show-controller
will provide some yaml that could be useful to see the controller state, or
juju status
may be preferred simply to make a read only call to the controller to see if it's running

microk8s-resources/actions/enable.kubeflow.sh
```diff
@@ -0,0 +1,68 @@
...
+}
+
+
+function bootstrap_and_deploy () {
+  "$SNAP/microk8s-juju.wrapper" bootstrap $CLOUD $CONTROLLER
+  "$SNAP/microk8s-juju.wrapper" add-model $MODEL $CLOUD
+  # Uncomment this line to present local disks into microk8s as Persistent Volumes
+  # microk8s.kubectl create -f storage/local-storage.yml || true
+  "$SNAP/microk8s-juju.wrapper" create-storage-pool operator-storage kubernetes storage-class=microk8s-hostpath
+  "$SNAP/microk8s-juju.wrapper" deploy kubeflow
```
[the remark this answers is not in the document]
wallyworld: explicit operator storage set up not needed since a while ago
juju works seamlessly with microk8s and uses hostpath storage automatically

microk8s-resources/actions/enable.kubeflow.sh
```diff
@@ -0,0 +1,68 @@
...
+  # juju-wait is needed?
+  #"$SNAP/microk8s-juju.wrapper" wait -vw
+
+  # General Kubernetes setup
+  "$SNAP/microk8s-kubectl.wrapper" create -n $MODEL -f "${SNAP}/actions/kubeflow" || true
+
+  "$SNAP/microk8s-juju.wrapper" config ambassador juju-external-hostname=localhost
+  "$SNAP/microk8s-juju.wrapper" expose ambassador
```
wallyworld: is expose needed on microk8s?
expose creates a k8s ingress resource, which requires an ingress controller etc
i've only seen this needed on firewalled public clouds with CDK
Can we check that it's needed here?

microk8s-resources/actions/kubeflow/sa.yaml
wallyworld: pod spec set v2 is coming which will add support for config maps, secrets, resources etc

microk8s-resources/actions/disable.kubeflow.sh
knkski: I left it off in the script that @ktsakalozos is adapting here, since the user will then get prompted about destroying things

microk8s-resources/actions/enable.kubeflow.sh
knkski: @ktsakalozos: Probably the easiest way at this point would be to just clone this repo and run `./scripts/deploy-microk8s {create|destroy}`. The charms won't work independently yet due to needing resources set up that Juju doesn't directly support yet.

ktsakalozos:
@knkski is this ready for a review?

knkski:
@ktsakalozos: Yep, this is ready for review

ktsakalozos:
@knkski this does not seem right:
```
Congratulations, Kubeflow is now available.
The dashboard is available at http://10.1.96.11/
To tear down Kubeflow and associated infrastructure, run:
```
http://10.1.96.11/ is a pod IP, there should be a service in front.

knkski:
@ktsakalozos: Are you concerned about it working at all, or it being the "proper" method? It is a PodIP, but that works fine for me locally.

ktsakalozos:
@knkski have you tested this in a cluster where the pod is not scheduled on the master node? Also what would happen if this pod gets rescheduled (eg after a restart/upgrade)?

ktsakalozos:
Thank you for this PR @knkski .

PR #567, opened 2019-07-30
allow hostpath provisioner to delete persistent volumes

bkc:
When RBAC has been enabled under v1.15.1, persistent volumes can be created using persistentvolumeclaims.

However after the persistentvolumeclaim is deleted, hostpath provisioner deletes the backing directory, but fails to delete the persistent volume due to a 'forbidden' error, e.g.

```
I0730 03:13:43.616042       1 controller.go:869] volume "pvc-0621280e-d5dc-4640-b232-38021643fb65" deleted
I0730 03:13:43.617107       1 controller.go:876] failed to delete volume "pvc-0621280e-d5dc-4640-b232-38021643fb65" from database: persistentvolumes "pvc-0621280e-d5dc-4640-b232-38021643fb65" is forbidden: User "system:serviceaccount:kube-system:microk8s-hostpath" cannot delete resource "persistentvolumes" in API group "" at the cluster scope
```
I was able to work around this by adding the `delete` verb to the ClusterRole for persistentvolumes.

### DIFF
--- a/microk8s-resources/actions/storage.yaml
+++ b/microk8s-resources/actions/storage.yaml
@@ -73,6 +73,7 @@ rules:
   - update
   - watch
   - create
+  - delete
 - apiGroups: [""]
   resources:
     - events


ktsakalozos:
Thank you for fixing this bug @bkc LGTM. Your fix will be available on the edge channel within the day and released to stable with the 1.15.2 release.

PR #576, opened 2019-08-03
Update Inspection Script

giorgos-apo:
Stores memory, disk, distribution, uptime, vm information. Checks for… SELinux and docker installation and prints warnings. Stores kubernetes info about pv and pvc.

### DIFF
--- a/scripts/inspect.sh
+++ b/scripts/inspect.sh
@@ -140,9 +140,9 @@ function suggest_fixes {
 
     if iptables -L | grep FORWARD | grep DROP &> /dev/null
     then
-        printf -- '\033[0;33m WARNING: \033[0m IPtables FORWARD policy is DROP. '
-        printf -- 'Consider enabling traffic forwarding with: sudo iptables -P FORWARD ACCEPT \n'
-        printf -- 'The change can be made persistent with: sudo apt-get install iptables-persistent\n'
+      printf -- '\033[0;33m WARNING: \033[0m IPtables FORWARD policy is DROP. '
+      printf -- 'Consider enabling traffic forwarding with: sudo iptables -P FORWARD ACCEPT \n'
+      printf -- 'The change can be made persistent with: sudo apt-get install iptables-persistent\n'
     fi
 
     ufw=$(ufw status)
--- a/scripts/inspect.sh
+++ b/scripts/inspect.sh
@@ -73,10 +73,10 @@ function store_kubernetes_info {
   # Collect some in-k8s details
   printf -- '  Inspect kubernetes cluster\n'
   mkdir -p $INSPECT_DUMP/k8s
-  /snap/bin/microk8s.kubectl version | sudo tee $INSPECT_DUMP/k8s/version > /dev/null
-  /snap/bin/microk8s.kubectl cluster-info | sudo tee $INSPECT_DUMP/k8s/cluster-info > /dev/null
-  /snap/bin/microk8s.kubectl cluster-info dump | sudo tee $INSPECT_DUMP/k8s/cluster-info-dump > /dev/null
-  /snap/bin/microk8s.kubectl get all --all-namespaces | sudo tee $INSPECT_DUMP/k8s/get-all > /dev/null
+  /snap/bin/microk8s.kubectl version 2>&1 | sudo tee $INSPECT_DUMP/k8s/version > /dev/null
+  /snap/bin/microk8s.kubectl cluster-info 2>&1 | sudo tee $INSPECT_DUMP/k8s/cluster-info > /dev/null
+  /snap/bin/microk8s.kubectl cluster-info dump 2>&1 | sudo tee $INSPECT_DUMP/k8s/cluster-info-dump > /dev/null
+  /snap/bin/microk8s.kubectl get all --all-namespaces 2>&1 | sudo tee $INSPECT_DUMP/k8s/get-all > /dev/null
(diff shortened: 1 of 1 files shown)


scripts/inspect.sh
```diff
@@ -88,22 +132,74 @@ function suggest_fixes {
...
+  fi
+
+  # check for selinux. if enabled, print warning.
+  if getenforce | grep -q 'Enabled'
+  then
+    printf -- '\033[0;33m WARNING: \033[0m SElinux is enabled. Consider disabling it.\n'
+  fi
+
+  # check for docker
+  if [ -d "/etc/docker/" ]; then
```
ktsakalozos: Could you test here that `localhost:32000` is not in `/etc/docker/`. You could do that with something like like  `if [ -d "/etc/docker/" ] && ! sudo grep localhost:32000 /etc/docker/daemon.json > /dev/null`

scripts/inspect.sh
```diff
@@ -88,22 +132,74 @@ function suggest_fixes {
...
   if echo $ufw | grep "Status: active" &> /dev/null && ! echo $ufw | grep cbr0 &> /dev/null
   then
-      printf -- '\033[0;33m WARNING: \033[0m Firewall is enabled. Consider allowing pod traffic '
-      printf -- 'with: sudo ufw allow in on cbr0 && sudo ufw allow out on cbr0\n'
+    printf -- '\033[0;33m WARNING: \033[0m Firewall is enabled. Consider allowing pod traffic '
+    printf -- 'with: sudo ufw allow in on cbr0 && sudo ufw allow out on cbr0\n'
+  fi
+
+  # check for selinux. if enabled, print warning.
+  if getenforce | grep -q 'Enabled'
```
ktsakalozos: Could we please do this: `if getenforce 2>&1 | grep  'Enabled' > /dev/null` so we do not print the stderr ?

scripts/inspect.sh
```diff
@@ -88,22 +132,74 @@ function suggest_fixes {
...
+
+    # check for selinux. if enabled, print warning.
+    if getenforce | grep -q 'Enabled'
+    then
+    	printf -- '\033[0;33m WARNING: \033[0m SElinux is enabled. Consider disabling it.\n'
+    fi
+
+
+    # check for docker
+    if [ -d "/etc/docker/" ]; then
```
ktsakalozos: This test seems to be a duplicate

scripts/inspect.sh
```diff
@@ -88,22 +132,74 @@ function suggest_fixes {
...
+
+    ufw=$(ufw status)
+    if echo $ufw | grep "Status: active" &> /dev/null && ! echo $ufw | grep cbr0 &> /dev/null
+    then
+      printf -- '\033[0;33m WARNING: \033[0m Firewall is enabled. Consider allowing pod traffic '
+      printf -- 'with: sudo ufw allow in on cbr0 && sudo ufw allow out on cbr0\n'
+    fi
+
+    # check for selinux. if enabled, print warning.
+    if getenforce | grep -q 'Enabled'
```
ktsakalozos: Duplicate test?

scripts/inspect.sh
```diff
@@ -88,22 +132,74 @@ function suggest_fixes {
     if lsof -Pi :8080 -sTCP:LISTEN -t &> /dev/null
     then
       printf -- '\033[0;33m WARNING: \033[0m Port 8080 seems to be in use by another application.\n'
+        if lsof -Pi :8080 -sTCP:LISTEN -t &> /dev/null
+        then
+            printf -- '\033[0;33m WARNING: \033[0m Port 8080 seems to be in use by another application.\n'
+        fi
+    fi
+
+    if iptables -L | grep FORWARD | grep DROP &> /dev/null
```
ktsakalozos: Probably something did not go well, this looks like a duplicate

giorgos-apo:
@ktsakalozos I have done what you 've asked. I 've also solved the duplicate issues. Please have a look.

scripts/inspect.sh
```diff
@@ -93,19 +137,34 @@ function suggest_fixes {
...
-}
 
+  # check for selinux. if enabled, print warning.
+  if getenforce 2>&1 | grep 'Enabled' > /dev/null
+  then
+    printf -- '\033[0;33m WARNING: \033[0m SElinux is enabled. Consider disabling it.\n'
+  fi
+
+  # check for docker
+  if [ -d "/etc/docker/" ] && ! sudo grep localhost:32000 /etc/docker/daemon.json > /dev/null; then
```
ktsakalozos: This gave me the following ugly output:
```
grep: /etc/docker/daemon.json: No such file or directory
 WARNING:  Docker is installed. You should mark the registry as insecure. Add the following to /etc/docker/deamon.json : 
 {
         "insecure-registries" : ["localhost:32000"] 
 }
```
Please test for the existence of the file with `[ -f "/etc/docker/daemon.json" ]`

ktsakalozos:
Instead of creating a separate directory for each info we gather could we put the  vm_name, disk_usage, uptime, lsb_release, openssl under sys. The reason for this is that these metrics have are system metrics with no "depth", so we might as well put them all under the sys category.

Thank you.

giorgos-apo:
@ktsakalozos  You are right about the output issues. I 've fixed both errors. Please check.

ktsakalozos:
LGTM +1 Thank you @giorgos-apo

PR #577, opened 2019-08-04
Adds link to medium story (flask + microk8s)

giorgos-apo:
Add links for medium stories, in which we execute a simple flask application as a microk8s service.


ktsakalozos:
Thank you @giorgos-apo

PR #578, opened 2019-08-04
"enable-admission-plugins" is not needed anymore

giner:
Starting from k8s 1.10 default plugins are enabled unless disabled by "disable-admission-plugins" so "enable-admission-plugins" can be removed.

> For Kubernetes version 1.10 and later, the recommended admission controllers are enabled by default (shown here), so you do not need to explicitly specify them. You can enable additional admission controllers beyond the default set using the --enable-admission-plugins flag (order doesn’t matter).

Links:
- https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#is-there-a-recommended-set-of-admission-controllers-to-use
- https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/#options


ktsakalozos:
LGTM +1. Less code less bugs :)

PR #580, opened 2019-08-05
Build from source

ktsakalozos:
With this PR we can build k8s binaries from a branch in any repository.

We also make sure we remove files that do not need to be in the snap package.


battlemidget:
I dont understand why you wouldn't want to use the binaries we already build from source on launchpad? You can just easily use `stage-snaps` at that point and be pulling from the same builds CDK runs.

ktsakalozos:
> You can just easily use `stage-snaps` at that point and be pulling from the same builds CDK runs

That would be fantastic! I am +1.

However based on a roadmap item we picked up these days we may want to push some patches to MicroK8s only. Let's sync up tomorrow.

battlemidget:
> That would be fantastic! I am +1.
> 
Great!

> However based on a roadmap item we picked up these days we may want to push some patches to MicroK8s only. Let's sync up tomorrow.

Sure! I've got a PoC for including upstream patches into our builds, I think we can expand that to allow microk8s specific patches as well.

PR #587, opened 2019-08-07
Cache kube binaries after download

joestringer:
This PR adds a quick-and-dirty cache for storing downloaded kubernetes
binaries under `parts/cache/kube-$KUBE_VERSION-$ARCH/` after the
first time that they're downloaded.

Otherwise every time you do `snapcraft clean microk8s` and `snapcraft`
it has to spend several minutes downloading these binaries again.

### DIFF
--- a/build-scripts/fetch-binaries.sh
+++ b/build-scripts/fetch-binaries.sh
@@ -10,9 +10,17 @@ echo $KUBE_VERSION > $KUBE_SNAP_BINS/version
     mkdir -p $KUBE_ARCH
     (cd $KUBE_ARCH
       echo "Fetching $app $KUBE_VERSION $KUBE_ARCH"
-      curl -LO \
-        https://dl.k8s.io/${KUBE_VERSION}/bin/linux/$KUBE_ARCH/$app
-      chmod +x $app
+      cachepath="../../../../../../cache/kube-$KUBE_VERSION-$KUBE_ARCH"
+      if [ -e $cachepath/$app ]; then
+        echo "Copying $app from local cache"
+        cp $cachepath/$app $app
+      else
+        curl -LO \
+          https://dl.k8s.io/${KUBE_VERSION}/bin/linux/$KUBE_ARCH/$app
+        chmod +x $app
+        mkdir -p $cachepath
+        cp $app $cachepath/$app
+      fi
       if ! file ${app} 2>&1 | grep -q 'executable'; then
         echo "${app} is not an executable"
         exit 1


ktsakalozos:
Hi @joestringer, this is a great PR! We have seen the need to skip the download step since the early development days of MicroK8s but we kept it a secret for some time and pease let me explain why. You may have seen our efforts to speedup compiling times in  https://github.com/ubuntu/microk8s/blob/master/snapcraft.yaml#L170. 

Let me first explain how this feature works. On the first build of MicroK8s all binaries (K8s included) will be fetched under `build/kube_bins/$KUBE_VERSION` (see https://github.com/ubuntu/microk8s/blob/master/build-scripts/fetch-binaries.sh#L5). After the build you can copy this directory to a desired location and reuse it by compiling with:
```
KUBE_SNAP_BINS="path/to/binaries" snapcraft
```
As you can see in https://github.com/ubuntu/microk8s/blob/master/snapcraft.yaml#L170 no binaries will be fetched and your build will be much faster.

Here is why we do not talk much about this feature. When you build the snap you are compiling a few components. These components need to be compiled against the snapcraft base core that MicroK8s depends on. The proper way to do this is to compile with `snapcraft cleanbuild` as that will start an LXC container with exactly what is available in the base core and do the build inside there. If you build on your system with a plain `snapcraft` any gaps on your build will be filled with local libraries and the snap will probably work on your system only. Try this for example, do a `snapcraft` on a bionic machine and try to deploy on a xenial, this is failing last time I checked. So, given the recommended way to build a snap is in an ephemeral container created with `snapcraft cleanbuild` you realize why these kind of caching tricks do not work. There is no way to pass an environment variable (pointing to a directory) in the build process running inside the LXC container. Instead of risking confusing the devs we promote the `snapcraft cleanbuild` and we do not talk about `KUBE_SNAP_BINS`.

Now that you know the full story here is how you can skip the long build times. As described in https://github.com/ubuntu/microk8s/blob/feature/build-from-source/docs/build.md#building-a-custom-microk8s-package first create a xenial (ubuntu 16.04) LXC container to act as your build machine. That container has to be xenial because this is the base core MicroK8s builds on top. Inside the container clone the MicroK8s repository, do a plain build with just `snapcraft`, copy the `build/kube_bins/$KUBE_VERSION` somewhere inside the container and from then on keep building by providing the `KUBE_SNAP_BINS` environment variable like so `KUBE_SNAP_BINS="path/to/binaries" snapcraft`

Your suggestion with this PR is very similar to the `KUBE_SNAP_BINS` approach as it automates the copy step. On the other hand a `snapcraft clean` does not have the expected result, you still reuse cached artifacts from previous builds (`"../../../../../../cache/kube-$KUBE_VERSION-$KUBE_ARCH"`).

joestringer:
Awesome, thanks for the explanation, I can close my PR then since the feature is already here :-)

PR #590, opened 2019-08-09
Harden MicroK8s

ktsakalozos:
- Removing the 8080 port on *new* installations. It was used in more places that expected.
- etcd listens non port 12379 and accessed through certificates.
- strict permissions on certificates, creds, and args directories

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -1,5 +1,13 @@
 #!/usr/bin/env bash
 
+exit_if_no_sudo() {
+  # test if we can access the default kubeconfig
+  if ! cat $SNAP_DATA/credentials/client.config 2>&1 > /dev/null; then
+    echo "You do not have enough permissions to access MicroK8s. Please try again with sudo."
+    exit 1
+  fi
+}
+
 exit_if_stopped() {
   # test if the snap is marked as stopped
   if [ -e ${SNAP_DATA}/var/lock/stopped.lock ]
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
-exit_if_no_sudo() {
+exit_if_no_permissions() {
   # test if we can access the default kubeconfig
-  if ! cat $SNAP_DATA/credentials/client.config > /dev/null 2>&1; then
+  if [ ! -r $SNAP_DATA/credentials/client.config ]; then
     echo "You do not have enough permissions to access MicroK8s. Please try again with sudo."
     exit 1
   fi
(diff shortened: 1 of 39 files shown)


ktsakalozos:
@tvansteenburgh @jdstrand @alexmurray, that is the changes we were discussing about. I would appreciate a review. Should I put out a branch you could get the snap from? Thank you.

microk8s-resources/actions/common/utils.sh
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env bash
 
+exit_if_no_sudo() {
+  # test if we can access the default kubeconfig
+  if ! cat $SNAP_DATA/credentials/client.config 2>&1 > /dev/null; then
```
alexmurray: This would be better as:
    if [ ! -r $SNAP_DATA/credentials/client.config ]; then

microk8s-resources/actions/common/utils.sh
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env bash
 
+exit_if_no_sudo() {
```
alexmurray: Strictly speaking this name is a bit misleading - might be better to call it `exit_if_no_permission` perhaps?

microk8s-resources/wrappers/microk8s-istioctl.wrapper
```diff
@@ -11,6 +11,11 @@ if [ ! -f "${SNAP_DATA}/bin/istioctl" ]; then
 fi
 
 source $SNAP/actions/common/utils.sh
+
+if echo "$*" | grep -v -q -- '--kubeconfig'; then
```
alexmurray: Should this be abstracted out to a separate function in microk8s-resources/actions/common/utils.sh?

ktsakalozos: Can I please keep this? I like this oneliner better than creating a function. I know it is in a couple of places but it is a singe call to grep.

microk8s-resources/actions/common/utils.sh
ktsakalozos: Done

microk8s-resources/actions/common/utils.sh
ktsakalozos: Thanks

microk8s-resources/wrappers/microk8s-istioctl.wrapper
alexmurray: Sure.

amozoss:
Is there a way to use microk8s without sudo, aka how it worked before this was merged?

alexmurray:
@ktsakalozos I thought we had discussed using an approach similar to lxd / libvirt, where the certificates etc would be group readable to a new `microk8s` group and so a user could add themselves to that group so they could then use microk8s without having to use sudo?

ktsakalozos:
@amozoss , @alexmurray let me know what you think about this: https://github.com/ubuntu/microk8s/pull/608

PR #592, opened 2019-08-13
Configure iptable forwarding policy for pods

ktsakalozos:
This is best described in: https://www.altoros.com/blog/kubernetes-networking-writing-your-own-simple-cni-plug-in-with-bash/


ktsakalozos:
Merging this :)

PR #594, opened 2019-08-13
Add helm, cilium addons

joestringer:
## Summary

Add support for enabling/disabling Cilium directly from microk8s. This should greatly simplify the instructions in the current [Cilium getting started guide](http://docs.cilium.io/en/latest/gettingstarted/microk8s/) to just:
1. `snap install microk8s --classic`
1. `microk8s.enable cilium`

This also provides a useful wrapper, `microk8s.cilium` which provides direct CLI access to Cilium without needing to open a shell into the Cilium container.

To install cilium, there are helm charts provided by the Cilium repository. So this PR also provides a way to install helm within the microk8s environment for use from Cilium.

### Installing
```
$ sudo snap run --shell  microk8s.enable
# ./microk8s-resources/actions/enable.cilium.sh
Enabling Helm
Helm is enabled
Restarting kube-apiserver
Enabling Cilium
Fetching cilium version v1.6.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   119    0   119    0     0    339      0 --:--:-- --:--:-- --:--:--   338
100 21.8M    0 21.8M    0     0  1199k      0 --:--:--  0:00:18 --:--:-- 1264k
configmap/cilium-config created
serviceaccount/cilium created
serviceaccount/cilium-operator created
clusterrole.rbac.authorization.k8s.io/cilium created
clusterrole.rbac.authorization.k8s.io/cilium-operator created
clusterrolebinding.rbac.authorization.k8s.io/cilium created
clusterrolebinding.rbac.authorization.k8s.io/cilium-operator created
daemonset.apps/cilium created
deployment.apps/cilium-operator created
Waiting for daemon set "cilium" rollout to finish: 0 of 1 updated pods are available...
daemon set "cilium" successfully rolled out
Restarting kubelet
Restarting containerd
Cilium is enabled

```

### CLI usage

```
# ./microk8s-resources/wrappers/microk8s-cilium.wrapper status
KVStore:                Ok   Disabled                 
ContainerRuntime:       Ok   containerD events watcher: Ok; cri-containerd client: Ok - cri daemon: Ok
Kubernetes:             Ok   1.15 (v1.15.2) [linux/amd64]                             
Kubernetes APIs:        ["CustomResourceDefinition", "cilium/v2::CiliumEndpoint", "cilium/v2::CiliumNetworkPolicy", "cilium/v2::CiliumNode", "core/v1::Endpoint", "core/v1::Namespace", "core/v1::Pods", "core/v1::Service", "networking.k8s.io/v1::NetworkPolicy"]
Cilium:                 Ok   OK                                                                       
NodeMonitor:            Disabled       
Cilium health daemon:   Ok                       
IPAM:                   IPv4: 11/255 allocated from 10.1.1.0/24,               
Controller Status:      53/53 healthy                                        
Proxy Status:           OK, ip 10.1.1.254, port-range 10000-20000                           
Cluster health:   1/1 reachable   (2019-08-13T21:22:50Z)
```

### Uninstalling
```
# ./microk8s-resources/actions/disable.cilium.sh
Disabling Cilium
configmap "cilium-config" deleted
serviceaccount "cilium" deleted
serviceaccount "cilium-operator" deleted
clusterrole.rbac.authorization.k8s.io "cilium" deleted
clusterrole.rbac.authorization.k8s.io "cilium-operator" deleted
clusterrolebinding.rbac.authorization.k8s.io "cilium" deleted
clusterrolebinding.rbac.authorization.k8s.io "cilium-operator" deleted
daemonset.apps "cilium" deleted
deployment.apps "cilium-operator" deleted
Restarting kubelet
Restarting containerd
Cilium is terminating
```

### Live testing

```
$ pytest -k cilium -s test-live-addons.py
============================================================ test session starts ============================================================
platform linux2 -- Python 2.7.16, pytest-3.10.1, py-1.7.0, pluggy-0.8.0
rootdir: /home/joe/git/microk8s/tests, inifile:
collected 10 items / 9 deselected                                                                                                           

test-live-addons.py .

================================================== 1 passed, 9 deselected in 37.91 seconds ==================================================
```

### Upgrade

```
$ UPGRADE_MICROK8S_FROM=$PWD/microk8s_v1.15.2_amd64.snap UPGRADE_MICROK8S_TO=$PWD/microk8s_v1.15.2_amd64.snap py                                     
test -s ./tests/test-upgrade.py
==================================================================== test session starts =====================================================================                                     
platform linux2 -- Python 2.7.16, pytest-3.10.1, py-1.7.0, pluggy-0.8.0
rootdir: /home/joe/git/microk8s, inifile:
collected 1 item                                                                                                                                                                                   

tests/test-upgrade.py Testing upgrade from /home/joe/git/microk8s/microk8s_v1.15.2_amd64.snap to /home/joe/git/microk8s/microk8s_v1.15.2_amd64.snap                                                
[sudo] password for joe:
The connection to the server 127.0.0.1:8080 was refused - did you specify the right host or port?
Retrying /snap/bin/microk8s.kubectl get -o yaml svc kubernetes
systemd-detect-virt did not detect a container
no indication of a container in /proc
Testing cilium
systemd-detect-virt did not detect a container
no indication of a container in /proc
```

## Remaining tasks
* [x] Check that Travis is happy
* [x] Write tests for addon integration
* [x] Fix version for full v1.6 release instead of master (Pulling from v1.6 branch)
* [x] Attempt installation in a fresh environment
* [x] Resolve `cilium-operator` API access issue (Cannot reproduce in fresh AWS Ubuntu VM)
* [ ] Resolve question about CLI tool permissions

### DIFF
--- a/microk8s-resources/actions/disable.cilium.sh
+++ b/microk8s-resources/actions/disable.cilium.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+set -e
+
+source $SNAP/actions/common/utils.sh
+
+echo "Disabling Cilium"
+
+if [ -L "${SNAP_DATA}/bin/cilium" ]
+then
+  "$SNAP/kubectl" "--kubeconfig=$SNAP_DATA/credentials/client.config" delete -f "$SNAP_DATA/actions/cilium.yaml"
+
+  sudo rm -f "$SNAP_DATA/args/cni-network/05-cilium-cni.conf"
+  sudo rm -f "$SNAP_DATA/opt/cni/bin/cilium-cni"
+  sudo rm -rf $SNAP_DATA/bin/cilium*
+  sudo rm -f "$SNAP_DATA/actions/cilium.yaml"
+  sudo rm -rf "$SNAP_DATA/actions/cilium"
+  sudo rm -rf "$SNAP_DATA/var/run/cilium"
+  sudo rm -rf "$SNAP_DATA/sys/fs/bpf"
+
+  echo "Restarting kubelet"
+  refresh_opt_in_config "network-plugin" "kubenet" kubelet
+  refresh_opt_in_config "cni-bin-dir" "\${SNAP}/opt/cni/bin/" kubelet
+  sudo systemctl restart snap.${SNAP_NAME}.daemon-kubelet
+  echo "Restarting containerd"
+  if ! grep -qE "bin_dir.*SNAP}\/" $SNAP_DATA/args/containerd-template.toml; then
+    sudo "${SNAP}/bin/sed" -i 's;bin_dir = "${SNAP_DATA}/opt;bin_dir = "${SNAP}/opt;g' "$SNAP_DATA/args/containerd-template.toml"
+  fi
+  sudo systemctl restart snap.${SNAP_NAME}.daemon-containerd
+
+  echo "Cilium is terminating"
+  cilium=$(wait_for_service_shutdown "kube-system" "k8s-app=cilium")
+  if [[ $cilium == fail ]]
+  then
+    echo "Cilium did not shut down on time. Proceeding."
+  fi
+fi
--- a/microk8s-resources/actions/enable.cilium.sh
+++ b/microk8s-resources/actions/enable.cilium.sh
@@ -1,0 +1,95 @@
+#!/usr/bin/env bash
+
+set -e
+
+source $SNAP/actions/common/utils.sh
+
+ARCH=$(arch)
+if ! [ "${ARCH}" = "amd64" ]; then
+  echo "Cilium is not available for ${ARCH}" >&2
+  exit 1
+fi
+
+"$SNAP/microk8s-enable.wrapper" helm
+
+echo "Restarting kube-apiserver"
+refresh_opt_in_config "allow-privileged" "true" kube-apiserver
+sudo systemctl restart snap.${SNAP_NAME}.daemon-apiserver
+
+# Reconfigure kubelet/containerd to pick up the new CNI config and binary.
+echo "Restarting kubelet"
+refresh_opt_in_config "network-plugin" "cni" kubelet
+refresh_opt_in_config "cni-bin-dir" "\${SNAP_DATA}/opt/cni/bin/" kubelet
+sudo systemctl restart snap.${SNAP_NAME}.daemon-kubelet
+
+if grep -qE "bin_dir.*SNAP}\/" $SNAP_DATA/args/containerd-template.toml; then
+  echo "Restarting containerd"
+  sudo "${SNAP}/bin/sed" -i 's;bin_dir = "${SNAP}/opt;bin_dir = "${SNAP_DATA}/opt;g' "$SNAP_DATA/args/containerd-template.toml"
+  sudo systemctl restart snap.${SNAP_NAME}.daemon-containerd
+fi
+
+echo "Enabling Cilium"
+
+read -ra CILIUM_VERSION <<< "$1"
+if [ -z "$CILIUM_VERSION" ]; then
+  CILIUM_VERSION="v1.6"
+fi
+CILIUM_ERSION=$(echo $CILIUM_VERSION | sed 's/v//g')
+
+if [ -f "${SNAP_DATA}/bin/cilium-$CILIUM_ERSION" ]
+then
+  echo "Cilium version $CILIUM_VERSION is already installed."
+else
+  CILIUM_DIR="cilium-$CILIUM_ERSION"
+  SOURCE_URI="https://github.com/cilium/cilium/archive"
+  CILIUM_CNI_CONF="plugins/cilium-cni/05-cilium-cni.conf"
+  CILIUM_LABELS="k8s-app=cilium"
+  NAMESPACE=kube-system
+
+  echo "Fetching cilium version $CILIUM_VERSION."
+  sudo mkdir -p "${SNAP_DATA}/tmp/cilium"
+  (cd "${SNAP_DATA}/tmp/cilium"
+  sudo "${SNAP}/usr/bin/curl" -L $SOURCE_URI/$CILIUM_VERSION.tar.gz -o "$SNAP_DATA/tmp/cilium/cilium.tar.gz"
+  if ! sudo gzip -f -d "$SNAP_DATA/tmp/cilium/cilium.tar.gz"; then
+    echo "Invalid version \"$CILIUM_VERSION\". Must be a branch on https://github.com/cilium/cilium."
+    exit 1
+  fi
+  sudo tar -xf "$SNAP_DATA/tmp/cilium/cilium.tar" "$CILIUM_DIR/install" "$CILIUM_DIR/$CILIUM_CNI_CONF")
+
+  sudo mv "$SNAP_DATA/args/cni-network/cni.conf" "$SNAP_DATA/args/cni-network/10-kubenet.conf" 2>/dev/null || true
+  sudo cp "$SNAP_DATA/tmp/cilium/$CILIUM_DIR/$CILIUM_CNI_CONF" "$SNAP_DATA/args/cni-network/05-cilium-cni.conf"
+
+  # Generate the YAMLs for Cilium and apply them
+  (cd "${SNAP_DATA}/tmp/cilium/$CILIUM_DIR/install/kubernetes"
+  ${SNAP_DATA}/bin/helm template cilium \
+      --namespace $NAMESPACE \
+      --set global.cni.confPath="$SNAP_DATA/args/cni-network" \
+      --set global.cni.binPath="$SNAP_DATA/opt/cni/bin" \
+      --set global.cni.customConf=true \
+      --set global.containerRuntime.integration="containerd" \
+      --set global.containerRuntime.socketPath="$SNAP_COMMON/run/containerd.sock" \
+      | sudo tee cilium.yaml >/dev/null)
+
+  sudo mkdir -p "$SNAP_DATA/actions/cilium/"
+  sudo cp "$SNAP_DATA/tmp/cilium/$CILIUM_DIR/install/kubernetes/cilium.yaml" "$SNAP_DATA/actions/cilium.yaml"
+  sudo sed -i 's;path: \(/var/run/cilium\);path: '"$SNAP_DATA"'\1;g' "$SNAP_DATA/actions/cilium.yaml"
+
+  microk8s.status --wait-ready >/dev/null
+  echo "Deploying $SNAP_DATA/actions/cilium.yaml. This may take several minutes."
+  "$SNAP/kubectl" "--kubeconfig=$SNAP_DATA/credentials/client.config" apply -f "$SNAP_DATA/actions/cilium.yaml"
+  "$SNAP/kubectl" "--kubeconfig=$SNAP_DATA/credentials/client.config" -n $NAMESPACE rollout status ds/cilium
+
+  # Fetch the Cilium CLI binary and install
+  CILIUM_POD=$("$SNAP/kubectl" "--kubeconfig=$SNAP_DATA/credentials/client.config" -n $NAMESPACE get pod -l $CILIUM_LABELS -o jsonpath="{.items[0].metadata.name}")
+  CILIUM_BIN=$(mktemp)
+  "$SNAP/kubectl" "--kubeconfig=$SNAP_DATA/credentials/client.config" -n $NAMESPACE cp $CILIUM_POD:/usr/bin/cilium $CILIUM_BIN >/dev/null
+  sudo mkdir -p "$SNAP_DATA/bin/"
+  sudo mv $CILIUM_BIN "$SNAP_DATA/bin/cilium-$CILIUM_ERSION"
+  sudo chmod +x "$SNAP_DATA/bin/"
+  sudo chmod +x "$SNAP_DATA/bin/cilium-$CILIUM_ERSION"
+  sudo ln -s $SNAP_DATA/bin/cilium-$CILIUM_ERSION $SNAP_DATA/bin/cilium
+
+  sudo rm -rf "$SNAP_DATA/tmp/cilium"
+fi
+
+echo "Cilium is enabled"
--- a/microk8s-resources/wrappers/microk8s-cilium.wrapper
+++ b/microk8s-resources/wrappers/microk8s-cilium.wrapper
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -eu
+
+export PATH="$SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH"
+export LD_LIBRARY_PATH="$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/x86_64-linux-gnu:$SNAP/usr/lib/x86_64-linux-gnu"
+
+if [ ! -f "${SNAP_DATA}/bin/cilium" ]; then
+  echo "Cilium not available, try enabling it with 'microk8s.enable cilium'" >&2
+  exit 0
+fi
+
+source $SNAP/actions/common/utils.sh
+exit_if_stopped
+export CILIUM_MONITOR_SOCK="$SNAP_DATA/var/run/cilium/monitor1_2.sock"
+"${SNAP_DATA}/bin/cilium" -H "unix:///var/snap/microk8s/current/var/run/cilium/cilium.sock" "$@"
--- a/microk8s-resources/wrappers/microk8s-helm.wrapper
+++ b/microk8s-resources/wrappers/microk8s-helm.wrapper
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -eu
+
+export PATH="$SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH"
+export LD_LIBRARY_PATH="$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/x86_64-linux-gnu:$SNAP/usr/lib/x86_64-linux-gnu"
+
+if [ ! -f "${SNAP_DATA}/bin/helm" ]; then
+  echo "Helm not available, try enabling it with 'microk8s.enable helm'" >&2
+  exit 0
+fi
+
+source $SNAP/actions/common/utils.sh
+"${SNAP_DATA}/bin/helm" --kubeconfig="${SNAP_DATA}"/credentials/client.config "$@"
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -18,7 +18,7 @@ refresh_opt_in_config() {
     local config_file="$SNAP_DATA/args/$3"
     local replace_line="$opt=$value"
     if $(grep -qE "^$opt=" $config_file); then
-        sudo "$SNAP/bin/sed" -i "s/^$opt=.*/$replace_line/" $config_file
+        sudo "$SNAP/bin/sed" -i "s;^$opt=.*;$replace_line;" $config_file
     else
         sudo "$SNAP/bin/sed" -i "$ a $replace_line" "$config_file"
     fi
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -118,6 +118,26 @@ wait_for_service() {
     fi
 }
 
+wait_for_service_shutdown() {
+    # Wait for a service to stop
+    # Return  fail if the service did not stop in 30 seconds
+
+    local namespace="$1"
+    local labels="$2"
+    local shutdown_timeout=30
+    local start_timer="$(date +%s)"
+    KUBECTL="$SNAP/kubectl --kubeconfig=$SNAP/client.config"
+
+    while ($KUBECTL get po -n "$namespace" -l "$labels" | grep -z " Terminating") &> /dev/null
+    do
+      now="$(date +%s)"
+      if [[ "$now" > "$(($start_timer + $shutdown_timeout))" ]] ; then
+        echo "fail"
+        break
+      fi
+      sleep 5
+    done
+}
 
 get_default_ip() {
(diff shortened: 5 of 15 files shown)


joestringer:
Travis is happy I didn't break anything:
https://travis-ci.org/ubuntu/microk8s/builds/571539643?utm_source=github_status&utm_medium=notification

Will try writing up a basic test.

joestringer:
Hmm, from the [latest Travis build](https://travis-ci.org/ubuntu/microk8s/builds/571557551?utm_source=github_status&utm_medium=notification):

```
Skipping cilium tests as we are under time pressure
```

joestringer:
I was able to validate the test locally:

```
$ pytest -k cilium -s test-live-addons.py
============================================================ test session starts ============================================================
platform linux2 -- Python 2.7.16, pytest-3.10.1, py-1.7.0, pluggy-0.8.0
rootdir: /home/joe/git/microk8s/tests, inifile:
collected 10 items / 9 deselected                                                                                                           

test-live-addons.py .

================================================== 1 passed, 9 deselected in 37.91 seconds ==================================================
```

microk8s-resources/actions/enable.cilium.sh
```diff
@@ -0,0 +1,83 @@
...
+  sudo gzip -f -d "$SNAP_DATA/tmp/cilium/cilium.tar.gz"
+  sudo tar -xf "$SNAP_DATA/tmp/cilium/cilium.tar" "$CILIUM_DIR/install" "$CILIUM_DIR/$CILIUM_CNI_CONF")
+
+  sudo mv "$SNAP_DATA/args/cni-network/cni.conf" "$SNAP_DATA/args/cni-network/10-kubenet.conf" 2>/dev/null || true
+  sudo cp "$SNAP_DATA/tmp/cilium/$CILIUM_DIR/$CILIUM_CNI_CONF" "$SNAP_DATA/args/cni-network/05-cilium-cni.conf"
+
+  (cd "${SNAP_DATA}/tmp/cilium/$CILIUM_DIR/install/kubernetes"
+  helm template cilium \
+      --namespace $NAMESPACE \
+      --set global.cni.confPath="$SNAP_DATA/args/cni-network" \
```
ktsakalozos: Could we also set the `global.cni.binPath` so that the binary lands inside a directory owned by MicroK8s? We try not to pollute the host. Something like `$SNAP_DATA/opt/cni/bin` or `$SNAP_COMMON/opt/cni/bin` would be great.

microk8s-resources/actions/enable.cilium.sh
```diff
@@ -0,0 +1,83 @@
...
+      --set global.cni.customConf=true \
+      --set global.containerRuntime.integration="containerd" \
+      --set global.containerRuntime.socketPath="$SNAP_COMMON/run/containerd.sock" \
+      | sudo tee cilium.yaml >/dev/null)
+
+  sudo mkdir -p "$SNAP_DATA/actions/cilium/"
+  sudo cp "$SNAP_DATA/tmp/cilium/$CILIUM_DIR/install/kubernetes/cilium.yaml" "$SNAP_DATA/actions/cilium.yaml"
+
+  "$SNAP/kubectl" "--kubeconfig=$SNAP/client.config" apply -f "$SNAP_DATA/actions/cilium.yaml"
```
ktsakalozos: Please take the kubeconfig file found under credentials (`--kubeconfig=$SNAP_DATA/credentials/client.config`), the one you are using is kept there for backwards compatibility reasons and will soon be removed.

microk8s-resources/wrappers/microk8s-cilium.wrapper
```diff
@@ -0,0 +1,15 @@
...
+export LD_LIBRARY_PATH="$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/x86_64-linux-gnu:$SNAP/usr/lib/x86_64-linux-gnu"
+
+if [ ! -f "${SNAP_DATA}/bin/cilium" ]; then
+  echo "Cilium not available, try enabling is with 'microk8s.enable cilium'" >&2
+  exit 0
+fi
+
+source $SNAP/actions/common/utils.sh
+exit_if_stopped
+"${SNAP_DATA}/bin/cilium" "$@"
```
ktsakalozos: Would you trust any local user to access cilium, or does the user need to be a sudoer?

microk8s-resources/actions/enable.cilium.sh
```diff
@@ -0,0 +1,83 @@
...
+set -e
+
+source $SNAP/actions/common/utils.sh
+
+if ! which helm 2>&1 >/dev/null; then
+	echo "Cilium installation requires helm. Please install via 'snap install helm'." >&2
+	exit 1
+fi
+
+ARCH=$(arch)
```
ktsakalozos: I would move the architecture check before the helm warning.

microk8s-resources/actions/enable.cilium.sh
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env bash
+
+set -e
+
+source $SNAP/actions/common/utils.sh
+
+if ! which helm 2>&1 >/dev/null; then
```
ktsakalozos: I would prefer to have helm part of MicroK8s instead of asking the user to install it manually. It is a better experience plus it is less risky if we ensure we deploy and configure a helm that we know it works.

For now, this approach is fine. Adding helm could be handled in a separate PR.

ktsakalozos:
Thank you for this PR @joestringer 

When we run tests on Travis we have a time limit of 50 minutes and this is where "time pressure" comes from. When releasing to beta, candidate and stable channels we run all the tests.

There is also the upgrade path that you may want to write tests for. As you may have noticed snaps get upgraded when a new release comes out. We do not want to break people when releasing and the https://github.com/ubuntu/microk8s/blob/master/tests/test-upgrade.py takes care of that. That test is called like so: `UPGRADE_MICROK8S_FROM=edge UPGRADE_MICROK8S_TO="/path/to/your/local/MicroK8s.snap" pytest -s ./tests/test-upgrade.py` In the code we snap install the snap from  `UPGRADE_MICROK8S_FROM` we enable any addons we want to see they survive the upgrade, then we snap install the snap from `UPGRADE_MICROK8S_TO` and we run the validators so we know what used to work is still fine. So if you want to ensure the addon will survive through releases you could consider adding a test there.

microk8s-resources/actions/enable.cilium.sh
```diff
@@ -0,0 +1,83 @@
...
+  sudo mkdir -p "$SNAP_DATA/bin/"
+  sudo mv $CILIUM_BIN "$SNAP_DATA/bin/cilium"
+  sudo chmod +x "$SNAP_DATA/bin/"
+  sudo chmod +x "$SNAP_DATA/bin/cilium"
+
+  sudo rm -rf "$SNAP_DATA/tmp/cilium"
+fi
+
+echo "Restarting kube-apiserver"
+refresh_opt_in_config "allow-privileged" "true" kube-apiserver
```
ktsakalozos: I think configuring the apiserver should come before trying to apply the manifest. Here is what i got:
```
$ microk8s.enable cilium
Enabling Cilium
Fetching cilium version master.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   121    0   121    0     0    614      0 --:--:-- --:--:-- --:--:--   614
100 21.8M  100 21.8M    0     0  45.5M      0 --:--:-- --:--:-- --:--:--  118M
configmap/cilium-config unchanged
serviceaccount/cilium unchanged
serviceaccount/cilium-operator unchanged
clusterrole.rbac.authorization.k8s.io/cilium unchanged
clusterrole.rbac.authorization.k8s.io/cilium-operator unchanged
clusterrolebinding.rbac.authorization.k8s.io/cilium unchanged
clusterrolebinding.rbac.authorization.k8s.io/cilium-operator unchanged
deployment.apps/cilium-operator configured
The DaemonSet "cilium" is invalid: 
* spec.template.spec.containers[0].securityContext.privileged: Forbidden: disallowed by cluster policy
* spec.template.spec.initContainers[0].securityContext.privileged: Forbidden: disallowed by cluster policy
Failed to enable cilium
``` 
That is mentioned here: https://gist.github.com/borkmann/84812fcac624fb659f66f0d514faa1c8#problems-deploying-cilium-daemon-set

ktsakalozos: And we probably need a `microk8s.status --wait-ready` before trying to apply the yaml manifests.

ktsakalozos:
We need to update the `microk8s.status` command to test that the addon is enabled or not: https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/microk8s-status.wrapper#L13

ktsakalozos:
I got this error when trying this addon:
```
$ microk8s.kubectl logs -n kube-system   pod/cilium-operator-69575d5f8-rr96l
level=info msg="Cilium Operator " subsys=cilium-operator
level=info msg="Starting apiserver on address :9234" subsys=cilium-operator
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=info msg="Establishing connection to apiserver" host="https://10.152.183.1:443" subsys=k8s
level=error msg="Unable to contact k8s api-server" error="Get https://10.152.183.1:443/api/v1/componentstatuses/controller-manager: dial tcp 10.152.183.1:443: connect: connection refused" ipAddr="https://10.152.183.1:443" subsys=k8s
level=fatal msg="Unable to connect to Kubernetes apiserver" error="unable to create k8s client: unable to create k8s client: Get https://10.152.183.1:443/api/v1/componentstatuses/controller-manager: dial tcp 10.152.183.1:443: connect: connection refused" subsys=cilium-operator
```
With the pod to be crashlooping.

microk8s-resources/actions/enable.cilium.sh
ktsakalozos: Our CI scripts need to install and configure helm so this passes testing. We run tests in lxc containers https://github.com/ubuntu/microk8s/blob/master/tests/test-distro.sh so the scripts for preparing these containers will need to be updated as well https://github.com/ubuntu/microk8s/tree/master/tests/lxc/install-deps.

ktsakalozos: It might be easier to add a `microk8s.enable helm` that would provide a `microk8s.helm` command. Do you think you could provide that as part of this PR?

microk8s-resources/actions/enable.cilium.sh
joestringer: Is there a path that's writable? Cilium usually writes the CNI binary into `cni.binPath` in an init container to ensure it's installed and up to date. Alternatively I guess we could look at allowing Cilium to disable that option and manually install it into the right place in these scripts.

microk8s-resources/wrappers/microk8s-cilium.wrapper
joestringer: In typical clusters I note that the `/var/run/cilium/cilium.sock` has file mode permissions `660` with `user=root` and `group=cilium`. When I deploy Cilium via this script in microk8s right now, it's inheriting the group of the user that creates the YAMLs.

Maybe there's something I can do here to ensure that there is a cilium group on the system and that the container is run under the cilium group?

That said, access to the cilium binary allows modifying network policies for pods/endpoints, but then at the same time the same user can presumably just run `kubectl delete netpol ...` and delete policies so I don't see a real distinction there.

microk8s-resources/actions/enable.cilium.sh
joestringer: I can take a stab at this.

microk8s-resources/actions/enable.cilium.sh
joestringer: Thanks for trying this out, I'm migrating from an existing setup that I had working to this so it's easy for me to miss stuff like this. I'll reorder the apiserver configuration.

joestringer:
@ktsakalozos Is the APIServer running on that IP/Port? The Cilium pod picks this up from the environment, so I'm surprised that `cilium-operator` gets stuck there, unless the cilium addon install steps are messing up the restart of the apiserver.

microk8s-resources/actions/enable.cilium.sh
joestringer: Looking at just disabling the write into the CNI binary path: https://github.com/cilium/cilium/pull/8913

joestringer:
@ktsakalozos I fixed up a bunch of things, but haven't come to a conclusion about the CLI yet. I also wasn't able to reproduce your issue with the apiserver connectivity when I installed on a fresh Ubuntu image in AWS so I'm not sure what's going wrong there. I have some suggestions to dig further in my comment above. If you get a chance I'd like to hear your feedback on the above and the latest version.

microk8s-resources/actions/enable.cilium.sh
ktsakalozos: Disabling the cni install looks great. I think that if we `--set global.cni.binPath="$SNAP_DATA/opt/cni/bin"` we will have similar results.

microk8s-resources/wrappers/microk8s-cilium.wrapper
ktsakalozos: Is there a way to set the path to the `cilium.sock`? We would rather not create the `/var/run/cilium/cilium.sock` in that path but instead place the socket inside $SNAP_DATA.

microk8s-resources/wrappers/microk8s-helm.wrapper
```diff
@@ -0,0 +1,14 @@
...
+export PATH="$SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH"
+export LD_LIBRARY_PATH="$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/x86_64-linux-gnu:$SNAP/usr/lib/x86_64-linux-gnu"
+
+if [ ! -f "${SNAP_DATA}/bin/helm" ]; then
+  echo "Helm not available, try enabling it with 'microk8s.enable helm'" >&2
+  exit 0
+fi
+
+source $SNAP/actions/common/utils.sh
+"${SNAP_DATA}/bin/helm" "$@"
```
ktsakalozos: I think we should specify the `--kubeconfig` argument here so that `microk8s.helm` will talk by default to microk8s. Similar to what we do in:  https://github.com/ubuntu/microk8s/blob/feature/insecure-port/microk8s-resources/wrappers/microk8s-linkerd.wrapper#L25

ktsakalozos:
@joestringer, the not being able to access the api server issue was an issue with my setup. Fixed it and now cilium works fine. Just a couple more easy fixes. Nice work so far!

microk8s-resources/wrappers/microk8s-cilium.wrapper
joestringer: Ah, I was generally under the impression that `$SNAP_DATA` isn't writable at runtime but it seems like I was just mixing it up with `$SNAP`. It looks like things like the socket path are all configurable, but will take a bit of trial and error to validate/fix them up.

microk8s-resources/actions/enable.cilium.sh
```diff
@@ -0,0 +1,94 @@
...
+sudo systemctl restart snap.${SNAP_NAME}.daemon-apiserver
+
+echo "Enabling Cilium"
+
+if [ ! -f "${SNAP_DATA}/bin/cilium" ]
+then
+  CILIUM_VERSION="v1.6"
```
ktsakalozos: I see you have hard coded the version here. It is fine with me. In case you want to allow more flexibility users could provide the version while they enable the add-on. In linkerd for example you can do  `microk8s.enable "linkerd:proxy-auto-inject;tls=optional;skip-outbound-ports=1234,3456"` see https://github.com/ubuntu/microk8s/blob/feature/insecure-port/microk8s-resources/actions/enable.linkerd.sh#L7

joestringer: Oh, that's cool. The only trick is that this name has to be the same as a branch name on the Cilium repository. For my local testing, I've just been installing using these scripts, then manually editing the image tag to change it. This could be useful though.

microk8s-resources/actions/disable.cilium.sh
```diff
@@ -0,0 +1,29 @@
...
+
+set -e
+
+source $SNAP/actions/common/utils.sh
+
+echo "Disabling Cilium"
+
+if [ -f "${SNAP_DATA}/bin/cilium" ]
+then
+  "$SNAP/kubectl" "--kubeconfig=$SNAP_DATA/credentials/client.config" delete -f "$SNAP_DATA/actions/cilium.yaml"
```
ktsakalozos: Disabling left this behind `kube-system   pod/cilium-52mcz   0/1     Terminating`. This is probably because we do not wait for the pod to finish termination and we restart kubelet immediately. Could we do an active wait as we do here: https://github.com/ubuntu/microk8s/blob/feature/insecure-port/microk8s-resources/actions/disable.dns.sh#L26 The reason I am asking this is because our tests will hang waiting for the pod to terminate :(

microk8s-resources/wrappers/microk8s-cilium.wrapper
joestringer: The latest snag is that the `cilium-cni` binary needs to know the cilium socket. The way that it gets this is either from the `CILIUM_SOCK` environment variable, or defaults to `unix:///var/run/cilium/cilium.sock`. I'm currently trying to see if I can make the CILIUM_SOCK environment variable available for CNI invocation via `microk8s-resources/wrappers/run-with-config-args`.

joestringer: I made a change like this for kubelet case, but this doesn't seem to work:

```
diff --git a/microk8s-resources/wrappers/run-with-config-args b/microk8s-resources/wrappers/run-with-config-args
index ea7cb0aa1131..10214288cc23 100755
--- a/microk8s-resources/wrappers/run-with-config-args
+++ b/microk8s-resources/wrappers/run-with-config-args
@@ -39,6 +39,9 @@ then
     refresh_opt_in_config "kubelet-cgroups" "/systemd/system.slice" kubelet
   fi
 
+  # Export Cilium socket path correctly for CNI
+  export CILIUM_SOCK="unix://${SNAP_DATA}/var/run/cilium/cilium.sock"
+
 elif [ "${app}" = "kube-apiserver" ] || [ "${app}" = "kube-proxy" ]
 then
   # Check if we advertise an address. If we do we do not need to wait for a default network interface.
```

joestringer: Figured it out, needs to be in `microk8s-resources/wrappers/run-containerd-with-args`. I'll update the branch shortly.

ktsakalozos:
Great work @joestringer!

Thank you for Cilium. The addon should be on edge within the day and should reach stable with the v1.15.3.

joestringer:
Awesome, thanks @ktsakalozos !

ktsakalozos:
@joestringer we will soon replace kubenet with flannel (https://github.com/ubuntu/microk8s/pull/529/files#diff-578244e09bb2c511a1e4fd5798771af8). Will this affect this addon? I should refactor the add-on based on this: https://docs.cilium.io/en/v1.6/gettingstarted/flannel-integration/ ?

joestringer:
@ktsakalozos ah, interesting. There are basically two options. I'd recommend the first, but either will work.

### Replace Flannel with Cilium

We could rework it so that the cilium addon continues to just replace flannel. This is the simplest configuration and the configuration of Cilium that gets the most testing and validation.

Requirements:
* The flannel vxlan device should be deleted, and/or Cilium should be configured with `--tunnel=disabled` (this should be doable in the helm options in the script). Otherwise you hit problems because Linux can only have one vxlan device configured on the system at a time.
* The CNI configuration file should prefer Cilium over Flannel when enabled (`05-cilium-cni.conf` is alphabetically prior to the default `10-flannel.conflist` so this should just work).

### Apply Cilium on top of flannel

I've tried this approach as well, and it works, although some more elaborate use cases are still beta and have some problems (perhaps some L7 policy cases, I can't recall what else but there are a few failures in our CI at the moment in this mode).

Requirements:
* Cilium must be configured with the flannel configuration options (`--set global.flannel.enabled=true`) and you might also want `global.flannel.manageExistingContainers=true` as well just so that Cilium learns about existing containers, otherwise users need to restart their containers before Cilium can begin to do things like apply policy. To support Cilium removal properly there is also `global.flannel.uninstallOnExit`.
* Cilium must start to create the `04-cilium-cni.conflist` CNI configuration file which configures the chaining for both Flannel and Cilium together. This must take preference over the flannel CNI config. As long as the flannel option is enabled in helm, this will be configured correctly.
* The Cilium CNI binary and all of the flannel CNI binaries would all need to be placed in the same CNI bin dir. The Cilium addon currently reconfigures this into $SNAP_DATA rather than $SNAP. Depending on how the flannel setup works, you may need to copy the binaries over.

There might be some other things I'm not remembering right now, but if you try either approach and have some problem then feel free to reach out or ask questions.

PR #599, opened 2019-08-15
Improve build caching

joestringer:
These commits improve build caching:
* Don't depend on the microk8s tree for the containerd build, so the part can be cached
* Support specifying `KUBE_SNAP_BINS` to a non-existent directory, in which case the build will create the directory and cache k8s artifacts there. Previously a developer would first have to come up with the cached artifacts somehow, and ensure they weren't deleted, then keep them somewhere safe and specify `KUBE_SNAP_BINS`. This makes the process a bit easier for new developers.

### DIFF
--- a/build-scripts/set-env-variables.sh
+++ b/build-scripts/set-env-variables.sh
@@ -28,7 +28,9 @@ export KUBE_TRACK="${KUBE_TRACK:-}"
 
 export KUBE_VERSION="${KUBE_VERSION:-}"
 export KUBE_SNAP_BINS="${KUBE_SNAP_BINS:-}"
-if [ -z "$KUBE_SNAP_BINS" ]; then
+if [ -e "$KUBE_SNAP_BINS/version" ]; then
+  export KUBE_VERSION=`cat $KUBE_SNAP_BINS/version`
+else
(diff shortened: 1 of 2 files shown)


joestringer:
Travis failure seems unrelated to the PR:

https://travis-ci.org/ubuntu/microk8s/builds/572399036#L3252
```
tests/test-upgrade.py Testing upgrade from edge to /home/travis/build/ubuntu/microk8s/microk8s_v1.15.2_amd64.snap
error: cannot perform the following tasks:
- Download snap "microk8s" (762) from channel "edge" (received an unexpected http response code (503) when trying to download https://fastly.cdn.snapcraft.io/download-origin/fastly/EaXqgt1lyCaxKaQCU349mlodBkDCXRcg_762.snap?token=1565906400_e6f69483c35d046f4b56c80cd6fe626ba489c919)
```

joestringer:
The only question I have with this is, I'm not sure whether this prevents containerd from being updated if you make changes in the microk8s tree to something like `microk8s-resources/wrappers/run-containerd-with-args`.

ktsakalozos:
LGTM +1, thank you @joestringer

PR #603, opened 2019-08-19
Clean cilium removal

ktsakalozos:


ktsakalozos:
@joestringer this ugly patch handles the clean removal. We wait for the pods to terminate before we restart services.

PR #607, opened 2019-08-21
miscellaneous updates for strict mode

jdstrand:
This is a first pass at getting the microk8s basic services to work in strict mode. As of this time, there are several TODOs to make various snap commands work under strict mode:

* stop use of sudo (grep -r sudo ./microk8s-resources/). Eg, check for root uid and exit if non-root
* stop use of systemctl (grep -r systemctl ./microk8s-resources/, but also meta/hooks). use snapctl instead
* don't run /snap/bin/... commands. Use $SNAP/path/to/thing instead

There are likely other things that should be done. Eg, you may also want to update to not require the use of layouts.

That said, with these changes, I was able to install microk8s in strict mode, create, use and delete a pod (see https://people.canonical.com/~jamie/microk8s/README) and have it all work after reboot. Testing was performed on an Ubuntu 18.04 server install with the 2.41 snapd in edge (snap refresh core --edge; several updates were made in 2.41 to make microk8s work with containerd, etc).

* default-args/kube-apiserver: listen on localhost:2379
* default-args/etcd: listen on localhost:2379
* wrappers/apiservice-kicker: use snapctl instead of systemctl
* run-containerd-with-args: load apparmor profile directly and create dir
* run-with-config-args: create various directories used in layouts
* snapcraft.yaml: add explicit hooks, layouts and misc interfaces

### DIFF
--- a/microk8s-resources/default-args/etcd
+++ b/microk8s-resources/default-args/etcd
@@ -1,3 +1,3 @@
 --data-dir=${SNAP_COMMON}/var/run/etcd
---advertise-client-urls=unix://etcd.socket:2379
---listen-client-urls=unix://etcd.socket:2379
+--advertise-client-urls=http://localhost:2379
+--listen-client-urls=http://localhost:2379
--- a/microk8s-resources/default-args/kube-apiserver
+++ b/microk8s-resources/default-args/kube-apiserver
@@ -1,6 +1,6 @@
 --insecure-bind-address=127.0.0.1
 --cert-dir=${SNAP_DATA}/certs
---etcd-servers='unix://etcd.socket:2379'
+--etcd-servers='http://localhost:2379'
 --service-cluster-ip-range=10.152.183.0/24
 --authorization-mode=AlwaysAllow
 --basic-auth-file=${SNAP_DATA}/credentials/basic_auth.csv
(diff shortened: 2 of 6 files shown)


microk8s-resources/default-args/etcd
```diff
@@ -1,3 +1,3 @@
 --data-dir=${SNAP_COMMON}/var/run/etcd
---advertise-client-urls=unix://etcd.socket:2379
---listen-client-urls=unix://etcd.socket:2379
+--advertise-client-urls=http://localhost:2379
+--listen-client-urls=http://localhost:2379
```
jdstrand: This is needed due to limitations in the go libraries that etcd uses surrounding UNIX sockets. This is not secure, but AIUI microk8s is being updated to use cert auth.

microk8s-resources/default-args/kube-apiserver
```diff
@@ -1,6 +1,6 @@
 --insecure-bind-address=127.0.0.1
 --cert-dir=${SNAP_DATA}/certs
---etcd-servers='unix://etcd.socket:2379'
+--etcd-servers='http://localhost:2379'
```
jdstrand: Same here

ktsakalozos:
Thank you @jdstrand. I am merging this and hopefully will return to it after the 1.16 release when I get some free time.

Issue #606, opened 2019-08-20
microk8s.kubectl requires sudo


amozoss:
```
$ snap info microk8s
installed:        v1.15.3                    (778) 171MB classic
```

```
$ microk8s.kubectl 
You do not have enough permissions to access MicroK8s. Please try again with sudo.
```
I noticed in the beta that `microk8s.kubectl` requires sudo (possibility related to #590).
Is there a way to add a user to a group or something so that they don't need to type sudo constantly?

ktsakalozos:
There is not specific group but in https://github.com/ubuntu/microk8s/blob/feature/dev-docs/docs/access-witout-sudo.md you can see a few methods to work around the sudo requirement including one that lets you permit your user/group to access the critical files.

tvansteenburgh:
@ktsakalozos That link 404s.

ktsakalozos:
> @ktsakalozos That link 404s.

Oups, fixed.

dcharbonnier:
after a `sudo snap install microk8s --candidate --classic` the solution `Grant access to a user group` is not working

ktsakalozos:
@dcharbonnier could you try the edge channel (`--edge` instead of `--candidate`)?

amozoss:
After updating, I now get this message
```
$ mk get pods
Insufficient permissions to access MicroK8s.
You can either try again with sudo or add the user dan to the 'microk8s' group:

    sudo usermod -a -G microk8s dan

The new group will be available on the user's next login.
```

I was able to resolve this issue by adding myself to the microk8s group. As suggested by the error message. 
```
sudo usermod -a -G microk8s ${USER}
# open a new shell for the user, with updated group membership 
su - ${USER}
```

liguobao:
Same to me。

aleksandaratanasov:
> There is no specific group. In https://github.com/ubuntu/microk8s/blob/feature/dev-docs/docs/access-without-sudo.md you can see a few methods to work around the sudo requirement including one that lets you permit your user/group to access critical files.

Giving access to critical files should not be an option here. If actually is, then the problem lies in most cases in the tool itself.

movsb:
> After updating, I now get this message
> 
> ```
> $ microk8s.kubectl 
> Insufficient permissions to access MicroK8s.
> You can either try again with sudo or add the user dan to the 'microk8s' group:
> 
>     sudo usermod -a -G microk8s dan
> 
> The new group will be available on the user's next login.
> ```
> 
[...]

PR #608, opened 2019-08-22
Allow users to be in a group instead of using sudo

ktsakalozos:
This PR is based on your feedback @balchua (on slack), @amozoss (on https://github.com/ubuntu/microk8s/issues/606) and @alexmurray. Thank you all.

~~If the *microk8s* group is present at installation time users of that group will be granted access to `microk8s.*` commands.
If the *microk8s* group is not present at installation time, the administrator can later set a group with trusted users with: `sudo snap set microk8s user.group=<GROUP_NAME>`.~~ 
(Updated due to the strict confinement we want to ultimately reach)

The *microk8s* group is created during installation or upgrade of the snap. Users in that group have access to `microks.*` commands. To add users you normally do a `sudo usermod -a -G microk8s $USER`.

Sudo users are always allowed to access MicroK8s.

Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ sudo microk8s.kubectl config view --raw > $HOME/.kube/config
 ```
 
 #### User access without sudo
-If the *microk8s* user group exist at MicroK8s `snap install` time, users on that group
+If the *microk8s* user group exists at MicroK8s `snap install` time, users in that group
 will be granted access to `microk8s` commands.
 
 At any point you grant access to a specific group with:
--- a/README.md
+++ b/README.md
@@ -73,20 +73,13 @@ sudo microk8s.kubectl config view --raw > $HOME/.kube/config
 If the *microk8s* user group exists at MicroK8s `snap install` time, users in that group
 will be granted access to `microk8s` commands.
 
-At any point you grant access to a specific group with:
-```
-sudo snap set microk8s user.group=<GROUP_NAME>
-```
-
-For example, to create the `microk8s` group and add a user to it:
+If the *microk8s* is not available during installation you can add it later with:
 ```
 # Create the group
-sudo addgroup microk8s
-# Add a user to the group. The new group will be available on the user's next login.
-sudo usermod -a -G microk8s myuser
+sudo groupadd --system microk8s
 
 # Let MicroK8s know which user group to trust
-sudo snap set microk8s user.group=microk8s
+cd /var/snap/microk8s/current; sudo chgrp -R microk8s ./credentials ./certs ./args
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -70,16 +70,10 @@ sudo microk8s.kubectl config view --raw > $HOME/.kube/config
 ```
 
 #### User access without sudo
-If the *microk8s* user group exists at MicroK8s `snap install` time, users in that group
-will be granted access to `microk8s` commands.
-
-If the *microk8s* is not available during installation you can add it later with:
+The *microk8s* user group is created during the snap installation. Users in that group
+are granted access to `microk8s` commands. To add a user to that group:
 ```
-# Create the group
-sudo groupadd --system microk8s
-
-# Let MicroK8s know which user group to trust
-cd /var/snap/microk8s/current; sudo chgrp -R microk8s ./credentials ./certs ./args
+sudo usermod -a -G microk8s <username>"
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -59,22 +59,43 @@ snap install microk8s --classic
 MicroK8s includes a `microk8s.kubectl` command:
 
 ```
-microk8s.kubectl get nodes
-microk8s.kubectl get services
+sudo microk8s.kubectl get nodes
+sudo microk8s.kubectl get services
 ```
 
 To use MicroK8s with your existing kubectl:
 
 ```
-microk8s.kubectl config view --raw > $HOME/.kube/config
+sudo microk8s.kubectl config view --raw > $HOME/.kube/config
 ```
+
+#### User access without sudo
+If the *microk8s* user group exist at MicroK8s `snap install` time, users on that group
+will be granted access to `microk8s` commands.
+
+At any point you grant access to a specific group with:
+```
+sudo snap set microk8s user.group=<GROUP_NAME>
+```
+
+For example, to create the `microk8s` group and add a user to it:
+```
+# Create the group
+sudo addgroup microk8s
+# Add a user to the group. The new group will be available on the user's next login.
+sudo usermod -a -G microk8s myuser
+
+# Let MicroK8s know which user group to trust
+sudo snap set microk8s user.group=microk8s
+```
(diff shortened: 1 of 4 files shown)


balchua:
This is great 👍

alexmurray:
This looks great - I think this strikes a good balance between usability and security. I wonder, to make it a bit easier, whether the error which gets printed when no permission should be updated to mention this as well? ie. it could point to the docs, or it could explicitly suggest to create a group, add the local user to it and then `snap set microk8s user.group` ?

README.md
```diff
@@ -59,22 +59,43 @@ snap install microk8s --classic
...
 
 To use MicroK8s with your existing kubectl:
 
 ```
-microk8s.kubectl config view --raw > $HOME/.kube/config
+sudo microk8s.kubectl config view --raw > $HOME/.kube/config
 ```
 
+#### User access without sudo
+If the *microk8s* user group exist at MicroK8s `snap install` time, users on that group
```
tvansteenburgh: exists
users in that group

microk8s-resources/actions/common/utils.sh
```diff
@@ -3,7 +3,18 @@
 exit_if_no_permissions() {
   # test if we can access the default kubeconfig
   if [ ! -r $SNAP_DATA/credentials/client.config ]; then
-    echo "You do not have enough permissions to access MicroK8s. Please try again with sudo."
+    echo "Not enough permissions to access MicroK8s."
```
tvansteenburgh: I would change "Not enough" to "Insufficient"

README.md
```diff
@@ -59,22 +59,43 @@ snap install microk8s --classic
...
+
+At any point you grant access to a specific group with:
+```
+sudo snap set microk8s user.group=<GROUP_NAME>
+```
+
+For example, to create the `microk8s` group and add a user to it:
+```
+# Create the group
+sudo addgroup microk8s
```
jdstrand: addgroup is a Debian-ism. I suggest:
```
sudo groupadd --system microk8s
```

README.md
```diff
@@ -59,22 +59,43 @@ snap install microk8s --classic
...
+# Add a user to the group. The new group will be available on the user's next login.
```
jdstrand: Note that `newgrp` and `sg` can be used immediately to permanently change the ruid/euid.

README.md
```diff
@@ -59,22 +59,43 @@ snap install microk8s --classic
...
+sudo usermod -a -G microk8s myuser
+
+# Let MicroK8s know which user group to trust
+sudo snap set microk8s user.group=microk8s
```
jdstrand: The idea is nice, but looking forward to strict mode microk8s (ie, microk8s on Ubuntu Core), I think this is going to be problematic since a) `<GROUP NAME>` in this PR can be anything the user chooses and b) the `chgrp` calls aren't going to be allowed in strict mode. Perhaps it would be better to omit this command and be done with it (ie, you perform the getent and chgrp to microk8s down below).

In this manner you avoid introducing a config option that will be taken away with strict mode. For 'b', this becomes a documentation issue. Eg, when move to strict, you adjust the documentation to have:
```
# Finally, allow anyone in the microk8s group access to microk8s commands
$ cd /var/snap/microk8s/current ; sudo chgrp -R microk8s ./credentials ./certs ./args
```

(or we could add a new 'snap_microk8s' user to snapd that you could use with the "new in 2.41" `system-usernames` yaml).

My main point is to not introduce something that will need to be revisited.

ktsakalozos: Thank you for the insight @jdstrand.

Let me clarify what the UX would be:
- If the microk8s group is available at install time it will be used to access microk8s.* commands
- If the microk8s group is **not** available at install time the administrator would need to create the group and then configure the snap like so:
```
sudo groupadd --system microk8s
cd /var/snap/microk8s/current; sudo chgrp -R microk8s ./credentials ./certs ./args
```

Is this right?

jdstrand: This is from a security and future-strict-mode POV acceptable.

You could streamline this a bit more and perform the groupadd and the chgrp yourself unconditionally since you are a classic snap. Then the only instruction for users is the usermod command. You may want to look at what the lxd snap is doing here.

When microk8s is ready for strict mode, it could start using:
```
system-usernames:
  snap_microk8s: shared
```

(after snapd is updated for `snap_microk8s`) then snapd will create the user for you and you will be free to perform the chgrp calls within strict confinement. This way you still have just the one usermod command (for `snap_microk8s`).

ktsakalozos:
Thank you all for the feedback.

README.md
ktsakalozos: @jdstrand is there a snap exercising that approach under strict confinement? Do I need to wait for snapd to be updated? How can I test-drive this feature?

jdstrand: @ktsakalozos - sorry for getting to this just now, this somehow slipped off my radar. No other snap does this. I did talk to @pedronis about this in Vancouver, and he felt that while snapd does support `snap_daemon` internally, he doesn't want to encode things like `snap_microk8s` or `snap_lxd` into snapd itself and instead prefers store support for these types users to land. That work is currently not roadmapped. If this becomes a hard requirement for you sooner, we can revisit with @pedronis.

ktsakalozos: @jdstrand, @pedronis in the strict MicroK8s PoC the `snap_daemon: shared` is used [1]. Based on the discussion at [2] this is "Phase 1" of the larger plan. Using the `snap_daemon`  group is a bit awkward because we would ask users to be members of it in order to operate MicroK8s without sudo. @jdstrand I am not sure what you mean by "instead prefers store support for these types users to land" where can I read more about this approach?

@tvansteenburgh, this affects what we can deliver at the end of the cycle.

[1] https://github.com/ubuntu/microk8s/blob/feature/strict-v3/snapcraft.yaml#L231
[2] https://forum.snapcraft.io/t/multiple-users-and-groups-in-snaps/1461/3

jdstrand: > @jdstrand, @pedronis in the strict MicroK8s PoC the `snap_daemon: shared` is used [1]. Based on the discussion at [2] this is "Phase 1" of the larger plan. Using the `snap_daemon` group is a bit awkward because we would ask users to be members of it in order to operate MicroK8s without sudo. @jdstrand I am not sure what you mean by "instead prefers store support for these types users to land" where can I read more about this approach?
> 
> @tvansteenburgh, this affects what we can deliver at the end of the cycle.
> 
> [1] https://github.com/ubuntu/microk8s/blob/feature/strict-v3/snapcraft.yaml#L231
> [2] https://forum.snapcraft.io/t/multiple-users-and-groups-in-snaps/1461/3

Phase 1 is to support a single user for any snap to drop to, aka, 'snap_daemon'. This is implemented.

Phase 2 aims for snapd to get a list of store-approved users that the snap may use, likely stored in the snap declaration. Once implemented, it would be possible for you, as the microk8s developer, to request a new 'snap_microk8s' user via the forum, a vote to take place, and a reviewer would grant the use via store review forms.

PR #617, opened 2019-08-29
Update README.md

xnox:
Use a > "quote" syntax to get the quote looking nice, instead of fixed-width `code` syntax.


xnox:
You can see how it looks with new formatting at https://github.com/xnox/microk8s/tree/patch-1

Issue #499, opened 2019-06-08
microk8s.enable ingress does not work when context namespace is not default


asfernandes:
When I set my context namespace to a custom one and use `microk8s.enable ingress`, the ingress controller objects are created on my namespace.

But when that happens it does not work when I create my ingress. Instead of display `127.0.0.1` in the address field of `kubectl get ingress`, it shows blank and the address is not accessible.

balchua:
Oh yes. The service account `nginx-ingress-microk8s-serviceaccount` in the `clusterrolebinding` is bound to `default` namespace.  You can try the crb manifest below.

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: nginx-ingress-microk8s
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: nginx-ingress-microk8s-clusterrole
subjects:
- kind: ServiceAccount
  name: nginx-ingress-microk8s-serviceaccount
  namespace: [your custom namespace]
```

balchua:
I think fixing the ingress to be in an `ingress` namespace by default is a better approach.  Wdyt?

asfernandes:
I think `microk8s.enable ingress` should create the objects in a fixed namespace, not in the one defined in the current context.

balchua:
@asfernandes yes. I meant the addon should consider using ingress namespace. 😁

ktsakalozos:
I am +1 to using a fixed namespace for ingress (we could have that configurable with a sane default). Just to confirm we are discussing adding namespaces to the parts of the ingress manifest https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/ingress.yaml, right?

balchua:
@ktsakalozos yup. That's the one.

balchua:
We can work on this one after this pr? https://github.com/ubuntu/microk8s/pull/484

balchua:
This is fixed now.  The ingress is now in `ingress` namespace.

PR #619, opened 2019-08-31
Nginx Ingress Changes and Linkerd Upgrade

balchua:
Changes on ingress:

- Upgrade to nginx 0.25.1, some security fixes.  Including arm64.
- Remove default backend controller.
- Move to ingress namespace.

Linkerd upgrade to v2.5.0

Need help in testing arm64.

### DIFF
--- a/build-scripts/fetch-other-binaries.sh
+++ b/build-scripts/fetch-other-binaries.sh
@@ -24,5 +24,10 @@ mkdir -p $KUBE_SNAP_BINS
     curl -L https://github.com/knative/serving/releases/download/$KNATIVE_SERVING_VERSION/serving.yaml -o ./knative-yaml/serving.yaml
     curl -L https://github.com/knative/eventing/releases/download/$KNATIVE_EVENTING_VERSION/release.yaml  -o ./knative-yaml/release.yaml
     curl -L https://github.com/knative/serving/releases/download/$KNATIVE_SERVING_VERSION/monitoring.yaml  -o ./knative-yaml/monitoring.yaml
+
+    # This patch should be removed as soon as https://github.com/knative/serving/issues/5599 is fixed
+    sed -i 's@extensions/v1beta1@apps/v1@g'  ./knative-yaml/monitoring.yaml
+    sed -i '570 i \ \ selector:\n    matchLabels:\n      app: kube-state-metrics' ./knative-yaml/monitoring.yaml
+    sed -i '6985 i \ \ selector:\n    matchLabels:\n      app: node-exporter' ./knative-yaml/monitoring.yaml
   fi
 )
--- a/microk8s-resources/actions/enable.knative.sh
+++ b/microk8s-resources/actions/enable.knative.sh
@@ -4,12 +4,6 @@ set -e
 
 source $SNAP/actions/common/utils.sh
 
-echo "Knative is not yet available in v1.16 MicroK8s. Please install the v1.15 release:"
-echo ""
-echo "sudo snap install microk8s --classic --channel=1.15/stable"
-echo ""
-exit 1
-
 echo "Enabling Knative"
 
 # Knative require istio
--- a/tests/test-addons.py
+++ b/tests/test-addons.py
@@ -1,6 +1,7 @@
 import pytest
 import os
 import platform
+import time
 
 from validators import (
     validate_dns_dashboard,
@@ -22,6 +23,7 @@ from validators import (
 from utils import (
     microk8s_enable,
     wait_for_pod_state,
+    wait_for_namespace_termination,
     microk8s_disable,
     microk8s_reset
 )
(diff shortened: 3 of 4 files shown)


balchua:
Will address this issue https://github.com/ubuntu/microk8s/issues/499.

balchua:
I need to figure out why linkerd validation is not finishing.

balchua:
I couldn't figure out why tests are failing.  I've skipped linkerd validation and yet it still timing out.

balchua:
The build is timing out when the `patch-kube-proxy.sh` is restarting the daemon.
`systemctl restart snap.microk8s.daemon-proxy`

ktsakalozos:
The upgrade test we run in travis https://github.com/ubuntu/microk8s/blob/master/.travis.yml#L20
```
sudo UPGRADE_MICROK8S_FROM=edge UPGRADE_MICROK8S_TO=`pwd`/`ls microk8s*.snap` pytest -s ./tests/test-upgrade.py
```
will first snap install microk8s from edge, enable the addons run all the validators, upgrade to the local build of the snap and then run the same validators. In this way we make sure the validators run correctly before and after the upgrade. This is how we make sure we do not break deployments.

With this PR we move the ingress in its own namespaces ("ingress"). The ingress validator (https://github.com/ubuntu/microk8s/blob/master/tests/validators.py#L80) needs to be passing in both the old snap and the new one. The problem is that the validator waits for the ingress controller in the "ingress" namespaces so it is waiting forever since in the edge snap the ingress controller is on the default namespace.

balchua:
thank you!  The logs didnt print that its enabling the addons from edge. 😁. Now i got it.

balchua:
Finally got it going. Thanks for the help.

ktsakalozos:
@balchua sorry for the delay in this review, I had to finish with the clustering work.

When deleting `emojivoto` the namespace stays terminating for ever causing the linkerd validator to fail/get stuck. Any idea what may be causing this?

ktsakalozos:
When linkerd is enabled, I see this message on the apiserver logs (`journalctl -u snap.microk8s.daemon-apiserver -f`):
```
 Sep 13 08:57:24 ip-172-31-20-243 microk8s.daemon-apiserver[9008]: E0913 08:57:24.955771    9008 available_controller.go:409] v1alpha1.tap.linkerd.io failed with: failing or missing response from https://10.152.183.133:443/apis/tap.linkerd.io/v1alpha1: bad status from https://10.152.183.133:443/apis/tap.linkerd.io/v1alpha1: 400
```

balchua:
No worries for the delay.  
About the hanging issue, I didn't see on my end.  Regarding the api server logs i will have to check.  Thanks for the review.

balchua:
@ktsakalozos when those logs shows up, is it also the time when emojivoto namespace hangs on delete? Or it just shows up as long as linkerd is enabled.  Thanks

ktsakalozos:
The error in the logs shows up as soon as the addon gets enabled.

balchua:
I could not reproduce both the issues you are seeing.  But im going to clean up the `disable linkerd` to simply remove the control plane.  Apps injected with the linkerd proxy must manually remove the annotations manually.

balchua:
Im going to ask the Linkerd folks in slack about the logs emitted by the api server.

balchua:
I think those errors that you see could be transient.  Every pod inside the Linkerd control plane is also meshed with linkerd.  The tap controller will only be accessible once the linkerd  proxy sidecar is up and running.  Can you try `microk8s.linkerd check` to see if all are good?

ktsakalozos:
@balchua could you rebase? Cannot think of anything else.

balchua:
Im sorry what do you mean?

ktsakalozos:
I still see this error, but you do not. So it could be that the branch you have does not have the latest patches that went into the master. Could you update your branch with the latest patches (normally with a `git rebase`)? I can't think of anything else.

balchua:
Aaa ok got it. Thanks.

balchua:
I am able to reproduce the error you are seeing.  Its strange.  I'm trying to get help from linkerd folks.

ktsakalozos:
@balchua I suspect that for the namespaces stuck in the terminating state we should first clear resources in those namespaces (eg crds).

I am going to merge this PR https://github.com/ubuntu/microk8s/pull/646 so we get ready for v1.16 release. We might get some minor conflicts.

balchua:
@ktsakalozos, thanks for the heads up.  Emojivoto doesn't have any crd, but yes i can try that one.

balchua:
I tried linkerd 2.5.0 on 1.15 stable. It is working. The apiserver is not complaining about the tap pod.  Emojivoto is also deleted. it takes a bit of time though but it goes away.

balchua:
@ktsakalozos I think i found the reason for the errors we are seeing in the apiserver logs.  When I removed this aggregation layer argument in the kube-apiserver `--requestheader-allowed-names=front-proxy-client` it started working.  My knowledge on the aggregation layer is low, may i ask if this argument is safe to remove?

ktsakalozos:
You can remove it when linkerd is enabled and enable it in the disable script. 

Please use the skip https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/common/utils.sh#L110
and refresh https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/common/utils.sh#L69 config option arguments.

ktsakalozos:
I am +1 to using `--runtime-config` flag while we wait.

PR #625, opened 2019-09-02
Re-enable cilium when upgrading

ktsakalozos:
@joestringer the cilium addon will not upgrade as expected. To see what happens deploy microk8s from beta enable some addons and cilium and then upgrade refresh to edge:
```
sudo snap install microk8s --classic --beta
sudo microk8s.status --wait-ready
sudo microk8s.enable dns dashboard
sudo microk8s.enable cilium
# simulate an upgrade
sudo snap refresh microk8s --channel edge
```
The cilium socket seems to be missing, but I was not able to revive it in an "elegant" way.

This PR will detect if cilium is enabled and and then force a re-enabling/re-installation. Not that pretty but I believe it works.

As this blocks our release cycle (failing upgrades will stop any new releases) I would like to merge it soon and perhaps do a proper fix at a later stage based on your feedback. Many thanks.

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -302,3 +302,9 @@ then
   systemctl restart snap.${SNAP_NAME}.daemon-apiserver
 fi
 
+if [ -f "${SNAP_DATA}/bin/cilium" ]
+then
+  echo "Cilium is enabled we need to reconfigure it."
+  sudo rm -rf $SNAP_DATA/bin/cilium*
+  ${SNAP}/microk8s-enable.wrapper cilium
+fi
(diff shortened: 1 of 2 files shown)


snap/hooks/configure
```diff
@@ -302,3 +302,9 @@ then
   systemctl restart snap.${SNAP_NAME}.daemon-apiserver
 fi
 
+if [ -f "${SNAP_DATA}/bin/cilium" ]
```
joestringer: Does this need to be `-L`? That's what the disable script uses:

https://github.com/ubuntu/microk8s/blob/b4b721cbdad1018d7e2e09895864e28e71abbd67/microk8s-resources/actions/disable.cilium.sh#L9

PR #639, opened 2019-09-09
add unsecured dashboard

phcollignon:
Unsecured dashboard for development purposes.  It can be accessed unsecurely (HTTP) and without any login and has full privileges. (like minikube dashboard).   
Enable with microk8s.enable unsecured-dashboard.
Access it at http://127.0.0.1:30090

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,13 @@ MicroK8s installs a barebones upstream Kubernetes. Additional services like dns 
 sudo microk8s.enable dns dashboard
 ```
 
+
+If you are running microk8s for development, you can enable unsecured-dashboard 
+```
+sudo microk8s.enable unsecured-dashboard
+``` 
+and access it at the following url  `http://127.0.0.1:30090` 
+
 Use `microk8s.status` to see a list of enabled and available addons. You can find the addon manifests and/or scripts under `${SNAP}/actions/`, with `${SNAP}` pointing by default to `/snap/microk8s/current`.
 
 ## Documentation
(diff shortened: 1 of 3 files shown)


phcollignon:
CI error + RBAC problem ... I check this on the fork

phcollignon:
CI error is not related to my commit. Dashboard seems to work also with RBAC (was not completely installed).

ktsakalozos:
Apologies for the late review on this PR, I was caught up with the v1.16 release.

If I understand this correctly the main change is on the deployment where we do the following, right?
```
<           - --auto-generate-certificates
---
>           - --disable-settings-authorizer
```

Wouldn't it be better if we had a flag in the enable command like: `microk8s.enable dashboard:insecure`? It would be less code to maintain.

Arguments are passed to the enable script as shown in the linkerd script https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/enable.linkerd.sh#L7 

We would also need to patch the dashboard.yaml by replacing (or not) the `auto-generate-certificates`.
with `disable-settings-authorizer`. That should be just a sed command in the .enable script. What do you think?

phcollignon:
Hello, it also defers by the service account linked to it.  It has full cluster admin rights.
Yes it might be added as a flag.   
Merge it or not there is no problem for me, I just found unsecured dashboard useful in development (this is similar behaviour as in minikube)

phcollignon:
Better to create a user and let dashboard secured .. I deleted the fork.

Issue #641, opened 2019-09-11
refreshing snap from 1.13/stable to 1.15/stable fails


lazzarello:
The following sequence results in reproducible failures with the following inputs / outputs.
```
root@devstand:~# snap switch --channel=1.15/stable microk8s
"microk8s" switched to the "1.15/stable" channel

root@devstand:~# snap refresh --list
Name      Version  Rev  Publisher   Notes
microk8s  v1.15.3  825  canonical✓  classic
consider re-refresh of "microk8s"
root@devstand:~# snap refresh --classic microk8s
error: cannot perform the following tasks:
- Run configure hook of "microk8s" snap if present (run hook "configure":
-----
+ source /snap/microk8s/825/actions/common/utils.sh
+ '[' '!' -d /var/snap/microk8s/825/certs ']'
+ '[' '!' -f /var/snap/microk8s/825/certs/csr.conf.template ']'
+ cp /snap/microk8s/825/certs/csr.conf.template /var/snap/microk8s/825/certs/csr.conf.template
+ grep requestheader-client-ca-file /var/snap/microk8s/825/args/kube-apiserver
--requestheader-client-ca-file=${SNAP_DATA}/certs/ca.crt
+ grep -E -- '--(requestheader-allowed-names|requestheader-extra-headers-prefix|requestheader-group-headers|requestheader-username-headers|proxy-client-cert-file|proxy-client-key-file)=' /var/snap/microk8s/825/args/kube-apiserver
+ echo 'Enabling Enable the aggregation layer'
Enabling Enable the aggregation layer
+ echo ''
+ echo --requestheader-allowed-names=front-proxy-client
+ echo --requestheader-extra-headers-prefix=X-Remote-Extra-
+ echo --requestheader-group-headers=X-Remote-Group
+ echo --requestheader-username-headers=X-Remote-User
+ echo '--proxy-client-cert-file=${SNAP_DATA}/certs/front-proxy-client.crt'
+ echo '--proxy-client-key-file=${SNAP_DATA}/certs/front-proxy-client.key'
+ systemctl restart snap.microk8s.daemon-apiserver
+ grep -e requestheader-client-ca-file=/var/snap/microk8s/.../certs/ca.crt /var/snap/microk8s/825/args/kube-apiserver
+ mkdir -p /var/snap/microk8s/825/var/lock/
+ '[' '!' -f /var/snap/microk8s/825/credentials/client.config ']'
+ echo 'Patching client config location'
Patching client config location
[...]

PR #649, opened 2019-09-16
Fixes to allow refresh from 1.13

ktsakalozos:
Fixes https://github.com/ubuntu/microk8s/issues/641

Upgrading from 1.13 to 1.16 is failing with: 
```
+ cp -r /snap/microk8s/860/default-args/cni-network/flannel.conflist /var/snap/microk8s/860/args/cni-network/
cp: cannot create regular file '/var/snap/microk8s/860/args/cni-network/': Not a directory
-----)
```
Although this upgrade is not recommended (see https://github.com/ubuntu/microk8s/issues/641) it has to work.

Other fixes
- Use snapctl to restart services in configure hook.
- Flanneld wait for config files before starting.

### DIFF
--- a/microk8s-resources/wrappers/run-flanneld-with-args
+++ b/microk8s-resources/wrappers/run-flanneld-with-args
@@ -8,6 +8,9 @@ source $SNAP/actions/common/utils.sh
 
 exit_if_service_not_expected_to_start flanneld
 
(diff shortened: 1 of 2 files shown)


ktsakalozos:
Thanks @hyperbolic2346

PR #581, opened 2019-08-05
RBAC metrics rolebinding

giner:
Commits:
* Add system:aggregated-metrics-reader Cluster Role
* Bind proxy client user to admin role

Together with #579 fixes #560

Links:
- https://github.com/kubernetes-incubator/metrics-server/blob/v0.3.3/deploy/1.8+/aggregated-metrics-reader.yaml

PR #664, opened 2019-09-21
Fix linkerd tap pod issue

balchua:
Fixes Linkerd Tap pod issue with the `--requestheader-allowed-names`.  Use the default `CN` defined in the csr.conf which is `127.0.0.1`


ktsakalozos:
@balchua thank you for this PR. This PR does not remove the `runtime-config` workaround [1]. How can I reproduce the issue that is fixed?

[1] https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/enable.linkerd.sh#L23

balchua:
@ktsakalosoz oh sorry about that. Will fix that one.

balchua:
@ktsakalozos i was confused a bit.  The workaround that you pointed out remains by adding the `--runtime-config` option.  

What this PR is about is these errors that you previsously see in the apiserver logs.

```
E0915 08:11:07.055065   13367 available_controller.go:409] v1alpha1.tap.linkerd.io failed with: failing or missing response from https://10.152.183.90:443/apis/tap.linkerd.io/v1alpha1: bad status from https://10.152.183.90:443/apis/tap.linkerd.io/v1alpha1: 400
```

balchua:
@ktsakalozos.  Ive upgraded linkerd to 2.6.0, which fixed all the deprecated apis.  I've removed the runtime config settings in the kube-apiserver. Thanks.

balchua:
Using the `request-header-allowednames =front-proxy-client` isn't working when i install keda.  I think the certs generated by this line https://github.com/ubuntu/microk8s/blob/3727db9c0f6e2451d4a77e078b3d8e079bcdb1c9/microk8s-resources/actions/common/utils.sh#L279 doesn't add the `front-proxy-client` since `CN` is not multi value.

ktsakalozos:
@balchua Looking at [1, 2] I think we may have introduced a bug when we added the front-proxy-client certificates. 

> If both --client-ca-file and --requestheader-client-ca-file are provided, then the request first checks the --requestheader-client-ca-file CA and then the --client-ca-file. Normally, different CAs, either root CAs or intermediate CAs, are used for each of these options; regular client requests match against --client-ca-file, while aggregation requests match against --requestheader-client-ca-file. However, if both use the same CA, then client requests that normally would pass via --client-ca-file will fail, because the CA will match the CA in --requestheader-client-ca-file, but the common name CN= will not match one of the acceptable common names in --requestheader-allowed-names. This can cause your kubelets and other control plane components, as well as end-users, to be unable to authenticate to the Kubernetes apiserver.

> For this reason, use different CA certs for the --client-ca-file option - to authorize control plane components and end-users - and the --requestheader-client-ca-file option - to authorize aggregation apiserver requests.

Do I get this right @giner? Not using `127.0.0.1` seems to be the right call but we also need to have a separate CA. 

I would like us to move forward with the update of linkerd to 2.6.0 and handle the certificate issue on a second PR. Are you OK with it?

[1] https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/#enable-kubernetes-apiserver-flags
[2] https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/#ca-reusage-and-conflicts
[3] https://github.com/ubuntu/microk8s/pull/581

balchua:
@ktsakalozos im good with updating linkerd to 2.6.0, to remove the `runtime-config` option enabled when using linkerd.  
This can confuse users when they have an enabled linkerd and use a deprecated api, it will let it through.
Once they disabled linkerd, every deprecated apis will start to fail.
Let's tackle the certs in a different issue.

ktsakalozos:
> Once they disabled linkerd, every deprecated apis will start to fail.

Indeed, but I do not see any other way around this. Keeping `runtime-config` forever does not seem right since the removed APIs have been deprecated for a fair amount of time.

ktsakalozos:
Created this follow-up issue: https://github.com/ubuntu/microk8s/issues/721

ktsakalozos:
Keeping `request-header-allowednames=127.0.0.1` will probably break the metrics server fix we did in https://github.com/ubuntu/microk8s/pull/581/files#diff-cf8bbbadbd4165db8d322f64c96c84fbR236

balchua:
Oh i see.  But how does it work, when there is conflict with the certs? 😁

ktsakalozos:
Thank you for this work @balchua

giner:
> Using the `request-header-allowednames =front-proxy-client` isn't working when i install keda. I think the certs generated by this line
> 
> https://github.com/ubuntu/microk8s/blob/3727db9c0f6e2451d4a77e078b3d8e079bcdb1c9/microk8s-resources/actions/common/utils.sh#L279
> doesn't add the `front-proxy-client` since `CN` is not multi value.

front-proxy-client is not added however the default value is replaced:
```
$ openssl < /var/snap/microk8s/current/certs/front-proxy-client.crt x509 -noout -subject
subject=CN = front-proxy-client
```

giner:
> @balchua Looking at [1, 2] I think we may have introduced a bug when we added the front-proxy-client certificates.
> 
> > If both --client-ca-file and --requestheader-client-ca-file are provided, then the request first checks the --requestheader-client-ca-file CA and then the --client-ca-file. Normally, different CAs, either root CAs or intermediate CAs, are used for each of these options; regular client requests match against --client-ca-file, while aggregation requests match against --requestheader-client-ca-file. However, if both use the same CA, then client requests that normally would pass via --client-ca-file will fail, because the CA will match the CA in --requestheader-client-ca-file, but the common name CN= will not match one of the acceptable common names in --requestheader-allowed-names. This can cause your kubelets and other control plane components, as well as end-users, to be unable to authenticate to the Kubernetes apiserver.
> 
> > For this reason, use different CA certs for the --client-ca-file option - to authorize control plane components and end-users - and the --requestheader-client-ca-file option - to authorize aggregation apiserver requests.
> 
> Do I get this right @giner? Not using `127.0.0.1` seems to be the right call but we also need to have a separate CA [3].
> 
> I would like us to move forward with the update of linkerd to 2.6.0 and handle the certificate issue on a second PR. Are you OK with it?
> 
> [1] https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/#enable-kubernetes-apiserver-flags
> [2] https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/#ca-reusage-and-conflicts
> [3] #581

From the above it looks like the change is needed. Will dig a bit more on this.

giner:
@balchua @ktsakalozos We have to generate a separate CA for Aggregation Layer: https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/#ca-reusage-and-conflicts

PR #678, opened 2019-09-26
Switch snapcraft.yml to base: core

knkski:
Allows building on non-xenial hosts without having to run cleanbuild. Cuts rebuild times significantly.


ktsakalozos:
Thank you @knkski

PR #683, opened 2019-09-26
Trivial: Print user-facing errors to stderr

joestringer:
Redirect these user warnings to stderr so that they can be clearly
interpreted by users (or redirected to ignore separately from command output).

When running as user with insufficient permissions:

```
$ microk8s.status >/dev/null
Insufficient permissions to access MicroK8s.
You can either try again with sudo or add the user factorio to the 'microk8s' group:

    sudo usermod -a -G microk8s factorio

The new group will be available on the user's next login.
$
```

When running as user with sufficient permissions:

```
$ microk8s.status >/dev/null
$
```

In some Cilium developer make targets we use the success of the command to determine whether microk8s is properly set up, and it's useful to get a warning if it failed due to privileges like this.

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -3,12 +3,12 @@
 exit_if_no_permissions() {
   # test if we can access the default kubeconfig
   if [ ! -r $SNAP_DATA/credentials/client.config ]; then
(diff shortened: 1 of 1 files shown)


ktsakalozos:
LGTM +1 thank you @joestringer

Issue #686, opened 2019-09-27
install v1.16.0 fails: "can't open config file: /usr/lib/ssl/openssl.cnf"


michael-db:
Neither microk8s v1.16.0 (920) nor edge (922) installs on my Centos 7.6 1810 VM. v1.15.3 installs without error.
```
sudo yum install epel-release
sudo yum install snapd
sudo systemctl enable --now snapd.socket
sudo ln -s /var/lib/snapd/snap /snap  # enable classic snap support
```
At this point, here is how things look wrt openssl.cnf
```
ll /etc/ssl/  # output:
total 0
lrwxrwxrwx. 1 root root 16 Sep 27 12:20 certs -> ../pki/tls/certs

ls -la `sudo find /|grep openssl.cnf` # output

-rw-r--r--. 1 root root 10923 Oct 30  2018 /etc/pki/tls/openssl.cnf
lrwxrwxrwx. 1 root root    14 Sep 27 12:20 /usr/share/man/man5/openssl.cnf.5ssl.gz -> config.5ssl.gz
```
Nothing looks wrong. Following the instructions at https://snapcraft.io/docs/installing-snap-on-centos, I reboot.
Note that I've done nothing else significant up to now, I did the above steps on a new VM.
Then
```
sudo snap install microk8s --channel=1.16/stable --classic  # output:

error: cannot perform the following tasks:
- Run install hook of "microk8s" snap if present (run hook "install": 
-----
+ export LD_LIBRARY_PATH=/snap/microk8s/920/lib:/snap/microk8s/920/usr/lib:/snap/microk8s/920/lib/x86_64-linux-gnu:/snap/microk8s/920/usr/lib/x86_64-linux-gnu
+ LD_LIBRARY_PATH=/snap/microk8s/920/lib:/snap/microk8s/920/usr/lib:/snap/microk8s/920/lib/x86_64-linux-gnu:/snap/microk8s/920/usr/lib/x86_64-linux-gnu
+ export PATH=/snap/microk8s/920/usr/sbin:/snap/microk8s/920/usr/bin:/snap/microk8s/920/sbin:/snap/microk8s/920/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/usr/bin:/usr/local/bin
[...]

PR #691, opened 2019-09-30
Changing Openssl to be system independet

DomFleischmann:
In issue #686  a user wasn't able to install microk8s in Cent OS 7
due to the openssl config being located in another directory than
Ubuntu. Now it will use the openssl included in the snap making it
that way system independent.

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -269,26 +269,29 @@ get_ips() {
 }
 
 gen_server_cert() (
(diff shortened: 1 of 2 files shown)


ktsakalozos:
Thank you @DomFleischmann. This addresses the issue. Could you also update utils.sh (https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/common/utils.sh) to use openssl with a full path (as you did in the install hook)?

ktsakalozos:
LGTM +1 Thanks @DomFleischmann

Issue #669, opened 2019-09-24
requests.exceptions.SSLError: Can't connect to HTTPS URL because the SSL module is not available.


davidthewatson:
I'm following this guide:

https://discuss.kubernetes.io/t/microk8s-clustering-preview/7472

They are all running Ubuntu 19.04 with passwordless ssh enabled.

I have a master and 2 nodes. I've done:

    sudo snap install microk8s --classic --channel=1.15/edge/clustering

on all nodes. 

I tried to add node with:

    sudo microk8s.add-node

and copied/pasted the join command. When I run that, I get:

```
watson@k8s-node1:~$ sudo  microk8s.join 10.1.90.0:25000/LZbwbrVEqKTkdKcquwyfpiCCpcDWPvcu
Traceback (most recent call last):
  File "/snap/microk8s/874/usr/lib/python3/dist-packages/requests/adapters.py", line 376, in send
    timeout=timeout
  File "/snap/microk8s/874/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 549, in urlopen
    conn = self._get_conn(timeout=pool_timeout)
  File "/snap/microk8s/874/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 251, in _get_conn
    return conn or self._new_conn()
  File "/snap/microk8s/874/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 764, in _new_conn
    raise SSLError("Can't connect to HTTPS URL because the SSL "
requests.packages.urllib3.exceptions.SSLError: Can't connect to HTTPS URL because the SSL module is not available.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/snap/microk8s/874/scripts/cluster/join.py", line 369, in <module>
    info = get_connection_info(master_ip, master_port, token, callback_token)
  File "/snap/microk8s/874/scripts/cluster/join.py", line 56, in get_connection_info
    verify=False)
  File "/snap/microk8s/874/usr/lib/python3/dist-packages/requests/api.py",
line 107, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/snap/microk8s/874/usr/lib/python3/dist-packages/requests/api.py",
line 53, in request
    return session.request(method=method, url=url, **kwargs)
  File "/snap/microk8s/874/usr/lib/python3/dist-packages/requests/sessions.py", line 480, in request
    resp = self.send(prep, **send_kwargs)
[...]

PR #700, opened 2019-10-03
Force the snap included SSL to be used in join

DomFleischmann:
In newer versions of Ubuntu like 19.04 a new version of ssl
has been distributed. This caused issues with the python version
included inside the snap (see issue #669). Therefore, now
the compatible ssl included within the snap is now used.


ktsakalozos:
Closing this in favor of https://github.com/ubuntu/microk8s/pull/703

Issue #698, opened 2019-10-01
jaeger add-on does not collect traces


romainr:
More details on this thread:
https://discuss.kubernetes.io/t/jaeger-auto-injecting-into-pods/8138/5

It works out when using the official jaeger operator and the inject=true annotation: https://www.jaegertracing.io/docs/1.14/operator/#installing-the-operator-on-kubernetes

ktsakalozos:
@DomFleischmann is working on this. Thanks.

ktsakalozos:
Thank you @DomFleischmann for your work on this issue. The update should be available within a day on the edge channel (`sudo snap install microk8s --classic --edge`).

PR #711, opened 2019-10-08
Update Jaeger installation files

DomFleischmann:
The previous files used for enabling jaeger were outdated and started
failing like in issue #698. The files have now been updated and
should fix those issues.

### DIFF
--- a/microk8s-resources/actions/disable.jaeger.sh
+++ b/microk8s-resources/actions/disable.jaeger.sh
@@ -3,7 +3,30 @@
 set -e
 source $SNAP/actions/common/utils.sh
 echo "Disabling Jaeger"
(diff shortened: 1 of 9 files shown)


DomFleischmann:
> Is it better to have jaeger in `observability` namespace instead of `default`? Wdyt?

The thing is that the deployments that need jaeger need to be in the same namespace. One possible solution could be making it configurable so that we could choose the namespace where jaeger is installed. @ktsakalozos thoughts?

ktsakalozos:
@DomFleischmann making the namespace configurable makes sense.

ktsakalozos:
Thank you for this work @DomFleischmann

Issue #699, opened 2019-10-03
Helm does not show up in microk8s.status


ktsakalozos:
The helm addon does not have a corresponding pod so the `microk8s.status` script [1] need to be updated to check for the existence of the helm binary before reporting the addon enabled or disabled.

[1] https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/microk8s-status.wrapper

balchua:
@ktsakalozos i think we can close this issue now.

ktsakalozos:
Thanks @balchua

PR #718, opened 2019-10-13
add helm to microk8s.status command

greenyouse:
Hi, this is a fix for #699 so helm shows up the the `.status` command. I built a development snap using the contribution instructions and it was working as expected.

helm status added

I held off adding it to the addon array on L:34 because it wasn't possible to determine the helm install status by keying off the namespaces or clusterroles. There may be another `kubectl` command that shows the status that I'm not aware of though.

### DIFF
--- a/microk8s-resources/wrappers/microk8s-status.wrapper
+++ b/microk8s-resources/wrappers/microk8s-status.wrapper
@@ -98,6 +98,12 @@ function format_output() {
           addon["$a"]="disabled"
         fi
       done
+      if [ -f "${SNAP_DATA}/bin/helm" ]
+        then
+          addon["helm"]="enabled"
+        else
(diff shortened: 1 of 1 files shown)


balchua:
Maybe adding the `${SNAP_DATA}/bin/helm` as part of the array then an additional condition to find the file based on the value of the array.

Like this.
```
addon[helm]="${SNAP_DATA}/bin/helm"
```
Just my 2 cents.

greenyouse:
Nice, that works and keeps the code organized a bit better. I added a file check for helm in the addon for loop.

balchua:
@greenyouse great! 👍

ktsakalozos:
LGTM +1. Thank you @greenyouse for your work and @balchua for the code review.

Issue #721, opened 2019-10-14
Separete CA for the aggregation layer


ktsakalozos:
Currently we use the same CA to authorize users and the aggregation apiserver requests [2]. As described in [1] this is not a good practice.

This issue seems to be related with https://github.com/ubuntu/microk8s/pull/664 

[1] https://kubernetes.io/docs/tasks/access-kubernetes-api/configure-aggregation-layer/#ca-reusage-and-conflicts
[2] https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/default-args/kube-apiserver

giner:
Best practices:
- https://kubernetes.io/docs/setup/best-practices/certificates/#single-root-ca
- https://kubernetes.io/docs/setup/best-practices/certificates/#all-certificates
- https://kubernetes.io/docs/setup/best-practices/certificates/#certificate-paths

balchua:
@ktsakalozos this is resolved.  We can close this one now.

ktsakalozos:
Yes, thank you

PR #722, opened 2019-10-14
Separate apiserver and front proxy CAs

giner:
See https://github.com/ubuntu/microk8s/issues/721


ktsakalozos:
Great work @giner. Thank you.

Issue #720, opened 2019-10-13
Microk8s clustering (join command) error caused by missing SSL Module


CalvinHartwell:
Hi all, 

Microk8s v1.16.0 rev 921 doesn't seem to start on arm64. The main issue I am seeing is related to a missing SSL library within the snap: 

```
Oct 13 22:59:55 battleship systemd[1]: Started Service for snap application microk8s.daemon-cluster-agent.
Oct 13 22:59:57 battleship microk8s.daemon-cluster-agent[20195]: Traceback (most recent call last):
Oct 13 22:59:57 battleship microk8s.daemon-cluster-agent[20195]:   File "/snap/microk8s/921/usr/bin/gunicorn3", line 9, in <module>
Oct 13 22:59:57 battleship microk8s.daemon-cluster-agent[20195]:     load_entry_point('gunicorn==19.4.5', 'console_scripts', 'gunicorn')()
Oct 13 22:59:57 battleship microk8s.daemon-cluster-agent[20195]:   File "/snap/microk8s/921/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 542, in load_entry_point
Oct 13 22:59:57 battleship microk8s.daemon-cluster-agent[20195]:     return get_distribution(dist).load_entry_point(group, name)
Oct 13 22:59:57 battleship microk8s.daemon-cluster-agent[20195]:   File "/snap/microk8s/921/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2569, in load_entry_point
Oct 13 22:59:57 battleship microk8s.daemon-cluster-agent[20195]:     return ep.load()
Oct 13 22:59:57 battleship microk8s.daemon-cluster-agent[20195]:   File "/snap/microk8s/921/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2229, in load
Oct 13 22:59:57 battleship microk8s.daemon-cluster-agent[20195]:     return self.resolve()
Oct 13 22:59:57 battleship microk8s.daemon-cluster-agent[20195]:   File "/snap/microk8s/921/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2235, in resolve
Oct 13 22:59:57 battleship microk8s.daemon-cluster-agent[20195]:     module = __import__(self.module_name, fromlist=['__name__'], level=0)
[...]

PR #723, opened 2019-10-14
Use the right path for libs

ktsakalozos:
Fixes: https://github.com/ubuntu/microk8s/issues/720


ktsakalozos:
@balchua @DomFleischmann @CalvinHartwell this seems to be working on arm. have a look if you have time. Thanks.

Issue #738, opened 2019-10-19
microk8s.ctr doesn't list any containers since k8s 1.16


joestringer:
```
$ microk8s.ctr c ls 
CONTAINER    IMAGE    RUNTIME
```

joestringer:
This command works:

```
$ microk8s.ctr -n k8s.io containers ls
```

PR #746, opened 2019-10-21
Fix default namespace for ctr

joestringer:
Without this patch, `microk8s.ctr` is unable to list any containers on the system because it's looking in the wrong namespace. With this patch, the containers show up when running commands like `microk8s.ctr containers list`.

Fixes: #738


ktsakalozos:
LGTM +1 thank you @joestringer

PR #751, opened 2019-10-22
Added exception handling in setting addons state.

shashi278:
`addons` state can only be one of "enable" or "disable" and thus it should handle for any other value.


ktsakalozos:
LGTM +1. Thank you @shashi278

PR #791, opened 2019-11-11
Reset StorageClasses and PriorityClasses

rbt:
These non-namespaced objects were left behind after a reset.


ktsakalozos:
Thank you for the PR @rbt. Sorry for the late review.

PR #801, opened 2019-11-14
Add support for kubeflow 0.6 dashboard

knkski:
Requires an HTTPS connection, which means we have to use an ingress instead of just connecting to the pod itself.


ktsakalozos:
LGTM +1

Issue #448, opened 2019-05-01
standard K8s GPU example doesn't work


tmbdev:
Standard K8s GPU pods don't get scheduled due to a lack of nvidia.com/gpu

See here:

https://gist.github.com/6a6b5d1f4a9b84d76ad3426be54ed160

[inspection-report-20190501_111237.tar.gz](https://github.com/ubuntu/microk8s/files/3135499/inspection-report-20190501_111237.tar.gz)

ktsakalozos:
Hi @tmbdev , indeed your node does not have the "nvidia.com/gpu" capacity. I am testing it now and it should look like this:
```
                "capacity": {
                    "cpu": "4",
                    "ephemeral-storage": "235852560Ki",
                    "hugepages-1Gi": "0",
                    "hugepages-2Mi": "0",
                    "memory": "16312140Ki",
                    "nvidia.com/gpu": "1",
                    "pods": "110"
                },
```

The `"nvidia.com/gpu": "1"` is added by the `nvidia-device-plugin-daemonset` in the `kube-system` namespace. Could you share the output of `microk8s.kubectl logs -n kube-system   pod/nvidia-device-plugin-daemonset-4rcjj` and `microk8s.kubectl describe -n kube-system   pod/nvidia-device-plugin-daemonset-4rcjj`. Also what happens if you `microk8s.disable gpu` and then `microk8s.enable gpu` again? Does the problem persist?

tmbdev:
Apparently, I need to make "Nvidia" the default runtime in /etc/docker (did I miss that in the documentation?). I have done that and restarted both docker and microk8s.

But that still doesn't fix the problem. The problem now is:

2019/06/11 18:28:00 Loading NVML
2019/06/11 18:28:00 Failed to initialize NVML: could not load NVML library.
2019/06/11 18:28:00 If this is a GPU node, did you set the docker default runtime to `nvidia`?
2019/06/11 18:28:00 You can check the prerequisites at: https://github.com/NVIDIA/k8s-device-plugin#prerequisites
2019/06/11 18:28:00 You can learn how to set the runtime at: https://github.com/NVIDIA/k8s-device-plugin#quick-start

The library seems to be installed:

$ sudo ldconfig -p | grep nvidia-ml
        libnvidia-ml.so.1 (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libnvidia-ml.so.1
[...]

PR #813, opened 2019-11-19
Revert to containerd 1.2.5 until we get gpu to work on 1.3.0

ktsakalozos:


ktsakalozos:
I am sorry @balchua I have to revert the 1.3.0 transition. We accidentally broke the gpu support, see https://github.com/ubuntu/microk8s/issues/448 .

balchua:
Oh shoot!! Im sorry.

ktsakalozos:
> Oh shoot!! Im sorry.

No worries, I should have tested it better

balchua:
Not too familiar with gpis, I was checking about this a little bit, according to this issue https://github.com/NVIDIA/k8s-device-plugin/issues/107
There is a need to change the runtime to use nvidia-container-runtime.

ktsakalozos:
> Maybe in the new containerd, we dont have to change the runtime.

Yes, it might be something as simple as that. Unfortunately at the moment I do not have enough time to dig into that.

PR #826, opened 2019-11-26
Remove timeout when enabling Kubeflow

### LINKED ISSUE #817: Unable to login into KubeFlow after the addon is enabled
[inspection-report-20191121_130543.tar.gz](https://github.com/ubuntu/microk8s/files/3876068/inspection-report-20191121_130543.tar.gz)

Installed microk8s on a 32-core/64GB workstation running 18.04LTS latest. The snap install and addon enablement went fine:

snap install microk8s --classic
microk8s.enable registry dns rbac storage knative kubeflow istio
...all worked...

KubeFlow is at 0.6.0-rc0

At the end of install, I see that KubeFlow is up and running (all objects in *kubeflow* namespace are RUNNING).

However, I have no way to login. The best I could come up with is:

$ microk8s.kubectl port-forward svc/ambassador -n kubeflow 8080:80

Which does seem to trigger Ambassador and redirect me to the https://localhost:8080/kflogin page. But at that point the SSL connection fails with either:

SSL_ERROR_RX_RECORD_TOO_LONG  (Firefox)
ERR_SSL_PROTOCOL_ERROR (Chrome)

I assume a self-signed certificate is installed via the microk8s addon process.

Again, the dashboard is definitely up since I can port-forward to it directly but then all links off of that page are 404 (I suspects the redirects are screwed up because of the direct connection to 8082). Help!

knkski:
A one-size-fits-all timeout won't work for everyone, so just remove it. Also print out login username/password on successful enablement.

Fixes #817

### DIFF
--- a/microk8s-resources/actions/enable.kubeflow.sh
+++ b/microk8s-resources/actions/enable.kubeflow.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 
-import argparse
 import json
 import os
 import random
@@ -10,6 +9,7 @@ import sys
 import tempfile
 import textwrap
 import time
+from itertools import count
 
 
 def run(*args, die=True):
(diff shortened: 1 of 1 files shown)


ktsakalozos:
LGTM +1 @knkski

Issue #822, opened 2019-11-23
Microk8s on Fedora 31


soumplis:
It seems that Fedora 31 drops support (by default) for Cgroups v1 [1] and microk8s.status returns always bad status. After enabling v1 all seems fine.

[1] https://www.redhat.com/sysadmin/fedora-31-control-group-v2

balchua:
Related to this https://github.com/ubuntu/microk8s/issues/651
Im not sure if containerd supports cgroups v2.  I know its being worked on perhaps in containerd 1.4.

ktsakalozos:
@soumplis how did you enable v1 cgroups? It would be great if we could detect this setup and suggest a fix in the `microk8s.inspect` [1] command.

[1] https://github.com/ubuntu/microk8s/blob/master/scripts/inspect.sh#L104

omkensey:
Here are more details on Fedora 31, containers, and cgroups, including how to switch back to CGroups V1 (requires a reboot though): https://docs.fedoraproject.org/en-US/fedora/f31/release-notes/developers/Containers/

PR #829, opened 2019-11-27
cgroups v1 check

soumplis:
Extra checks for Fedora as since Fedora 31 Cgroups v1 support is not enabled and Docker does not support Cgroups v2 yet 

Relates to: #822


ktsakalozos:
Thank you @soumplis LGTM +1

PR #839, opened 2019-11-30
Use python3 to run tests

balchua:
Move the conditional checks to `@pytest.mark.skipif` to avoid duplicating codes to skip certain tests.


balchua:
Moving to python3 to run tests. plus some few changes to how we conditionally disable test.

ktsakalozos:
I like this PR but it is mixed with the work on containerd 1.3.

balchua:
I will close this one and send a new pr without the containerd changes.

PR #848, opened 2019-12-06
use python3 for tests

balchua:
Usage of `pytest.mark` markers to disable or enable tests.  Upgrade to python3.


balchua:
snap store is inaccessible, thats why the build failed.  How to retrigger the build?

PR #872, opened 2019-12-19
Switch to stable Juju 2.7 by default for addon

knkski:
Also allow overriding the version used with an environment variable.


knkski:
Sorry about that, looks like I dropped a `.0` at the end of the version string, should be fixed now.

Issue #841, opened 2019-12-01
Is microk8s supporting kubectl plugins?


yuxx0248:
Please run `microk8s.inspect` and attach the generated tarball to this issue.
```
@working-computer:~$ microk8s.inspect
Inspecting services
  Service snap.microk8s.daemon-cluster-agent is running
 FAIL:  Service snap.microk8s.daemon-flanneld is not running
For more details look at: sudo journalctl -u snap.microk8s.daemon-flanneld
  Service snap.microk8s.daemon-containerd is running
  Service snap.microk8s.daemon-apiserver is running
  Service snap.microk8s.daemon-apiserver-kicker is running
  Service snap.microk8s.daemon-proxy is running
  Service snap.microk8s.daemon-kubelet is running
  Service snap.microk8s.daemon-scheduler is running
  Service snap.microk8s.daemon-controller-manager is running
  Service snap.microk8s.daemon-etcd is running
  Copy service arguments to the final report tarball
Inspecting AppArmor configuration
Gathering system information
  Copy processes list to the final report tarball
  Copy snap list to the final report tarball
  Copy VM name (or none) to the final report tarball
  Copy disk usage information to the final report tarball
  Copy memory usage information to the final report tarball
  Copy server uptime to the final report tarball
  Copy current linux distribution to the final report tarball
  Copy openSSL information to the final report tarball
  Copy network configuration to the final report tarball
Inspecting kubernetes cluster
  Inspect kubernetes cluster

Building the report tarball
  Report tarball is at /var/snap/microk8s/1073/inspection-report-20191130_233811.tar.gz

```

Is microk8s currently supporting plugins? Or I just mess up some setups...
I installed some popular plugins and kubectl was able to find them with plugin list
```
working-computer:~$ sudo microk8s.kubectl plugin list
The following compatible plugins are available:

/usr/local/bin/kubectl-ctx
/usr/local/bin/kubectl-krew
/usr/local/bin/kubectl-ns

```

But the ctl cannot find it when I try to use them
```
working-computer:~$ sudo microk8s.kubectl krew
Error: unknown command "krew" for "kubectl"
Run 'kubectl --help' for usage.
```

At this moment, I am not sure whether this is a microk8s issue or kubectl issue, but since I am able to use these plugins on my mac with minikube, I am opening the issue here first. 

Thank you for your help!

balchua:
@yuxx0248 can you try running `microk8s.kubectl` without sudo?

[...]

PR #874, opened 2019-12-20
allow microk8s.kubectl to use plugins.

balchua:
This PR removes the need to use the `kubeconfig` argument when using `microk8s.kubectl`, instead it relies on the env variable `KUBECONFIG` set in the wrapper script.
This will fix this issue https://github.com/ubuntu/microk8s/issues/841


microk8s-resources/wrappers/microk8s-kubectl.wrapper
```diff
@@ -17,6 +17,5 @@ if echo "$*" | grep -v -q -- '--kubeconfig'; then
   exit_if_no_permissions
 fi
 
-
-declare -a args="($(cat $SNAP_DATA/args/kubectl))"
-"${SNAP}/kubectl" "${args[@]}" "$@"
+export KUBECONFIG=$SNAP_DATA/credentials/client.config
```
ktsakalozos: I am not sure about this. In the past we had users complaining that they could not change the location of the kubeconfig file so we added this arguments file.

What if we had a kubectl-env file with `export KUBECONFIG=$SNAP_DATA/credentials/client.config` and we sourced that file instead of hardcoding the client.config location here? We could leave the args/kubectl file empty incase someone needs it. What do you think?

balchua: Oh yes. I didn't consider that users can change the content of the config.
Sounds fair.

balchua:
Needs review.

ktsakalozos:
@balchua what do you think about the last commit? The intention is to keep the current setup (kubectl with whatever arguments already set plus no kubectl-env file) when the kubeconfig file pointed in kubectl is not the default one. If the `--kubeconfig` argument is the default one we remove it and use the kubectl-env to set the KUBECONFIG env variable.

balchua:
Lgtm

balchua:
@ktsakalozos Im thinking of closing this one. It is currently being tracked in sig-cli upstream. https://github.com/kubernetes/kubectl/issues/727.
Wdyt?

ktsakalozos:
@balchua quickly reading the commants in the proposed fix in https://github.com/kubernetes/kubernetes/pull/85796 makes me doubt if there is going a fix. Lets move on with this PR. I just did not had time to do a proper testing.

Issue #877, opened 2019-12-22
microk8s.reset hangs when addons are enabled


balchua:
Sometimes when addons using CRDs or api-services, resetting the cluster hangs during namespace deletion.
Reset also does not properly cleanup addon other resources such as service accounts, clusterroles, etc..

mayankshah1607:
Hi, I am not really sure if its the addons that's causing this problem. I tried disabling the addons and running `microk8s.reset` then. The command hangs after the namespace deletion. Any suggestions?

balchua:
This usually happens when crds are still being used.  If somewhere you have crds used in a namespace, you must delete it first prior to deleting the namespace.

balchua:
@mayankshah1607 thanks for the info.  I was thinking about this further, perhaps should move the `crd` deletion earlier.

https://github.com/ubuntu/microk8s/blob/dbc90b6332ee0d96d94ebeaab044b2815adfe01f/microk8s-resources/wrappers/microk8s-reset.wrapper#L54

Will try this one out. Thanks 👍

ukd1:
I'm still seeing this as well, even after trying @balchua's suggestion.

balchua:
@ukd1 `apiservices` also needs to be deleted before the namespace.  Otherwise it will be stuck in `Terminating` state.

`kubectl delete apiservices --all`. IIRC.

PR #878, opened 2019-12-22
disable all addons upon reset to have clean uninstall and doesnt hang.

balchua:
This PR will first disable all addons to cleanly remove addon resources created.
Prevents namespace stuck in terminating phase, when addons are cleanly disabled and removed.


balchua:
address this issue: https://github.com/ubuntu/microk8s/issues/877

microk8s-resources/wrappers/microk8s-reset.wrapper
```diff
@@ -17,6 +17,15 @@ Options:
 EOF
 }
 
+disable_addons() {
+  echo "Disabling all addons."
+  actions="$(get_all_addons)"
+  for action in $actions; do
+    echo "Disabling addon : $action"
+    microk8s.disable $action > /dev/null 2>&1 || true
```
ktsakalozos: After enabling storage its disable process gets stuck in:
```
Remove PVC storage at /var/snap/microk8s/common/default-storage ? (Y/N)
```

Having a non interactive enable/disable mode for all addons would have made our life easier.

balchua: I will work on this one. Thanks for reviewing.

balchua: @ktsakalozos will it be fine to force an argument to `microk8s.disable storage:destroy-storage`  without specifying the argument will assume to keep the storage. wdyt?
thanks.

ktsakalozos: Sorry I missed that @balchua. Yes this is a reasonable approach.

balchua:
I have to update `pip3` in the `travis.yaml` file `pip3 install --upgrade pip`, otherwise it kept failing when installing `pytest`

balchua:
Ive modified `microk8s.disable storage` to include `destroy-storage` args.

If arg is not specified, it will not delete the filesystem.

`microk8s.disable storage:destroy-storage`

microk8s-resources/wrappers/microk8s-reset.wrapper
ktsakalozos: @balchua I was wrong to suggest we could change the UI of `microk8s.disable storage` mid-cycle. On the last commit you can see I maintained the old behavior of asking for confirmation in deleteing the storage *if* no arguments were passed.

ktsakalozos:
LGTM +1 thank you @balchua

PR #901, opened 2020-01-08
Reënable Kubeflow tests in Jenkins

knkski:
With Kubeflow stable, we can enable testing of microk8s.enable kubeflow


knkski:
Superceded by #848

PR #914, opened 2020-01-15
Add Juju disable script

knkski:
Otherwise, if someone attempts to change the Juju version, they'll be stuck with the old version unexpectedly.

### DIFF
--- a/microk8s-resources/actions/disable.juju.sh
+++ b/microk8s-resources/actions/disable.juju.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -eu
+
+source $SNAP/actions/common/utils.sh
+
+function disable_juju() {
+  echo "Removing Juju..."
+  if [ -f "${SNAP_DATA}/bin/juju" ]; then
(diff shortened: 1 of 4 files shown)


ktsakalozos:
I am not sure about this PR @knkski. What problem do we solve? Why don't we implement the `microk8s.disable juju` command so as to allow users to remove and install juju at will?

ktsakalozos:
> attempts to change the Juju version, they'll be stuck with the old version unexpectedly

The user should not expect to change the JUJU_VERSION and have microk8s follow. In the same way we ship specific/blessed versions of all other components. Are you sure your kubeflow bundle will work with juju 2.9.0?

ktsakalozos:
@knkski what if we implement the `microk8s.disable juju` command and when we do a `microk8s.enable kubeflow` we call the disabling of juju?

knkski:
@ktsakalozos, @johnsca: Thanks, updated the PR to disable Juju when Kubeflow gets disabled.

ktsakalozos:
@johnsca thank you for your input on this. Keeping juju uptodate is going to be a tricky and we need to find a way to do this properly.

PR #876, opened 2019-12-21
Make `microk8s.status` addons more readable

balchua:
With the list of addons growing, it is becoming difficult to show which ones are enabled or disabled.
This PR now groups the enabled and disabled addons (added some colors).
Example:
Non yaml output.
```
microk8s.status 
microk8s is running
addons:
enabled:
  dns
  rbac

disabled:
  cilium
  dashboard
  fluentd
  gpu
  helm
  ingress
  istio
  jaeger
  juju
  knative
  kubeflow
  linkerd
  metallb
  metrics-server
  prometheus
  registry
  storage
```

Yaml output:
```
microk8s:
  running: true
addons:
  enabled:
  - rbac
  - dns
  disabled:
  - knative
  - jaeger
  - juju
  - fluentd
  - gpu
  - cilium
  - storage
  - helm
  - registry
  - metallb
  - ingress
  - metrics-server
  - linkerd
  - prometheus
  - istio
  - dashboard
  - kubeflow
```

PR #927, opened 2020-01-23
Don't re-enable addons if already enabled

grebennikov:
Don't re-enable addons if already enabled


johnsca:
#876 also [includes this check](https://github.com/ubuntu/microk8s/pull/876/files#diff-5360b0716ef17c99facbddaf9fd3a184R31-R36), along with some changes to make the check more efficient.

ktsakalozos:
I will merge this for now because the https://github.com/ubuntu/microk8s/pull/876 needs some refactoring either way.

PR #928, opened 2020-01-23
Preserve stopped.lock across refresh

johnsca:
Also adds improves messages from status, start, and stop commands when run when stopped.


ktsakalozos:
@johnsca closing this PR as the issue of restarting daemons after a refresh is addressed by snapd.

PR #929, opened 2020-01-23
Do not restart the apiserver on configure

ktsakalozos:
Does not restart the apiserver upon configure.


ktsakalozos:
@johnsca this is why the api-server restarts upon configure. Unfortunately this behavior was hidden by the snapd restarting all services upon refresh.

PR #940, opened 2020-01-29
Set up magic IP for API services to bind to

### LINKED ISSUE #351: Auto-configure Magic IP 10.0.2.2 to talk to host OS
Considering Microk8s a solution to run locally, it lacks a very useful feature to allow Pods connect to Host services smoothly: Magic IP 10.0.2.2 (like minikube). 

I was presenting at Scale17x and my Pod Configmap connected to wlan0 interface to connect my local Elasticsearch service, for some reason my IP address changed and could not reach Elasticsearch anymore, this could be solved if 10.0.2.2 existed and always routed to the host OS.

knkski:
Sets up a user-configurable IP that is stable between network changes. Defaults to 10.5.2.1, and allows configuring a different IP in case of conflicts.

Also adds microk8s.local to /etc/hosts, since K8s Ingresses don't allow straight IP addresses.

Fixes #351

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -271,15 +271,7 @@ wait_for_service_shutdown() {
 }
 
 get_default_ip() {
-    # Get the IP of the default interface
-    local DEFAULT_INTERFACE="$($SNAP/bin/netstat -rn | $SNAP/bin/grep '^0.0.0.0' | $SNAP/usr/bin/gawk '{print $NF}' | head -1)"
-    local IP_ADDR="$($SNAP/sbin/ip -o -4 addr list "$DEFAULT_INTERFACE" | $SNAP/usr/bin/gawk '{print $4}' | $SNAP/usr/bin/cut -d/ -f1 | head -1)"
-    if [[ -z "$IP_ADDR" ]]
-    then
-        echo "none"
-    else
-        echo "${IP_ADDR}"
-    fi
+    echo "$(snapctl get api.ip)"
 }
 
 get_ips() {
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -4,6 +4,7 @@ set -eux
 
 source $SNAP/actions/common/utils.sh
 
+
 # This is a one-off patch. It will allow us to refresh the beta snap without breaking the user's deployment.
 # We make sure the certificates used by the deployment from beta do not change. We copy them to SNAP_DATA
 # and make sure the respective services use them.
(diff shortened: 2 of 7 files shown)


ktsakalozos:
@knkski I do not fully understand this PR. Why do we change all services to talk to this magic IP? We want only the apiserver to listen to that magic IP. To do this just create the interface, assign an IP and restart the apiserver. The apiserver listens on all available interfaces and that would include the new interface. Also could we have this functionality as an addon as this is a feature that normally you do not have it.

microk8s-resources/actions/common/utils.sh
```diff
@@ -271,15 +271,7 @@ wait_for_service_shutdown() {
 }
 
 get_default_ip() {
```
ktsakalozos: This is used in a couple of different places and cannot be changed so easily. For example `microk8s.config` will not be returning the kubeconfig to access microk8s in a VM.

snap/hooks/configure
```diff
@@ -371,4 +372,39 @@ then
...
+
+    $SNAP/bin/sed -i "s/.*microk8s.local/${API_IP}\tmicrok8s.local/g" /etc/hosts
+
+    snapctl restart ${SNAP_NAME}
+  fi
+else
+  $SNAP/sbin/ip link add uk8s0 type bridge
+  $SNAP/sbin/ip address add $API_IP/24 dev uk8s0
+  $SNAP/sbin/ip link set uk8s0 up
+  $SNAP/bin/echo "${API_IP}\tmicrok8s.local" >> /etc/hosts
```
ktsakalozos: I suspect we cannot do this since we are trying to go to reach strict confinement. @joedborg what is the right way to alter `/etc/hosts`.

snap/hooks/configure
```diff
@@ -371,4 +372,39 @@ then
...
+
+  if [[ "$API_IP" != "$CURRENT_IP" ]]; then
+    $SNAP/sbin/ip link delete uk8s0
+    $SNAP/sbin/ip link add uk8s0 type bridge
+    $SNAP/sbin/ip address add $API_IP/24 dev uk8s0
+    $SNAP/sbin/ip link set uk8s0 up
+
+    for conf in client controller kubelet proxy scheduler; do
+      $SNAP/bin/sed -i 's/IP/'"${API_IP}"'/g' ${SNAP_DATA}/credentials/${conf}.config
+      $SNAP/bin/sed -i 's|server: .*|server: https://'"${API_IP}"':16443|g' ${SNAP_DATA}/credentials/${conf}.config
```
ktsakalozos: This does not look right in the case of a cluster node. Lets say this microk8s instance is a node part of a cluster and the apiserver is a remote server. We will be telling kubelet to contact the local apiserver that is not running.

snap/hooks/remove
```diff
@@ -34,7 +34,11 @@ rm -rf ${SNAP_COMMON}/run/containerd/* || true
...
-  $SNAP/sbin/ip link delete cni0
-fi
+for link in cni0 uk8s0; do
+  if $SNAP/sbin/ip link show $link
+  then
+    $SNAP/sbin/ip link delete $link
+  fi
+done
+
+$SNAP/bin/sed -i '/microk8s.local/d' /etc/hosts
```
ktsakalozos: Same here, I do not think this is the right way to change the `/etc/hosts`

snap/hooks/configure
```diff
@@ -371,4 +372,39 @@ then
...
         chgrp microk8s -R "$SNAP_DATA/juju" "$SNAP_DATA/juju-home" || true
     fi
-fi
\ No newline at end of file
+fi
+
+# Set up a bridge with a known IP address that services can bind to instead of
+# whatever IP gets handed out by their router. Doesn't affect services that bind
+# to 127.0.0.1.
+API_IP="$(snapctl get api.ip)"
```
ktsakalozos: Up until now we have not used the snapctl key-value store. The reason is that there are so many arguments that the user can change that we cannot have corresponding entries in the snapctl key value store. What we do instead is to have the directory $SNAP_DATA/args with files the user can change. Sure it is not a great experience but I am pretty sure the `snapctl get/set` experience will not be that great either.

microk8s-resources/actions/common/utils.sh
knkski: @ktsakalozos: It looks like with this PR, there's two opposing use cases, making microk8s available from outside a VM, and making microk8s available at a magic IP address (which wouldn't be available from outside a VM). Should we introduce a flag such as `--magic/--no-magic` to `microk8s.config` that would disambiguate between these two use cases?

snap/hooks/configure
knkski: @ktsakalozos: How do we figure out if microk8s is clustered, and how do we figure out what the correct IP is?

microk8s-resources/actions/common/utils.sh
ktsakalozos: `microk8s.config` will give you a kubeconfig. If you are sitting in front of the host the kubeconfig would better have the 127.0.0.1 endpoint for accessing the apiserver. If MicroK8s is in a VM and/or you are not on the host you should be using the endpoint though which you can access the apiserver and this is probably the default IP and not the magic IP. So I am not sure when you would want a kubeconfig with the magic IP on.

The usecase I see for the magic IP is when you have a service on the host and you need to access it from a pod. I this case the default IP is not good because it may change and you may not want to make that service publicely available.

snap/hooks/configure
ktsakalozos: If a node act like a cluster it has the ${SNAP_DATA}/var/lock/clustered.lock file, see https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/run-etcd-with-args#L10

ktsakalozos:
I am closing this @knkski since you ended up using the metallb loadbalancer.

PR #959, opened 2020-02-11
Windows and MacOS binaries

joedborg:
Leverages AppVeyor build platform to build static binaries for Windows and MacOS.


joedborg:
@ktsakalozos Tested and working after code tidy.

installer/vm_providers/_snap.py
```diff
@@ -0,0 +1,329 @@
+# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
```
ktsakalozos: Is it possible we do not need this file?

joedborg: I'm pretty sure this is what's doing the `snap install microk8s --classic`, but I can make a test build

PR #964, opened 2020-02-12
Add `--assume-yes` flag to `micok8s install`

joedborg:
This is make testing easier for us and scripted installs easier for everyone else.  Tested in all expected conditions.

### DIFF
--- a/installer/cli/microk8s.py
+++ b/installer/cli/microk8s.py
@@ -104,7 +104,7 @@ def install(args) -> None:
     parser.add_argument('--cpu', default=definitions.DEFAULT_CORES, type=int)
     parser.add_argument('--mem', default=definitions.DEFAULT_MEMORY, type=int)
     parser.add_argument('--disk', default=definitions.DEFAULT_DISK, type=int)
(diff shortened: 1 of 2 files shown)


joedborg:
Merging into installer branch as no risk to master.

installer/cli/microk8s.py
```diff
@@ -104,6 +104,7 @@ def install(args) -> None:
     parser.add_argument('--cpu', default=definitions.DEFAULT_CORES, type=int)
     parser.add_argument('--mem', default=definitions.DEFAULT_MEMORY, type=int)
     parser.add_argument('--disk', default=definitions.DEFAULT_DISK, type=int)
+    parser.add_argument('--assume-yes', action='store_true', default=definitions.DEFAULT_ASSUME)
```
tvansteenburgh: Two comments:
1. This new arg should be added in `_show_install_help()` also.
2. There really should be a `-y` shortcut for this also, since `-y` is kind of the universally used shorthand for "run this command and automatically say 'Yes' anywhere you would have prompted for my opinion."

PR #966, opened 2020-02-14
Add `--assume-yes` to doc and fix PyInstaller issue on Linux

joedborg:
Put the `-y` flag on `--assume-yes` and refect this in the docstring.
Format pyinstaller spec file and add hidden import to fix https://github.com/pypa/setuptools/issues/1963


ktsakalozos:
LGTM +1 @joedborg

The failing tests is because of MacOS

Issue #945, opened 2020-02-03
Certificates not trusted on macOS Catalina


VonRehberg:
With latest macOS new requirements for trusted certificates got introduced (see https://support.apple.com/en-us/HT210176 )

One requirement is, that certificates should not be valid for more than 825 days, which collides with the -days 10000 option that is currently in place.

Are there plans to adopt to these changes?
Is there any possibility for a quickfix? Replacing all certs after a restart seems cumbersome.

balchua:
@VonRehberg thank you for reporting this.  Unfortunately as you can see it is written as part of the command when generating the certs.

https://github.com/ubuntu/microk8s/blob/69ea17d4c7a135573d24cfb320edff394b6b0f6d/microk8s-resources/actions/common/utils.sh#L327 and so on.

@ktsakalozos is it possible to add a non-standard attribute to the `csr.conf.template` like `cert_days`, we can grep the value?
Why i mention to add it to the `csr.conf.template` so that the api service kicker will pick up the changes.  I just dont know if openssl will complain about unknown properties.

VonRehberg:
Is there a specifiy reason for not choosing a more reasonable period like 2 years?

ktsakalozos:
@VonRehberg how does this issue manifests itself? What do you see failing? I am asking just for a way to reproduce the issue so we are all on the same page.

The reason we did not set on a more reasonable time period is that 
a) we were not aware of the 825 day constraint and 
b) we did not have in our roadmap to work on refreshing the certificates.

I believe we should comply with the 825 day requirement so we provide a good out of the box user experience. As a follow up we should work on assisting users in all aspects of certificate management eg issue, cycle, update, integrate with certificate management systems etc. @tvansteenburgh this would a good roadmap item.

@VonRehberg would you be interesting in updating the certificate creation in [1] to address you needs and provide some feedback if this fixes the problem? In [2] you can read how to build MicroK8s.

Thank you

[1] https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/common/utils.sh
[2] https://github.com/ubuntu/microk8s/blob/master/docs/build.md

VonRehberg:
![image](https://user-images.githubusercontent.com/8611608/74147858-65387380-4c04-11ea-9518-295c90ca4e2e.png)
As you can see, the certificate is blocked on OS level and I can't work around it using Chrome or Safari. I need to manually add the public key to the keychain and adjust the trust level.

[...]

PR #970, opened 2020-02-18
Reduce certificate expiry time

joedborg:
Trying to address https://github.com/ubuntu/microk8s/issues/945


joedborg:
This is still blocked by both Chrome and Safari (although I believe this is a bug in Safari as it lets you add the exception and later acknowledges this), but means that the dates on certificates are now valid.

microk8s-resources/actions/common/utils.sh
```diff
@@ -366,6 +375,26 @@ produce_certs() {
     fi
 }
 
+render_ca_conf() {
+    # Render ca.conf.template to csr.conf.rendered
+
+    local IP_ADDRESSES="$(get_ips)"
+
+    cp ${SNAP_DATA}/certs/ca.conf.template ${SNAP_DATA}/certs/ca.conf.rendered
+    if ! [ "$IP_ADDRESSES" == "127.0.0.1" ] && ! [ "$IP_ADDRESSES" == "none" ]
```
ktsakalozos: This part does not look right. 

The `ca.conf.template` does not have a MOREIPS line. I think we do not need this part.

microk8s-resources/actions/common/utils.sh
```diff
@@ -323,26 +323,35 @@ produce_certs() {
         fi
     done
 
+    render_ca_conf
+    if ! [ -f "${SNAP_DATA}/certs/ca.conf" ]; then
+        echo "changeme" >  "${SNAP_DATA}/certs/ca.conf"
+    fi
+
+    if ! "${SNAP}/usr/bin/cmp" -s "${SNAP_DATA}/certs/ca.conf.rendered" "${SNAP_DATA}/certs/ca.conf"; then
+        cp ${SNAP_DATA}/certs/ca.conf.rendered ${SNAP_DATA}/certs/ca.conf
```
ktsakalozos: This does not look right. I think that if there is a change in the ca we should set the force flag on this function so all related certificates are recreated.

microk8s-resources/actions/common/utils.sh
```diff
@@ -323,26 +323,35 @@ produce_certs() {
...
+    fi
+
     # Generate apiserver CA
     if ! [ -f ${SNAP_DATA}/certs/ca.crt ]; then
-        ${SNAP}/usr/bin/openssl req -x509 -new -nodes -key ${SNAP_DATA}/certs/ca.key -subj "/CN=10.152.183.1" -days 10000 -out ${SNAP_DATA}/certs/ca.crt
+        ${SNAP}/usr/bin/openssl req -x509 -new -sha256 -nodes -key ${SNAP_DATA}/certs/ca.key -subj "/CN=10.152.183.1" -out ${SNAP_DATA}/certs/ca.crt -config ${SNAP_DATA}/certs/ca.conf
```
ktsakalozos: This creates the certificate only the first time. Should we follow the pattern we have for the rest of the certificates where is someone changes the template we trigger the re-issue of the certs?

microk8s-resources/actions/dashboard.yaml
```diff
@@ -167,18 +167,14 @@ spec:
     spec:
       containers:
         - name: kubernetes-dashboard
-          image: kubernetesui/dashboard:v2.0.0-beta5
+          image: kubernetesui/dashboard:v2.0.0-rc5
```
ktsakalozos: We may want to have this as a separete PR so it does not get blocked by the rest of the work we are doing here.

balchua:
Safari is reducing the certificate validity to 1 year.  Im not a mac user, but just wonderin if that will have an impact on this PR.

ktsakalozos:
> Safari is reducing the certificate validity to 1 year

@balchua, @joedborg, I see bot of you agree on a 365 certificate validity so lets go with that (reverted the last commit).

PR #971, opened 2020-02-20
Fix core not installing on arm

ktsakalozos:


ktsakalozos:
Not happy with it, but it unblocks the builds.

Issue #975, opened 2020-02-24
microk8s.enable juju fails on curl step on Centos7


dliebsterwb:
Hello, 

'microk8s.enable juju' fails to pull juju-2.7.0-centos7.tar.gz from typepad.net with curl error 77. Centos 7's native curl pulls the file ok, there seems to be an issue using the snap's version of curl:

`#  /var/lib/snapd/snap/microk8s/1176/usr/bin/curl -L https://launchpad.net/juju/2.7/2.7.0/+download/juju-2.7.0-centos7.tar.gz -o /root/1176/tmp/juju.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (77) Problem with the SSL CA cert (path? access rights?)

[root@centos7 ~]# /bin/curl -L https://launchpad.net/juju/2.7/2.7.0/+download/juju-2.7.0-centos7.tar.gz -o /root/1176/tmp/juju.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  116M  100  116M    0     0  1239k      0  0:01:36  0:01:36 --:--:--  943k
[root@centos7 ~]# uname -a
Linux centos7.localdomain 3.10.0-1062.4.3.el7.x86_64 #1 SMP Wed Nov 13 23:58:53 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
[root@centos7 ~]# cat /etc/redhat-release 
CentOS Linux release 7.7.1908 (Core)
`

So far I've yum installed ca-certificates based on another post, microk8s.enable juju still still with the came curl error.   Would there be any unexpected microk8s management issues if the enable.juju.sh script is run manually outside of microk8s.enable? 

Thanks!
-Dan
[inspection-report-20200224_171509.tar.gz](https://github.com/ubuntu/microk8s/files/4245845/inspection-report-20200224_171509.tar.gz)
Please run `microk8s.inspect` and attach the generated tarball to this issue.

We appreciate your feedback. Thank you for using microk8s.

knkski:
[...]

PR #989, opened 2020-02-27
Including juju binary in snap instead of downloading it

knkski:
Downloading it with `microk8s.enable juju` leads to lots of networking issues with restrictive firewalls/proxies/etc.

Fixes issues such as #975


balchua:
This will make microk8s binary huge.

knkski:
@balchua: The juju binary that gets included in this PR is only ~14MB, and the snap is already 222MB. It's not nothing, but wouldn't increase the size considerably.

balchua:
I see i thought juju tarball carries 116MB, atleast thats what is shown when i curl it.

knkski:
The tarball we were downloading was that big, yeah, but there's a minified version that we're also switching to in this PR (the `centos7` vs `k8s` tarball)

balchua:
Oh i see. Cool.

ktsakalozos:
LGTM +1 thank you @knkski

Issue #401, opened 2019-04-01
Provide support for ZFS


myers:
I would like full support for ZFS in microk8s.  If microk8s sees that the root fs is ZFS it should use settings that work on ZFS.

## background

With the newest release (492) microk8s switched from docker to containerd.  Because I use ZFS on all my disks this did not go well.  

I had previously altered `args/docker-daemon.json` with the directions here: https://docs.docker.com/storage/storagedriver/zfs-driver/ (adding `  "storage-driver": "zfs"`) 

With the snap autoupgrade this left me with a broken k8s, where none of the pods would start.  

After some investigation I attempted to switch containerd to use zfs:

I updated `containerd-template.toml` changing `snapshotter = "overlayfs"` to `snapshotter = "zfs"` and restarted.  Then it complained about a missing `metadata.db`.  Creating a new ZFS dataset and mounting it where it expected it made that error go away, but no pods start.  There were errors logged but I don't recall what they were.

I apologize for missing information in this issue.  The system in question is offline due to a hardware problem, but when It comes back up I will edit in more details.

myers:
with snapshotter set to "zfs" we get this:

```
$ kubectl describe pod lukluk-6677ddcbb7-cn8cm
[...]
  Warning  FailedCreatePodSandBox  9s (x9 over 117s)  kubelet, gir       Failed create pod sandbox: rpc error: code = Unknown desc = failed to reserve sandbox name "lukluk-6677ddcbb7-cn8cm_default_a67f2dcd-5a0b-11e9-8f86-002590129c4e_0": name "lukluk-6677ddcbb7-cn8cm_default_a67f2dcd-5a0b-11e9-8f86-002590129c4e_0" is reserved for "51d79673e8314d1b2677c4a6899d4cd6ad08e4754803626a3f30eff39864559f"
```

myers:
Would it be possible to mark the change from docker to containerd as a new epoch of the snap?  I'm finding containerd very poorly documented compared to docker, making it difficult to debug problem.

myers:
A solution: the problem is that microk8s started up and attempted to create containers for your pods, but is unable to finish because you are using ZFS and cannot use overlay. It's not enough to just switch the snapshotter and mount a new zfs dataset in the right location.  You need to remove all the state from these containers as well.

myers:
# so you want to use microk8s on ZFS?

1. stop microk8s.  `microk8s.stop`
[...]

PR #1000, opened 2020-03-02
Use native containerd snapshotter if ZFS is detected

sevangelatos:
This is a workaround to allow microk8s to run on ZFS systems.
See issue: #401

Using the zfs snapshotter would be another option, but currently
it seems that the zfs snapshotter requires it's own dedicated filesystem
and is still missing features such as Usage().

https://github.com/containerd/zfs/issues/17

### DIFF
--- a/microk8s-resources/wrappers/run-containerd-with-args
+++ b/microk8s-resources/wrappers/run-containerd-with-args
@@ -28,7 +28,17 @@ if [ -e "${SNAP_DATA}/var/lock/gpu" ]; then
 else
   RUNTIME="runc"
 fi
-sed 's@${SNAP}@'"${SNAP}"'@g;s@${SNAP_DATA}@'"${SNAP_DATA}"'@g;s@${RUNTIME}@'"${RUNTIME}"'@g' $SNAP_DATA/args/containerd-template.toml > $SNAP_DATA/args/containerd.toml
+
+# Determine the underlying filesystem that containerd will be running on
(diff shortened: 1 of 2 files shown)


microk8s-resources/wrappers/run-containerd-with-args
```diff
@@ -28,7 +28,17 @@ if [ -e "${SNAP_DATA}/var/lock/gpu" ]; then
 else
   RUNTIME="runc"
 fi
-sed 's@${SNAP}@'"${SNAP}"'@g;s@${SNAP_DATA}@'"${SNAP_DATA}"'@g;s@${RUNTIME}@'"${RUNTIME}"'@g' $SNAP_DATA/args/containerd-template.toml > $SNAP_DATA/args/containerd.toml
+
+# Determine the underlying filesystem that containerd will be running on
+FSTYPE=$(/usr/bin/stat -f -c %T "${SNAP_COMMON}")
```
ktsakalozos: Here you are using `/usr/bin/stat` but do we know this will exist on the host machine? We know that stat is packaged with the snap. It is better that we do not set the full path to stat here and let it resolve based on the $PATH set. According to the PATH set above the in-snap stat will be used.

sevangelatos: I agree. I will amend the code to just call stat and based on the PATH being set above, use the stat packaged with the snap.

PR #1024, opened 2020-03-11
feat: Add an option to customize the size of the container registry

### LINKED ISSUE #926: Registry addon config
Hello, since dealing with microk8s for some time now, I would like to ask if the registry can be configured to use less than the default 20GB space. 
In other words, is it known the location of the predefined .yaml manifest files that start the addons?

cyril-corbon:
fixes #926 
Description

feat: 
Add an option to the `microk8s.enable registry` command to customize the size of the container registry the default value is still set to 20Gi and the pvc cannot be lower than 20Gi
Add a test to assert that the value is equal or higher to 20Gi

### DIFF
--- a/microk8s-resources/actions/enable.registry.sh
+++ b/microk8s-resources/actions/enable.registry.sh
@@ -10,7 +10,8 @@ regex_args='([a-zA-Z]+=[a-zA-Z0-9]+(,|))'
 
 read -ra ARGUMENTS <<<"$1"
 if [ -z "${ARGUMENTS[@]}" ]; then
-  echo "Your pvc will be created with the default value of 20Gi. You can customize it with the following command eg: microk8s.enable registry:size=30Gi"
+  echo "The registry will be created with the default size of 20Gi."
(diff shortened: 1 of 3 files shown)


balchua:
Does it look like this `microk8s.enable registry:30Gi` now if i want to use a different value other than the default?

cyril-corbon:
Yes if you want to set a different value it look like this : `microk8s.enable registry:30Gi`

ktsakalozos:
Thank for this PR @cyril-corbon would it be too much to ask for a slightly change that would allow us to extend your approach with more arguments? Would you be willing to extend the syntax of arguments like this `microk8s.enable registry:size=30Gi`? I very quickly put together this awk based argument extractor `echo "val1=item1;val2=item2" | awk -F "[;, =]" '{ for(i = 1; i <= NF; i++) { if ( $i == "val1" )  print $(i+1)  } }'` but I am sure you can come up with a smarter approach. Many thanks.

cyril-corbon:
Thanks for your feed back :smiley: .
I updated the syntax of the command line as you suggest : `microk8s.enable registry:size=30Gi`. I slightly modified your logic to first extract all the key-value pair and then I treat them.

Issue #1023, opened 2020-03-11
Some commands not working with a user in the microk8s group


mystfox:
Running on a fresh Ubuntu LTS image through multipass, host OS is Windows 10 Pro, using Hyper-V. `microk8s.status`, `microk8s.kubectl` just hang when run as `ubuntu`, even though it's a member of `microk8s`. However, `microk8s.inspect`, `microk8s.stop`, `microk8s.start` at the least work as `ubuntu`. Attached the inspection report, ran as `ubuntu`, from before figuring out that stop and start worked too - but obviously after figuring out that inspect worked:
[inspection-report-20200311_105526.tar.gz](https://github.com/ubuntu/microk8s/files/4319607/inspection-report-20200311_105526.tar.gz)

```
ubuntu@microk8s-vm:~$ microk8s.status
^C
ubuntu@microk8s-vm:~$ id
uid=1000(ubuntu) gid=1000(ubuntu) groups=1000(ubuntu),4(adm),20(dialout),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),108(lxd),114(netdev),999(microk8s)
```

balchua:
Can you clean you `~/.kube` and then try again?

mystfox:
> 
> 
> Can you clean you `~/.kube` and then try again?

Ahh, thank you! It was owned by `root`, for whatever reason, and after removing, everything seems to be working fine.

ktsakalozos:
This has cause problems to a lot of our users. I think we should improve our messaging and ask users to set the right permissions to the `.kube` files when adding the user to the microk8s group.

PR #1027, opened 2020-03-12
Ask users to get ownership of their .kube when joining the microk8s group

ktsakalozos:
When users start using MicroK8s with sudo ~/.kube is created with root permissions. If they later join the microk8s group accessing the ~/.kube directory is blocked and the interaction with the cli is is very slow to the point many report commands not working. With this PR we ask users to grant access to their user when switching to the microk8s group.

Fixes: https://github.com/ubuntu/microk8s/issues/1023


ktsakalozos:
> We probably need to add this to the docs too.

Updated the quick start guide: https://github.com/canonical-web-and-design/microk8s.io/pull/228

Issue #1017, opened 2020-03-08
metallb pods in CrashLoopBackoff after enable


davecahill:
I enabled metallb with:
```microk8s.enable metallb```

This appeared successful, however the metallb pods got stuck in CrashLoopBackOff:
```
$ kubectl get pods --namespace=metallb-system
NAME                          READY   STATUS             RESTARTS   AGE
controller-65895b47d4-92zld   0/1     CrashLoopBackOff   1          12s
speaker-cjptj                 0/1     CrashLoopBackOff   1          12s
```

Logs show the same error for both pods:
```
$ kubectl logs controller-65895b47d4-92zld --namespace=metallb-system
standard_init_linux.go:207: exec user process caused "operation not permitted"

$ kubectl logs speaker-cjptj --namespace=metallb-system
standard_init_linux.go:207: exec user process caused "operation not permitted"
```

This issue seemed possibly related: https://github.com/metallb/metallb/issues/21

[inspection-report-20200308_210918.tar.gz](https://github.com/ubuntu/microk8s/files/4303182/inspection-report-20200308_210918.tar.gz)

**Comment by ktsakalozos** [comment]
Hi @davecahill 

I see some kernel problems in your dmesg. This could be potentially related to the issue you are having.

The problem seems similar to https://github.com/ubuntu/microk8s/issues/533. Could you make sure you have your system updated (`apt-get update/upgrade`)? You may also have more luck with other than the generic kernel.

Another suggestion is to remove `allowPrivilegeEscalation: false` from:
```
microk8s.kubectl edit -n metallb-system  deployment/controller
```
as mentioned in https://github.com/ubuntu/microk8s/issues/533#issuecomment-509285866

**Comment by davecahill** [comment]
Thanks @ktsakalozos !

I made the below change, rebuilt the snap, installed it, and disabled / enabled metallb - success! 🎉 
```
--- a/microk8s-resources/actions/metallb.yaml
+++ b/microk8s-resources/actions/metallb.yaml
@@ -13,7 +13,6 @@ metadata:
   name: speaker
   namespace: metallb-system
 spec:
-  allowPrivilegeEscalation: false
   allowedCapabilities:
   - NET_ADMIN
   - NET_RAW
@@ -223,7 +222,6 @@ spec:
             cpu: 100m
             memory: 100Mi
         securityContext:
-          allowPrivilegeEscalation: false
           capabilities:
             add:
             - NET_ADMIN
@@ -279,7 +277,6 @@ spec:
[...]

PR #1028, opened 2020-03-13
Fix privilege escalation of metallb on xenial

davecahill:
See https://github.com/ubuntu/microk8s/issues/1017 - metallb fails to enable correctly due to privilege escalation on Xenial.

Note that this fix is similar to https://github.com/ubuntu/microk8s/pull/551 which fixed a similar privilege escalation issue for coredns on Xenial.

### DIFF
--- a/microk8s-resources/actions/enable.metallb.sh
+++ b/microk8s-resources/actions/enable.metallb.sh
@@ -14,6 +14,15 @@ fi
 
 echo "Enabling MetalLB"
 
+ALLOWESCALATION=false
+if grep  -e ubuntu /proc/version | grep 16.04 &> /dev/null
+then
+  ALLOWESCALATION=true
+fi
+declare -A map
+map[\$ALLOWESCALATION]="$ALLOWESCALATION"
+use_manifest metallb apply "$(declare -p map)"
+
 read -ra ARGUMENTS <<< "$1"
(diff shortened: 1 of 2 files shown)


microk8s-resources/actions/enable.metallb.sh
```diff
@@ -14,6 +14,15 @@ fi
 
 echo "Enabling MetalLB"
 
+ALLOWESCALATION=false
```
davecahill: Note that unlike the [coredns fix](https://github.com/ubuntu/microk8s/pull/551), quotes are not used here or on line 20.

I initially tried with quotes, but when I installed the snap and tried to install, I got the following error message:
```
Applying registry manifest
namespace/metallb-system unchanged
error: error validating "STDIN": error validating data: ValidationError(PodSecurityPolicy.spec.allowPrivilegeEscalation): invalid type for io.k8s.api.policy.v1beta1.PodSecurityPolicySpec.allowPrivilegeEscalation: got "string", expected "boolean"; if you choose to ignore these errors, turn validation off with --validate=false
Failed to enable metallb
```

That made me think the same issue might apply to the coredns fix linked above,  but I tested several times and the coredns fix seems fine as-is.

davecahill:
Closing this, need to revisit.

PR #1043, opened 2020-03-20
Fixes and updates in addons yaml

ktsakalozos:
This PR introduces some minor updates to the yaml list of addons:
- remove the ":" causing non valid yaml output
- update the linkerd version
- juju is not an addon anymore


ktsakalozos:
Thank you @iskitsas for spotting this.

PR #1044, opened 2020-03-20
Parity between linux, macos and windows

joedborg:
This is so we don't have to re-write all the docs etc.  Is a bit annoying as this will have to be updated along with MicroK8s but is the best I can think of currently.


ktsakalozos:
Should we close this PR?

joedborg:
Yes :)

PR #1048, opened 2020-03-23
Mocking up a 30 day auto sync

joedborg:
Trying to address the issue of MicroK8sVM and the wrapper becoming out of sync.

The way this works currently is:

1. Upon install, we write to the user's `bashrc` and `zshenv` or Windows registry to source the aliases file (https://github.com/ubuntu/microk8s/pull/1048/files#diff-f2062a194b762346d37aeb4feb960c87R111).

2. After install, we enforce a sync.  This writes a checkpoint file and generates the alias file that step 1 will source (https://github.com/ubuntu/microk8s/pull/1048/files#diff-fb29bc3034a18e0c18302bb7f6fb756bR13).

3. Every time the CLI is called, the checkpoint file is checked https://github.com/ubuntu/microk8s/pull/1048/files#diff-fb29bc3034a18e0c18302bb7f6fb756bR50

4. If outdated, go to step 2.

### DIFF
--- a/installer/cli/checkpoint.py
+++ b/installer/cli/checkpoint.py
@@ -1,0 +1,84 @@
+import os
+import pickle
+from datetime import datetime
+from vm_providers.factory import get_provider_for
+
+
+CHECKPOINT_FILE = '.microk8s.checkpoint'
+CHECKPOINT_PATH = os.path.join(os.path.expanduser('~'), CHECKPOINT_FILE)
+ALIAS_FILE = '.microk8s.rc'
+ALIAS_PATH = os.path.join(os.path.expanduser('~'), ALIAS_FILE)
+
+
+def sync(force: bool = False) -> None:
+    """
+    Synchronise the alias file with the available commands.
+
+    :return: None
+    """
+    if not _check_checkpoint() or force:
+        _sync_commands()
+        _make_checkpoint()
+
+
+def _sync_commands() -> None:
+    """
+    Rewrite the alias file.
+
+    :return: None
+    """
+    if platform in ['linux', 'darwin']:
+        with open(ALIAS_PATH, 'w') as f:
+            for i in _get_microk8s_commands():
+                f.write('alias microk8s.{0}="microk8s {0}"'.format(i))
+    elif platform in ['win32']:
+        with open(ALIAS_PATH, 'w') as f:
+            for i in _get_microk8s_commands():
+                f.write('doskey microk8s.{0}=microk8s {0}'.format(i))
+
+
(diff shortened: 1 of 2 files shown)


installer/cli/checkpoint.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    Run this every time.  We're trying to keep the aliases in sync with the underlying CLI.
+
+    :return: Boolean true if okay false if outdated.
+    """
+    if os.path.isfile(CHECKPOINT_PATH):
+        with open(CHECKPOINT_PATH) as f:
+            previous = pickle.load(f)
+    
+        if (datetime.now() - previous).days > 30:
```
ktsakalozos: Why don't we compare snap revisions and we check the 30 days?

installer/cli/checkpoint.py
```diff
@@ -0,0 +1,84 @@
+import os
+import pickle
+from datetime import datetime
+from vm_providers.factory import get_provider_for
+
+
+CHECKPOINT_FILE = '.microk8s.checkpoint'
+CHECKPOINT_PATH = os.path.join(os.path.expanduser('~'), CHECKPOINT_FILE)
```
ktsakalozos: A crazy idea. We could store some of this info inside the VM. Would this work better towards cross platform?

installer/cli/checkpoint.py
```diff
@@ -0,0 +1,84 @@
...
+def _sync_commands() -> None:
+    """
+    Rewrite the alias file.
+
+    :return: None
+    """
+    if platform in ['linux', 'darwin']:
+        with open(ALIAS_PATH, 'w') as f:
+            for i in _get_microk8s_commands():
+                f.write('alias microk8s.{0}="microk8s {0}"'.format(i))
```
ktsakalozos: Will this write the same aliases as many times as it is called? Do we clean the aliases?

tvansteenburgh:
@joedborg Could you please add a description to the PR that explains how the sync works (i.e. what this code does). I would like to make sure I understand the approach w/o needing to piece it together from the code.

joedborg:
@tvansteenburgh added.

installer/cli/checkpoint.py
joedborg: This will write all aliases in 1 file (the `open` is above the `for`).  Because the `open` is in write mode, rather than append, the file will be overwritten each time.

ktsakalozos:
I guess we can close this one?

joedborg:
Yep :)

PR #1051, opened 2020-03-24
Keep only one if-condition for checking when an addon is enabled

iskitsas:
In microk8s-enable-wrapper script, two if conditions we checking for same thing: "if an addon is already enabled"

Kept the latest one committed.

Screenshot from testing local build:
keep-one-if

### DIFF
--- a/microk8s-resources/wrappers/microk8s-enable.wrapper
+++ b/microk8s-resources/wrappers/microk8s-enable.wrapper
@@ -28,7 +28,7 @@ for addon in "$@"; do
   action="$(addon_name $addon)"
   arguments="$(addon_arguments $addon)"
   # Check if the addon is already enabled, then skip.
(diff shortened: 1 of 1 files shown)


microk8s-resources/wrappers/microk8s-enable.wrapper
```diff
@@ -24,15 +24,7 @@ exit_if_no_permissions
 
 result=1
 for addon in "$@"; do
-  # Making sure the cluster is up and running befor each addon
-  STATUS=$(${SNAP}/microk8s-status.wrapper --wait-ready --timeout 30)
```
ktsakalozos, line 28: Could you add the arguments "--wait-ready --timeout 30" to the microk8s-status.wrapper call below so we wait for at most 30 seconds for any service restarts caused by enabling addons?

iskitsas: done

PR #1057, opened 2020-03-24
Pass through subcommands as if there were called directly

joedborg:
This is nearly all "automatic".  The only real issue is the `inspect` command; which will need fixing for strict anyway as `sudo` is not allowed.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -16,6 +16,8 @@ confinement: classic
 base: core
 
 apps:
+  microk8s:
+    command: microk8s.wrapper
   daemon-etcd:
     command: run-etcd-with-args
     daemon: simple
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -333,8 +333,8 @@ parts:
       echo "Creating inspect hook"
       cp $KUBE_SNAP_ROOT/scripts/inspect.sh .
 
-      # Add bash completion for microk8s.kubectl. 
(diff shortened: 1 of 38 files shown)


snapcraft.yaml
```diff
@@ -331,8 +333,8 @@ parts:
       echo "Creating inspect hook"
       cp $KUBE_SNAP_ROOT/scripts/inspect.sh .
 
-      # Add bash completion for microk8s.kubectl. 
-      ./kubectl completion bash | sed "s/complete -o default -F __start_kubectl kubectl/complete -o default -F __start_kubectl microk8s.kubectl/g" | sed "s/complete -o default -o nospace -F __start_kubectl kubectl/complete -o default -o nospace -F __start_kubectl kubectl/g" > kubectl.bash
+      # Add bash completion for microk8s kubectl.
+      ./kubectl completion bash | sed "s/complete -o default -F __start_kubectl kubectl/complete -o default -F __start_kubectl microk8s kubectl/g" | sed "s/complete -o default -o nospace -F __start_kubectl kubectl/complete -o default -o nospace -F __start_kubectl kubectl/g" > kubectl.bash
```
knkski: Is it worth keeping both of the completions? It seems useful to keep around for the snap users.

joedborg: Yeah I agree

ktsakalozos:
@joedborg do you think we could add subcommand descriptions in the `microk8s --help`? We do something very similar in https://github.com/ubuntu/microk8s/blob/master/installer/common/definitions.py#L3

joedborg:
Yeah, let me think how we can make those shared (this would also remove the need for the ls)

PR #1076, opened 2020-03-30
Feature/extend cluster-agent REST API

iskitsas:
Expose more APIs from cluster agent to support manage and view operations for MicroK8s, for easy integration with other apps, like dashboards.

The exposed APIs will help to integrate the [MicroK8s Web Console](https://github.com/sastix/microk8s-addons) ([microdash](https://hub.docker.com/repository/docker/sastix/microdash)) as an AddOn.

Along with the new APIs, I am enabling Swagger-UI as a complete documentation page and easy testing web tool for the endpoints. The page is exposed under https://127.0.0.1:25000/swagger

Attaching some screenshots:
microk8s-swaggerui-home
microk8s-swaggerui-version
microk8s-swaggerui-version-tryout
microk8s-swaggerui-configure
microk8s-swaggerui-get-services
microk8s-swaggerui-overview-tryout
microk8s-swaggerui-addon-enable
microk8s-swaggerui-service-logs
microk8s-swaggerui-service-stop


balchua:
Quick question, can `/status` and `/version` be `GET` requests?  I don't see any security implications on these.  Unless i missed it.

iskitsas:
> I like this PR, but i do have a quick question, can `/status` and `/version` be `GET` requests? I don't see any security implications on these. Unless i missed it.

Yes it is very easy. Just keep in mind that we send a body with the POST request where we have the callback token for validation, eg:

```
{
  "callback": "xyztoken"
}
```

It is not common to send body with the GET request (it is possible though). The callback token could be replaced with a query parameter, but this will **expose the token on the url**, eg:

```
https://127.0.0.1:25000/cluster/api/v1.0/version?cbtoken=xyztoken
```

OR 

we could remove the token validation for those 2 calls.

So it is just decision we have to make :)

balchua:
Im thinking for `GET` method there is no need to pass tokens.  @ktsakalozos wdyt?

balchua:
I have another request if possible.  What about a secure method to get the kubeconfig too. 😁

iskitsas:
> I have another request if possible. What about a secure method to get the kubeconfig too.

Yes, it is possible :)

iskitsas:
> > I have another request if possible. What about a secure method to get the kubeconfig too.
> 
> Yes, it is possible :)
Hi @balchua, @ktsakalozos, 

I have updated the PR by adding a new POST **/config** endpoint for getting microk8s.config info and expose **/version** and **/status** under a GET request, with no need to pass the callback token. 

In order to enable those, **you must create a file** with the name 'enable-extended-api' under '/var/snap/microk8s/current/args/enable-extended-api', HAVING as content THE SAME TOKEN VALUE with '/var/snap/microk8s/current/credentials/callback-token.txt'. 

For the POST /config the user must also send the callback token in the body of the request as json (like we are doing for the other POST endpoints), and for the GET /version and /status the user does not need to send it.

I have also updated the README with full examples to test it from a shell with curl, and swaggerUI so you can try them from browse.

Some screenshots from Swagger UI with new/updated endpoints:

Home:
swagger-ui-home

GET-version-with-disabled-extended-api:
swagger-ui-GET-version-with-disabled-extended-api

GET-version-with-enabled-extended-api:
swagger-ui-GET-version-with-enabled-extended-api

GET-status-with-enabled-extended-api:
swagger-ui-GET-status-with-enabled-extended-api

GET-config-with-enabled-extended-api:
swagger-ui-GET-config-with-enabled-extended-api

balchua:
Thanks @iskitsas.  The more i think about this, i was thinking we should put the token as part of the header, just like what most token based authn/authz systems do.
For example `Authorization Bearer:`

This way it is easy to determine which methods do not pose side effects like `GET`.

What do you guys think?

iskitsas:
> Thanks @iskitsas. The more i think about this, i was thinking we should put the token as part of the header, just like what most token based authn/authz systems do.
> For example `Authorization Bearer:`
> 
> This way it is easy to determine which methods do not pose side effects like `GET`.
> 
> What do you guys think?

Can easily be changed to header based checking :)

iskitsas:
> > Thanks @iskitsas. The more i think about this, i was thinking we should put the token as part of the header, just like what most token based authn/authz systems do.
> > For example `Authorization Bearer:`
> > This way it is easy to determine which methods do not pose side effects like `GET`.
> > What do you guys think?
> 
> Can easily be changed to header based checking :)

Just added another way to pass the token. You can now use the header "Callback-Token: <value>" to validate

scripts/cluster/agent.py
```diff
@@ -423,6 +443,285 @@ def configure():
...
+    v = callback_token_validation(request)
+    if not v["valid"]: return v["response"]
+    if "addon" not in request.json or len(request.json["addon"].strip()) == 0:
+        error_msg = {"error": "Empty addon provided"}
+        return Response(json.dumps(error_msg), mimetype='application/json', status=400)
+    output = subprocess.check_output("{}/microk8s-disable.wrapper {}".format(snap_path, request.json["addon"]).split())
+    return output
+
+
+@app.route('/{}/start'.format(CLUSTER_API), methods=['POST'])
```
ktsakalozos, line 611: Given you will need to start MicroK8s from the command line, I do not think we should allow the stopping of services via the REST api. It would be possible users may be surprised by this.

ktsakalozos: Also, I understand you know why this call is not going to be called.

scripts/cluster/agent.py
```diff
@@ -423,6 +443,285 @@ def configure():
...
+    """
+    Web call for microk8s.stop
+    """
+    v = callback_token_validation(request)
+    if not v["valid"]: return v["response"]
+    output = subprocess.check_output("{}/microk8s-stop.wrapper".format(snap_path).split())
+    return output
+
+
+@app.route('/{}/overview'.format(CLUSTER_API), methods=['POST'])
```
ktsakalozos, line 633: Getting access tot he cluster like this is not something we recommend. If you want to have an overview of the cluster you can query the cluster's API server directly.

scripts/cluster/agent.py
```diff
@@ -423,6 +443,285 @@ def configure():
...
+    """
+    Web call to get the microk8s.kubectl get all --all-namespaces
+    """
+    v = callback_token_validation(request)
+    if not v["valid"]: return v["response"]
+    output = subprocess.check_output("{}/microk8s-kubectl.wrapper get all --all-namespaces".format(snap_path).split())
+    return output
+
+
+@app.route('/{}/status'.format(CLUSTER_API), methods=['POST', 'GET'])
```
ktsakalozos, line 644: I am not sure about the GET part. Yes it makes sense to be a get but you need to pass the callback token in a secure way.

scripts/cluster/agent.py
```diff
@@ -423,6 +443,285 @@ def configure():
...
+
+    if request.json and "addon" in request.json:
+        json_output = {"addon": request.json["addon"], "status": output.decode().strip('\n')}
+    else:
+        json_output = yaml.full_load(output)
+
+    return app.response_class(response=json.dumps(json_output, sort_keys=False, indent=4), status=200, mimetype='application/json')
+
+
+@app.route('/{}/config'.format(CLUSTER_API), methods=['POST'])
```
ktsakalozos, line 669: I do not think you need this for MicroDash. Please make a case why this is needed in a separate PR.

scripts/cluster/agent.py
```diff
@@ -423,6 +443,285 @@ def configure():
...
+    with open(os.path.join(os.path.dirname(__file__), "static/swagger.yaml")) as f:
+        content = f.read()
+    return app.response_class(response=json.dumps(yaml.full_load(content), sort_keys=False, indent=4), status=200, mimetype='application/json')
+
+
+def callback_token_validation(request):
+    if 'Content-Type' in request.headers and request.headers['Content-Type'] == 'application/json':
+        callback_token = request.json['callback']
+    elif 'Callback-Token' in request.headers:
+        callback_token = request.headers['Callback-Token']
```
ktsakalozos, line 699: Does this need to be part of this PR? Could you put this in a saparate PR so we can discuss with @balchua the name of the header we would like to pass the token in?

scripts/cluster/agent.py
```diff
@@ -423,6 +443,285 @@ def configure():
...
+        callback_token = request.form['callback']
+    valid = is_valid(callback_token, callback_token_file)
+    resp = None
+    if not valid:
+        error_msg = {"error": "Invalid token"}
+        resp = Response(json.dumps(error_msg), mimetype='application/json', status=500)
+    return {"valid": valid, "response": resp}
+
+
+def is_extended_api_enabled():
```
ktsakalozos, line 710: Same, here does this need to be part of the MicroDash PR? I understand this is a flag to enable/disable certain functionalities. I think this belongs into its own PR so we discuss on what we want to feature-flag and how. Thank you.

scripts/cluster/agent.py
```diff
@@ -423,6 +443,285 @@ def configure():
...
+    v = callback_token_validation(request)
+    if not v["valid"]: return v["response"]
+    if "service" not in request.json or len(request.json["service"].strip()) == 0:
+        error_msg = {"error": "Empty service provided"}
+        return Response(json.dumps(error_msg), mimetype='application/json', status=400)
+    output = subprocess.check_output("systemctl disable snap.{}.service".format(request.json["service"]).split())
+    return output
+
+
+@app.route('/{}/service/logs'.format(CLUSTER_API), methods=['POST'])
```
ktsakalozos, line 561: Could you put the log collection under the status endpoint? Thank you.

scripts/cluster/agent.py
balchua: Just my 2 cents, would passing the token through header be acceptable?
The same thing as what most APIs are doing like oauth2.
Wdyt?

scripts/cluster/agent.py
balchua: I may have a use case if this method returns the kubeconfig? 
If yes, it can be helpful when working with clusterapi when u have to return the kubeconfig through kubernetes secret.

scripts/cluster/agent.py
iskitsas: The way it is implemented now, is allowing GET request with no token passed, if and only if 'enable-extended-api' is true.

Additionally, passing the token through Header or a json body, both ways are supported. Token is never exposed though a query param

scripts/cluster/agent.py
iskitsas: This is only accessible if you 'enable-extended-api', passing also the callback-token using a Header or the json body

scripts/cluster/agent.py
iskitsas: I could make a separate PR, however It is just 2 lines of code and a decision for the name. I would propose @balchua to have a look, and if still is the need for a separate PR, I can do it

scripts/cluster/agent.py
iskitsas: Do you mean /status/service/logs? Or extend the json body which is sent to /status, to support the service logs with new properties?

eg on /status:
`{"callback":"xyztoken", "servicelogs":{"name":"microk8s.daemon-flanneld", "lines":30}}

scripts/cluster/agent.py
iskitsas: will do in a separate PR

scripts/cluster/agent.py
iskitsas: Will remove it

scripts/cluster/agent.py
```diff
@@ -423,6 +443,285 @@ def configure():
...
+    if request.method.upper() == 'POST':
+        v = callback_token_validation(request)
+        if not v["valid"]: return v["response"]
+    output = subprocess.check_output("snap info microk8s".split())
+
+    json_output = yaml.full_load(output)
+    return json_output["installed"].split(' ')[0]
+
+
+@app.route('/{}/services'.format(CLUSTER_API), methods=['POST'])
```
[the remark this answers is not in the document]
iskitsas: will remove it from this PR and make a separate, with the suggestions for making a cleaner, more readable and domain specific API

scripts/cluster/agent.py
```diff
@@ -12,18 +13,37 @@
...
 certs_request_tokens_file = "{}/credentials/certs-request-tokens.txt".format(snapdata_path)
+enable_extended_api_file = "{}/args/enable-extended-api".format(snapdata_path)
 default_port = 25000
 default_listen_interface = "0.0.0.0"
-
+ext_api_error_message = "{} request is not allowed if Extended API is not enabled. You can enable it by creating " \
+                        "a file with the name 'enable-extended-api' under '" + enable_extended_api_file + "', " \
+                        "HAVING THE SAME TOKEN VALUE with '" + callback_token_file + "'"
+
+# -- swagger specific --
```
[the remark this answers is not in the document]
iskitsas: Will introduce a new approach on separate PR

scripts/cluster/agent.py
ktsakalozos: The cluster agent API is not meant to be a proxy for the k8s API server.

MicroDash runs as a pod in MicroK8s and thus should access to the cluster API server as described in the official docs [1]. This way we comply with the authn & authz policies set by the cluster administrator.

[1] https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/#accessing-the-api-from-within-a-pod

Issue #1093, opened 2020-04-04
command not found after install


sweetpotatoz:
`sudo snap install microk8s --classic --channel=1.18/stable`
microk8s (1.18/stable) v1.18.0 from Canonical✓ installed

`sudo microk8s status --wait-ready`
Traceback (most recent call last):
  File "/snap/microk8s/1319/scripts/wrappers/status.py", line 149, in <module>
    available_addons = get_available_addons(get_current_arch())
  File "/snap/microk8s/1319/scripts/wrappers/common/utils.py", line 113, in get_available_addons
    addons = yaml.load(file, Loader=yaml.FullLoader)
AttributeError: module 'yaml' has no attribute 'FullLoader'

lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.4 LTS
Release:        18.04
Codename:       bionic

uname -r
4.15.0-91-generic

```
microk8s
zsh: command not found: microk8s
```
snap services show services are running, snap logs for each of them, I not do see errors except flannelId, pasted below

snap services                          
Service                             Startup  Current  Notes
microk8s.daemon-apiserver           enabled  active   -
microk8s.daemon-apiserver-kicker    enabled  active   -
microk8s.daemon-cluster-agent       enabled  active   -
microk8s.daemon-containerd          enabled  active   -
microk8s.daemon-controller-manager  enabled  active   -
microk8s.daemon-etcd                enabled  active   -
microk8s.daemon-flanneld            enabled  active   -
microk8s.daemon-kubelet             enabled  active   -
microk8s.daemon-proxy               enabled  active   -
microk8s.daemon-scheduler           enabled  active   -

```
sudo snap logs microk8s.daemon-flanneld          
2020-04-04T12:05:01Z microk8s.daemon-flanneld[16703]: I0404 20:05:01.342266   16703 main.go:386] Found network config - Backend type: vxlan
2020-04-04T12:05:01Z microk8s.daemon-flanneld[16703]: I0404 20:05:01.342416   16703 vxlan.go:120] VXLAN config: VNI=1 Port=0 GBP=false DirectRouting=false
2020-04-04T12:05:01Z microk8s.daemon-flanneld[16703]: I0404 20:05:01.389635   16703 local_manager.go:234] Picking subnet in range 10.1.1.0 ... 10.1.255.0
[...]

PR #1094, opened 2020-04-05
[Feature Request] TTL based bootstrap token

balchua:
This PR can create a TTL based cluster bootstrap token.

Use case:
* Allows adding multiple nodes using one token.
* Make it easier to automate the creation of cluster using terraform or ansible.

Introduced new wrapper `microk8s add-token --token <custom token> --ttl <time to live in seconds>`
It will store the expiry time in the `cluster-tokens.txt` file, separated by `|`

Sample:
```shell
$ microk8s add-token --token y6tG0w1uG1liSwx5SIln --ttl 3600

$ cat /var/snap/microk8s/current/credentials/cluster-tokens.txt 
y6tG0w1uG1liSwx5SIln|1586072159
```
When joining a node, it first remove all expired tokens before checking for its validity.


balchua:
Was thinking maybe we shud just make this generic in such a way that it can add different type of token such as `bootstrap` or `callback`.

Thoughts anyone?

ktsakalozos:
> make this generic in such a way that it can add different type of token such as `bootstrap` or `callback`.

What do you have in mind? Can you offer some UX examples?

balchua:
Something like this:
 `microk8s add-token --type bootstrap --token gagghdheh123 --ttl 3600`
This one goes to the `cluster-token` file
 `microk8s add-token --type callback --token PkYFjGGY1g453 --ttl 3600`
While this one goes to the `callback-token` file.

snapcraft.yaml
```diff
@@ -53,6 +53,8 @@ apps:
     completer: kubectl.bash
   add-node:
     command: microk8s-add-node.wrapper
+  add-token:
```
ktsakalozos, line 56: I am not much in favor of exposing the tokens mechanism to the end user. I believe it is something internal and if there is a real need lets keep it internal so we can change it in the future if we need to.

microk8s-resources/wrappers/microk8s-add-node.wrapper
```diff
@@ -37,11 +37,8 @@ exit_if_stopped
 # We throw away some randomness by only selecting bytes in the A-Za-z range and taking the first 32 such bytes,
 # so instead of 32*8 = 256 bits of randomness you get approx 182 bits. This is good enough for this short lived tokens.
 token=$(< /dev/urandom tr -dc A-Za-z | head -c32;echo;)
-echo ${token} >> $SNAP_DATA/credentials/cluster-tokens.txt
-if getent group microk8s >/dev/null 2>&1
-then
-  chgrp microk8s $SNAP_DATA/credentials/cluster-tokens.txt || true
-fi
+# Call add-token wrapper to add it to the cluster-tokens.txt file.
+$SNAP/microk8s-add-token.wrapper --token $token --ttl 3600
```
ktsakalozos, line 41: Instead of setting the token to always `--ttl 3600` let's keep the old default in the `add-node`  command.

We could have:
```
microk8s.add-node # will give you a one time token
microk8s.add-node --ttl-token 4000 # will give you a token that expires after 4000 seconds
microk8s.add-node --multinode-token 40 # will give you a token that can be used 40 times
...
```

scripts/cluster/agent.py
```diff
@@ -280,12 +283,14 @@ def join_node():
...
 
+    # Remove expired tokens
+    remove_expired_token_from_file(cluster_tokens_file)
+
     if not is_valid(token):
         error_msg={"error": "Invalid token"}
         return Response(json.dumps(error_msg), mimetype='application/json', status=500)
 
     add_token_to_certs_request(token)
-    remove_token_from_file(token, cluster_tokens_file)
```
ktsakalozos, line 288: Lets keep the old type of tokens for backwards compatibility. The ttl based tokens may be considered unsafe in some cases.

scripts/cluster/agent.py
```diff
@@ -228,7 +228,10 @@ def is_valid(token_line, token_type=cluster_tokens_file):
 
     with open(token_type) as fp:
         for _, line in enumerate(fp):
-            if token == line.strip():
+            token_in_file = line.strip()
+            if "|" in line :
+                token_in_file = line.strip().split('|')[0]
```
ktsakalozos, line 233: Shouldn't we be checking the validity of the token before returning it? I think it would make the code much cleaner if we did this check here at the source.

scripts/cluster/add_token.py
```diff
@@ -0,0 +1,39 @@
+import os
```
ktsakalozos, line 1: This file could (with changes) replace the add-node wrapper.

snapcraft.yaml
balchua: One of the idea of being able to specify a token by a user, is that it is difficult to extract the token automatically if it is generated.
This is also a feature which kubeadm also uses to allow it to be easily used by automation tools such as terraform or ansible.

scripts/cluster/agent.py
balchua: Yes we can remove all expired token just before checking for validity if that is fine.

microk8s-resources/wrappers/microk8s-add-node.wrapper
balchua: Interesting feature, instead of time to live, its based on number of nodes.  Didn't thought about that.
Let me see how to implement that.

snapcraft.yaml
balchua: Can we allow `add-node` to take in a token args? We can then hide the `add-token` for public use.

balchua: Make add-token wrapper private.  Still maintain the usual way `microk8s.add-node`, will generate one time token use.

balchua:
I removed the add-token wrapper, instead leverage on existing `add-node` wrapper.
To create a one time token, it remains the same `microk8s add-node`.
To create a TTL based token, use `microk8s add-node --token [32 chars long] --ttl_token 3600`

Thank you for reviewing.

scripts/cluster/add_token.py
```diff
@@ -0,0 +1,54 @@
...
+            fp.write(token_with_expiry.format(token, expiry))
+        else:
+            fp.write(token_without_expiry.format(token))
+
+
+if __name__ == '__main__':
+
+    # initiate the parser with a description
+    parser = argparse.ArgumentParser(description='Produce a connection string for a node to join the cluster.', prog='microk8s add-node')
+    parser.add_argument("--ttl_token", "-l", help="Specify how long the token is valid, before it expires.  Value of \"-1\" indicates that the token is usable only once (i.e. after joining a node, the token becomes invalid)", type=int,
```
tvansteenburgh, line 35: I would prefer `--token-ttl` here.

balchua: No problem with that.

balchua:
command to add a node
```shell
# default one time token usage.
microk8s add-node

# TTL based token with 1 hour expiry
microk8s add-node --token-ttl 3600 --token qwertyuiopasdfghjklzxcvbnm123456
```

ktsakalozos:
The tests timing out were run manually.

This also addresses https://github.com/ubuntu/microk8s/issues/1093

balchua:
Thanks @ktsakalozos 👍

PR #1109, opened 2020-04-13
[New Addon] Provide a new addon to enable access to host services from pods

iskitsas:
The **host-ip** addon enables access from pods to local host services from a specific local ip, defined when enabling the addon.

Related with #351

Examples of usage:
- adds 10.0.7.17 by default
`
microk8s.enable host-ip
`
- user defined
`
microk8s.enable host-ip:9.0.0.2
`
- validating wrong ip when defined by user
`microk8s.enable host-ip:A.B.C.D
Bad IP[A.B.C.D]. Please provide a valid IP address. Default is 10.0.7.17
`
- disable addon
`microk8s.disable host-ip
`


microk8s-resources/actions/disable.host-ip.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+
+echo "Disabling host-ip"
+
+if [ -f "${SNAP_DATA}/var/lock/host-ip-enabled" ]
+then
+  IP_ADDRESS=$(<"${SNAP_DATA}/var/lock/host-ip-enabled")
+  sudo ifconfig lo:17 $IP_ADDRESS down
```
ktsakalozos, line 10: what is the 17 in lo:17?

microk8s-resources/actions/enable.host-ip.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -e
+
+source $SNAP/actions/common/utils.sh
+
+read -ra IP_ADDRESS <<< "$1"
```
ktsakalozos, line 7: Could you tweak this a bit so the arguments passed are named ie `host-ip:ip=10.0.1.1` vs `host-ip:10.0.1.1`. You can look at the registry addon on how this can be done. Thanks

microk8s-resources/actions/disable.host-ip.sh
ktsakalozos, line 10: There is a utility function (in utils) called `run_with_sudo` that makes better handling of libs. Could you also make sure we bring ifconfig in the snap and call it from there (use full path under /snap/....)?

iskitsas: just a random number, not to conflict with any other existing lo:interface-number which a user might have added. Change it to lo:1?

microk8s-resources/actions/enable.host-ip.sh
iskitsas: sure, I wanted to avoid the repeat of word 'ip'. Another suggestion would be `host-ip:address=10.0.1.1`. So, `host-ip:ip` or `host-ip:address`?

balchua: Perhaps call it `microk8s.host-access`?

iskitsas: @ktsakalozos you call :)

microk8s-resources/actions/disable.host-ip.sh
iskitsas: done

microk8s-resources/actions/enable.host-ip.sh
iskitsas: done as proposed initially by @ktsakalozos

microk8s-resources/actions/disable.host-ip.sh
iskitsas: changed to lo:1

iskitsas:
A screenshot from when testing it locally:
microk8s-host-ip-addon-usage

microk8s-resources/actions/disable.host-ip.sh
ktsakalozos: Shouldn't we try to select a number that will not conflict with the existing user setup?

microk8s-resources/actions/enable.host-ip.sh
ktsakalozos: Lets do `host-address:ip=x.y.z.j` to keep @balchua happy :)

microk8s-resources/actions/disable.host-ip.sh
ktsakalozos: Could you call the interface lo:microk8s instead of trying to do arithmetic with the lo: interfaces?

iskitsas: done

microk8s-resources/actions/enable.host-ip.sh
iskitsas: done

ktsakalozos:
Thank you @iskitsas for the PR.

@knkski with this addon if you have a service running on the host you can assume it is listening on the designated port. We were talking about this some weeks ago. Also note this is for a single node.

PR #1110, opened 2020-04-13
Check total memory before enabling kubeflow

knkski:
Adds a minimum recommended amount of memory that will get checked against total system memory before enabling kubeflow. Allows overriding the check with an environment variable.

### DIFF
--- a/microk8s-resources/actions/enable.kubeflow.sh
+++ b/microk8s-resources/actions/enable.kubeflow.sh
@@ -12,6 +12,8 @@ import time
 from itertools import count
 from distutils.util import strtobool
 
+MIN_MEM_GB = 14
+
 
 def run(*args, die=True, debug=False, stdout=True):
     # Add wrappers to $PATH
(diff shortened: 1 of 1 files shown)


microk8s-resources/actions/enable.kubeflow.sh
```diff
@@ -108,6 +110,24 @@ def main():
     no_proxy = os.environ.get("KUBEFLOW_NO_PROXY") or None
     hostname = os.environ.get("KUBEFLOW_HOSTNAME") or None
     debug = strtobool(os.environ.get("KUBEFLOW_DEBUG") or "false")
+    ignore_min_mem = strtobool(os.environ.get("KUBEFLOW_IGNORE_MIN_MEM") or "false")
+
+    with open("/proc/meminfo") as f:
```
ktsakalozos, line 115: @joedborg I wonder if we will get into trouble with strict confinement if we try to reacj `/proc/meminfo`. Do you know that?

ktsakalozos: I guess we will care about this when the time comes.

PR #1119, opened 2020-04-15
Script to transition to calico cni

ktsakalozos:
This PR introduces:
- the calico cni manifest and a way to produce it
- an upgrade script to apply the manifest and rollback in the context of a cluster


ktsakalozos:
Merging this to continue with the reset of the upgrade scripts

Issue #390, opened 2019-03-27
Trust insecure-registry "registry.container-registry" by default


robinchm:
Currently to pull images from the bundled private registry one needs to use the address "localhost:32000".

When DNS is also enabled, the private registry service is assigned a DNS record "registry.container-registry". From any pod on the cluster, `curl registry.container-registry:5000/v2/` returns expected result. But using this address for pulling images in the pod creation will fail:
```
  Warning  Failed     2m7s (x4 over 3m36s)  kubelet, xxx-xxxxxxx  Failed to pull image "registry.container-registry:5000/box": rpc error: code = Unknown desc = failed to resolve image "registry.container-registry:5000/box:latest": no available registry endpoint: failed to do request: Head https://registry.container-registry:5000/v2/box/manifests/latest: dial tcp: lookup registry.container-registry: no such host
```
From what I understand it tries to pull via https because this address is not added to insecure-registries when docker spins up. "localhost:32000" works because that address is added to the docker config.

Given that both DNS and registry are bundled DNS, can "registry.container-registry" be added to docker config so that we can use it in pods' yaml?

ktsakalozos:
Hi @robinchm 

Can you please try this, edit `/var/snap/microk8s/current/args/containerd-template.toml` and right under the  `localhost:32000` entry add (not sure about the https vs http):
```
      [plugins.cri.registry.mirrors."local.insecure-registry.io"]
          endpoint = ["http://registry.container-registry:5000"]
```

Then do a `microk8s.stop` and after a few seconds a `microk8s.start`.

If that works we could put it in https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/default-args/containerd-template.toml#L52 so that it is there by default.

carlososiel:
Hi @ktsakalozos 
I have the same issue, and I tried with these configuration, but still try to pull from https instead of http.

ktsakalozos:
@carlososiel can you please provide me with a way to reproduce the error you see? This "pull from https instead of http" sounds like a docker daemon needs to be configured to use the insecure registry MicroK8s provides. Please have a look at the comment https://github.com/ubuntu/microk8s/issues/382#issuecomment-477174066  where we have to tell the local dockerd that there is an insecure registry it can push images to.

carlososiel:
@ktsakalozos I have installed version 1.14 of microk8s, with docker installed in version 18.09.3. The docker daemon has configurations for use insecure registry and with docker info show this information: 
```
[...]

PR #1122, opened 2020-04-16
Update containerd-template.toml

SamyCoenen:
Allow Microk8s to pull images from it's own private registry.
Especially useful for local development with tools like Skaffold.

Steps to enable:
```sh
microk8s.enable registry
microk8s.enable dns
microk8s.stop
microk8s.start
```

Fixes https://github.com/ubuntu/microk8s/issues/390


ktsakalozos:
Thank you for this PR @SamyCoenen. Can you point me to a tutorial or a use case where this fix exercised? I am interested to know how Skaffold is taking these internal dns entries (eg registry.container-registry.svc) and exposes them to containerd running on the host.

SamyCoenen:
@ktsakalozos I'm using it with a workflow that builds the images inside the cluster with Kaniko, deploys the new tags using Helm and syncs file changes to running containers. 
What you are referring to
>  I am interested to know how Skaffold is taking these internal dns entries (eg registry.container-registry.svc) and exposes them to containerd running on the host.

is an open issue, as the host doesn't need to have access to the registry: https://github.com/GoogleContainerTools/skaffold/issues/2217
I solved this by doing 2 things for now:
```sh
echo '127.0.0.1 registry.container-registry.svc.cluster.local' | sudo tee --append /etc/hosts
kubectl port-forward service/registry -n container-registry 5000:5000
```

an example Skaffold file might start with this

```yml
apiVersion: skaffold/v2beta1
kind: Config
build:
  cluster: {}
  insecureRegistries:
    - registry.container-registry.svc.cluster.local:5000
  artifacts:
    - image: registry.container-registry.svc.cluster.local:5000/my-image
```

SamyCoenen:
The kubelet can pull the images because the dns addon updates it's configuration
https://github.com/ubuntu/microk8s/blob/395bc12671a12ca14e79f5a72d06765873068210/microk8s-resources/actions/enable.dns.sh#L23

This results in the following `kubelet` configuration file on my host at `/var/snap/microk8s/current/args/kubelet`

```sh
--kubeconfig=${SNAP_DATA}/credentials/kubelet.config
--cert-dir=${SNAP_DATA}/certs
--client-ca-file=${SNAP_DATA}/certs/ca.crt
--anonymous-auth=false
--network-plugin=cni
--root-dir=${SNAP_COMMON}/var/lib/kubelet
--fail-swap-on=false
--cni-conf-dir=${SNAP_DATA}/args/cni-network/
--cni-bin-dir=${SNAP}/opt/cni/bin/
--feature-gates=DevicePlugins=true
--eviction-hard="memory.available<100Mi,nodefs.available<1Gi,imagefs.available<1Gi"
--container-runtime=remote
--container-runtime-endpoint=${SNAP_COMMON}/run/containerd.sock
--node-labels="microk8s.io/cluster=true"
--runtime-cgroups=/systemd/system.slice
--kubelet-cgroups=/systemd/system.slice
--cluster-domain=cluster.local
--cluster-dns=10.152.183.10
```

ktsakalozos:
It is hard to accept this PR because we will be adding entries for a registry that is not available. The user would still need to expose the dns entry of the hosted registry.

Have you tried the approach outlined in [1, 2] ?

[1] https://skaffold.dev/docs/environment/image-registries/
[2] https://odino.org/local-k8s-development-in-2020/

SamyCoenen:
@ktsakalozos Alright, if the DNS entry is not acceptable then 2 options remain:
- Wait for this issue to be fixed https://github.com/GoogleContainerTools/skaffold/issues/2217
- Use Docker to build the images instead of Kaniko as suggested https://github.com/ubuntu/microk8s/pull/1122#issuecomment-617023823

balchua:
I actually hit this issue too when working with kaniko.  I end up adding the host ip to the cri registry config.
Ex.
"192.168.1.1:32000"

Obviously expose the service via `NodePort`

SamyCoenen:
@balchua 192.168.1.x/24 is a private network, it could be 172.16.x.x or 192.168.0.x for others every time they start their system so that is why I didn't go with that. It would work at my home network but not at work for example.

balchua:
Absolutely right. Didn't thought about that scenario.

PR #1123, opened 2020-04-17
Introduce GH action

ktsakalozos:
Getting GH actions to build the project so we get the proposed snap.

Thank you @joedborg for this work.


ktsakalozos:
It is very convenient that we can have snaps build on each PR as part of the GH action artifact. Thank you @joedborg

PR #1134, opened 2020-04-22
MacOS disk space check

joedborg:
Check MacOS disk space remaining.

### DIFF
--- a/installer/cli/microk8s.py
+++ b/installer/cli/microk8s.py
@@ -7,7 +7,7 @@ from sys import exit, platform
 import click
 
 from cli.echo import Echo
-from common.auxillary import Windows
+from common.auxillary import Windows, MacOS
 from common.errors import BaseError
 from vm_providers.factory import get_provider_for
 from vm_providers.errors import ProviderNotFound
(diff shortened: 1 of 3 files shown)


joedborg:
[MicroK8s Installer - MacOS restrictions](https://trello.com/c/v1uGqEWp/1459-microk8s-installer-macos-restrictions)

installer/cli/microk8s.py
```diff
@@ -142,6 +142,12 @@ def install(args) -> None:
                 echo.error('Cannot continue without enabling Hyper-V')
                 exit(1)
 
+    if platform == 'darwin':
+        if not MacOS().is_enough_space():
```
ktsakalozos, line 146: Given that the disk space is typically dynamically allocated do we really want to stop the installation? Is disk space pre-allocated with multipass on win and mac?

Even if we want to ensure the disk space can be pre-allocated, shouldn't we check for the disk space the user asked for with the `--disk` argument (that may be the default size of 50GB)?

ktsakalozos: Have a look at this https://github.com/canonical/multipass/issues/62#issuecomment-360060429 . A warning would be more suitable.

joedborg: Yeah, we’ve been asked to do it this way and it’s already in the Windows code.

Issue #1139, opened 2020-04-24
Microk8s network implementation is faulted. SNAT issue.


peix2:
Default installation of Microk8s fails test from "Kubernetes tutorial" described here:
https://kubernetes.io/docs/tutorials/services/source-ip/

Full output of test below:

```
px2@sm3:~$ kubectl config set-context --current --namespace=test
Context "microk8s" modified.
px2@sm3:~$ kubectl create deployment source-ip-app --image=k8s.gcr.io/echoserver:1.4

deployment.apps/source-ip-app created
px2@sm3:~$ 
px2@sm3:~$ kubectl expose deployment source-ip-app --name=clusterip --port=80 --target-port=8080
service/clusterip exposed
px2@sm3:~$ kubectl run busybox -it --image=busybox --restart=Never --rm^C
px2@sm3:~$ kubectl get svc
NAME        TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)   AGE
clusterip   ClusterIP   10.152.183.84   <none>        80/TCP    17s
px2@sm3:~$ kubectl run busybox -it --image=busybox --restart=Never --rm
If you don't see a command prompt, try pressing enter.
/ # ip a 
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
3: eth0@if7: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1450 qdisc noqueue 
    link/ether 06:e8:ba:a1:40:d9 brd ff:ff:ff:ff:ff:ff
    inet 10.1.95.3/24 scope global eth0
       valid_lft forever preferred_lft forever
/ # wget -qO - 10.152.183.84
CLIENT VALUES:
client_address=10.1.95.1
command=GET
real path=/
query=nil
request_version=1.1
request_uri=http://10.152.183.84:8080/

SERVER VALUES:
server_version=nginx: 1.10.0 - lua: 10001

HEADERS RECEIVED:
connection=close
host=10.152.183.84
user-agent=Wget
BODY:
-no body in request-/ # 
```

IP address expected to be presented is busybox pod IP address - 10.1.95.3
IP address presented as source reach out service - 10.1.95.1

This means that packets are being SNATed. 
[...]

PR #1145, opened 2020-04-26
Set IP masq to true

ktsakalozos:
@Cynerva this is the PR with the ip-masq fix. It seems to be working fine for me. Any reasons why we would not need to turn ip-masq n by default? Thanks.

Fixes https://github.com/ubuntu/microk8s/issues/1139

### DIFF
--- a/microk8s-resources/default-args/flanneld
+++ b/microk8s-resources/default-args/flanneld
@@ -4,3 +4,4 @@
 --etcd-certfile=${SNAP_DATA}/certs/server.crt
 --etcd-keyfile=${SNAP_DATA}/certs/server.key
 --subnet-file=${SNAP_COMMON}/run/flannel/subnet.env
+--ip-masq=true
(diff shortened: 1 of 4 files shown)


Cynerva:
Setting --ip-masq on flanneld is necessary to prevent SNAT between pods on different nodes, so this is a good change.

Also +1 to the kube-proxy --cluster-cidr change.

One thing to be aware of. This will fix new deployments, but it may not fix upgraded deployments that already have pods. This is because flannel CNI is what creates the iptables rules that cause the unwanted SNAT, and it won't delete those iptables rules until all pods that were created with the old ip-masq config have been removed. In other words, you may need to recreate all pods for a complete fix on upgraded deployments.

ktsakalozos:
@Cynerva @peix2 I updated the PR so it tries to load the br_netfileter module. If the modprobe fails we set the proxy-mode of kube-proxy to usermode.

Cynerva:
LGTM

PR #1153, opened 2020-04-27
Extend CA certs to last 10 years

joedborg:
Tested locally to check flags are valid.


ktsakalozos:
LGTM +1 thank you @joedborg

Issue #1150, opened 2020-04-27
root cert expired after a month, cluster does not respond anymore


ThomasSchoenbeck:
running `microk8s inspect` does not work as well as talking to the cluster. error is this: x509: certificate has expired or is not yet valid

How can i renew the root cert?

How can i make it last longer than a month?

balchua:
I've seen this issue happen when the date of the machine changed.  The certs duration is atleast 365 days.

ThomasSchoenbeck:
but than even after 365 days there needs to be a solution on how to update the root ca, right?

balchua:
There is a way to force generate the certs, by either changing the `csr.conf.template` say adding a DNS entry.
I have not personally tried but maybe cert manager can be used here.

ThomasSchoenbeck:
Since i currently are not able to use my previous installation of microk8s, i did a reinstall and checked the root ca on the new cluster using `openssl x509 -enddate -noout -in /var/snap/microk8s/current/certs/ca.crt`

the result: `notAfter=May 27 12:57:33 2020 GMT`

again only valid for one month

balchua:
You're right, I just checked the code the `CA` cert is not specified with the `-days` param when requesting for a cert. I think this defaults to 30 days.  
@ktsakalozos the `ca.crt` and the `front-proxy-ca.crt` is now 30 days old any particular reason why?
Thanks.

EzraBrooks:
Same happened to my system yesterday. Is there no way to regenerate both `ca.crt` and `server.crt` so I can get back in action?

balchua:
I have not tried this. 😊
Maybe we can try this:
* Stop microk8s. `microk8s.stop`
* Delete the `.crt` and `.key` in `/var/snap/microk8s/current/certs`.
* Modify the file `csr.conf.template` by adding DNS or comment.
* Start microk8s. `microk8s.start`
You may want to backup the content of `/var/snap/microk8s/current/certs` first.

ktsakalozos:
This in an oversight from our part (the missing -days arguments).

In the past hours we have patched the affected tracks, tested and released a new snap. This way any *new* deployments should not have this issue.

We will continue our work on a fix for the already existing deployments. The approach @balchua suggests seems promising but we will also need to recreate the kubeconfigs below line https://github.com/ubuntu/microk8s/blob/master/snap/hooks/install#L18

[...]

PR #1159, opened 2020-04-28
Refresh certs command

ktsakalozos:
This PR adds a refresh-certs command that would refresh the CA certs. The options given in refresh-certs are:
         -u, --undo          Revert the last refresh performed
         -c, --check         Check the expiration time of the installed CA
         -i, --install       Install new CA certificates

Mitigates https://github.com/ubuntu/microk8s/issues/1150


ktsakalozos:
Got confirmation that this approach works. I am merging it and we can refine the command at a later.

balchua:
Great!

PR #1164, opened 2020-04-29
Rename microk8s-info slot for juju

ktsakalozos:
This is was requested for the juju integration.


ktsakalozos:
@ycliuhw this is the porting of your patch to master.

PR #1168, opened 2020-04-30
dashboard-proxy command

joedborg:
Mostly brings a new command, `dashboard-proxy`, which scripts the enabling and exposure of the Kubernetes dashboard.

On top of this, 
- .venv should be ignored.
- trying to standardise on double quotes (is still a mix).
- `InstanceInfo` should return the IPv4 address too.

Initially a draft, to get the EXE for Windows testing.


joedborg:
Need to look at the delay, on slower machines, of the token becoming available after enabling the dashboard.  Currently fails on line 225, not sure why as I gate that with a check.  After checking the dashboard has come up, I can re run the command and it works fine.

joedborg:
image

installer/cli/microk8s.py
```diff
@@ -188,6 +191,64 @@ def uninstall() -> None:
...
+            raise provider_error
+
+    instance = vm_provider_class(echoer=echo)
+
+    echo.info("Checking if Dashboard is running.")
+    command = ["microk8s.enable", "dashboard"]
+    output = instance.run(command, hide_output=True)
+    if b"Addon dashboard is already enabled." not in output:
+        echo.info("Waiting for Dashboard to come up.")
+        command = ["microk8s.kubectl", "-n", "kube-system", "wait",
```
ktsakalozos, line 215: I go this error:
```
jackal@aurora:~/workspace/microk8s/installer$ ./dist/microk8s dashboard-proxy
Checking if Dashboard is running.
Waiting for Dashboard to come up.
error: timed out waiting for the condition on deployments/kubernetes-dashboard
An error occurred when trying to execute 'sudo microk8s.kubectl -n kube-system wait deployment kubernetes-dashboard --for condition=available' with 'multipass': returned exit code 1.
``` 

I think we should add a `--timeout=240s` or something large enough to catch the case of a dashboard that will not come up within the default 30s of the waiting time.

ktsakalozos: Added the timeout. Feel free to merge if you agree with it.

PR #1180, opened 2020-05-06
Ensure packet forwarding is enabled when transitioning to calico

ktsakalozos:
With this PR we make sure packets can be forwarded between nodes when forming the calico cluster.


Cynerva:
LGTM

PR #1182, opened 2020-05-07
Fix ctr using overlayfs on zfs

hpidcock:
Some ctr commands use a snapshotter locally. By default this is `overlayfs`, but when using zfs we need to use `native` (or `zfs`) snapshotter.

#1000 fixes this issue for containerd, but ctr doesn't use the containerd config.

Commands affected by this are:
- ctr containers create
- ctr images check
- ctr images import
- ctr images pull
- ctr run
- ctr snapshots

Snapshotter selection should probably be consolidated into one place.

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -172,6 +172,17 @@ arch() {
 }
 
 
+snapshotter() {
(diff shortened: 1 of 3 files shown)


ktsakalozos:
I did the consolidation of the snapshotter detection part.

Issue #896, opened 2020-01-07
Prometheus - 401 Unauthorized / Empty Dashboards


dannyyy:
I have a clean install of microk8s 1.17 on Ubuntu 18.04.

The following listing shows, which addins I have enabled:
```
dannyyy@k8s:~$ sudo microk8s.status
microk8s is running
addons:
cilium: disabled
dashboard: disabled
dns: enabled
fluentd: disabled
gpu: disabled
helm: disabled
ingress: disabled
istio: disabled
jaeger: disabled
juju: disabled
knative: disabled
kubeflow: disabled
linkerd: disabled
metallb: disabled
metrics-server: enabled
prometheus: enabled
rbac: enabled
registry: disabled
storage: enabled
```

I'm facing different issues regarding prometheus:
- Prometheus is not able to get the metrics from the kubelets: `401 Unauthorized`  
It's independent of having ´RBAC´ enabled or not.
![image](https://user-images.githubusercontent.com/3018863/71918319-56cce700-3182-11ea-81cf-45d0361da1e4.png)

- Different kind of RBAC issue.  
e.g. Pod `prometheus-k8s-0` is not able to scrape other namespaces because the appropriate right is missing. As an example the Helm chart of ´vernemq´ could be used:
```
# Default values for vernemq.
# This is a YAML-formatted file.
# Declare variables to be passed into your templates.

replicaCount: 1

image:
  repository: vernemq/vernemq
  tag: 1.10.1
  pullPolicy: IfNotPresent

nameOverride: ""
fullnameOverride: ""

serviceMonitor:
  create: true

service:
  # NodePort - Listen to a port on nodes and forward to the service.
  # ClusterIP - Listen on the service internal to the cluster only.
  # LoadBalancer - Create a LoadBalancer in the cloud provider and forward to the service.
  type: ClusterIP
#  clusterIP: 10.1.2.4
#  externalIPs: []
#  loadBalancerIP: 10.1.2.4
#  loadBalancerSourceRanges: []
  mqtt:
    enabled: true
    port: 1883
    # This is the port used by nodes to expose the service
    nodePort: 1883
  mqtts:
    enabled: false
    port: 8883
    # This is the port used by nodes to expose the service
    nodePort: 8883
  ws:
    enabled: false
    port: 8080
    # This is the port used by nodes to expose the service
    nodePort: 8080
  annotations: {}
  labels: {}

## VerneMQ resources requests and limits
[...]

PR #1195, opened 2020-05-13
Enable token authentication on kubelet

double73:
After enabling the Prometheus module, I checked the status of the targets, and the status of the kubelet endpoint appears to be Down. (401 Unauthorized)

Screen Shot 2020-05-13 at 1 35 27 PM

The token authentication appears to have been blocked with the --client-ca-file set up following the #323 modification.
https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-authentication-authorization/#kubelet-authentication

Adding the --authentication-token-webbook option to /var/snap/microk8s/current/args/kubelet changes the status to Up.
I think this is also related to Pod's CPU Usage not being displayed on the dashboard.

Related issue #896

So I hope you will review the PR and add it as a basic option if there is no problem.

### DIFF
--- a/microk8s-resources/default-args/kubelet
+++ b/microk8s-resources/default-args/kubelet
@@ -13,3 +13,4 @@
 --container-runtime-endpoint=${SNAP_COMMON}/run/containerd.sock
 --containerd=${SNAP_COMMON}/run/containerd.sock
 --node-labels="microk8s.io/cluster=true"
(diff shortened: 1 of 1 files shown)


ktsakalozos:
Thank you for this PR @double73 . I am closing this in favour of https://github.com/ubuntu/microk8s/pull/1202

PR #1198, opened 2020-05-15
Adding dashboard-proxy command to MicroK8s snap...

joedborg:
for parity with the Windows and MacOS wrappers.


ktsakalozos:
Thank you @joedborg 

@evilnick we should add this to the docs.

PR #1201, opened 2020-05-16
Upgrade metrics server 0.3.6

balchua:
Upgrade metrics-server to 0.3.6


microk8s-resources/actions/disable.metrics-server.sh
```diff
@@ -0,0 +1,30 @@
...
+$KUBECTL delete configmap -n kube-system metrics-server-config || true
+$KUBECTL delete deployment -n kube-system metrics-server-v0.2.1 || true 
+
+use_manifest metrics-server delete 
+# wait for pod to terminate before restarting kubelet
+echo "Waiting for pods to be terminated."
+sleep 10
+
+skip_opt_in_config "authentication-token-webhook" kubelet
+skip_opt_in_config "authorization-mode" kubelet
```
ktsakalozos, line 20: What you are doing here is correct. Sometimes we do not do a clean disable because we do not know if other addons/workloads are using the arguments the addon enabled. I think we should leave the kubelet as is.

microk8s-resources/actions/enable.metrics-server.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e
+
+source $SNAP/actions/common/utils.sh
+
+echo "Enabling Metrics-Server"
+
+"$SNAP/microk8s-enable.wrapper" rbac
```
ktsakalozos, line 9: Do we need RBAC to be enabled?

microk8s-resources/actions/enable.metrics-server.sh
```diff
@@ -0,0 +1,24 @@
...
+if [[ $apiserver == fail ]]
+then
+  echo "kube-apiserver did not start on time.  Proceeding."
+fi
+
+KUBECTL="$SNAP/kubectl --kubeconfig=${SNAP_DATA}/credentials/client.config"
+use_manifest metrics-server apply
+
+refresh_opt_in_config "authentication-token-webhook" "true" kubelet
+refresh_opt_in_config "authorization-mode" "Webhook" kubelet
```
ktsakalozos, line 21: Is webhook a requirement? I would expect it to work with the default AlwaysAllow

microk8s-resources/actions/disable.metrics-server.sh
```diff
@@ -0,0 +1,30 @@
...
+
+restart_service kubelet
+kubelet=$(wait_for_service kubelet)
```
ktsakalozos, line 23: I do not think we need to wait.

microk8s-resources/actions/metrics-server.yaml
```diff
@@ -209,28 +152,15 @@ subjects:
...
+- kind: User
+  name: 127.0.0.1
   apiGroup: rbac.authorization.k8s.io
+roleRef:
   kind: ClusterRole
-  name: admin
-subjects:
-- apiGroup: rbac.authorization.k8s.io
-  kind: User
-  name: front-proxy-client
```
ktsakalozos, line 236: I think we need to take into account the work on front proxy https://github.com/ubuntu/microk8s/pull/581

microk8s-resources/actions/disable.metrics-server.sh
balchua: At first i let it restart, but soon i noticed that the metrics server Pod is stuck in `Terminating` phase for quite some time.
Let me test this again.

microk8s-resources/actions/disable.metrics-server.sh
balchua: I'll update this one. Thanks.

microk8s-resources/actions/metrics-server.yaml
balchua: Done

microk8s-resources/actions/enable.metrics-server.sh
balchua: not really required.  I was confused with the metrics-server documentation. :)

microk8s-resources/actions/disable.metrics-server.sh
balchua: If there is no need to change the kubelet options, this can go away.

microk8s-resources/actions/enable.metrics-server.sh
balchua: Not really needed

balchua:
I have updated the PR considering the comments.  Thanks again for the review.

sstubbs:
Is there anything that I can do to help with this? I would really like to have metrics running properly.

balchua:
The PR is ready for a next review. So far my tests looks good.  Even tested it with rbac enabled.

sstubbs:
That is amazing I hope this can get merged.

PR #1202, opened 2020-05-18
Enable token authentication on kubelet

ktsakalozos:


balchua:
I added the same thing on the metric server upgrade PR.
But instead of keeping it static in the kubelet config, i was using the utils to and remove the option.
I should probably update the pr

ktsakalozos:
> I should probably update the pr

The reason for setting the default-args is to catch the case where in a cluster you add a new node. The new node will not know about the extra argument. I do not think it is a problem if the argument comes enabled by default.

Adding the argument as part of the enable script is to properly deploy in already existing clusters.

ktsakalozos:
Thank you for this PR @double73  and @balchua for the review.

PR #1208, opened 2020-05-20
Feature: HA

ktsakalozos:


ktsakalozos:
> `s/no/disabled/`?

Thank you for the suggestion @MichaelGooden. The reason we did not go for the disable/enable is because it may be confused with the `ha-cluster` addon. For example, you snap install microk8s and you check the status and you see high-availability reporting "no". If we had high-availability set to disabled users may be trying to "microk8s enable ha-cluster" not realizing that high-availability to become "enabled" needs a three node cluster.

PR #1209, opened 2020-05-20
Transition scripts from dqlite and calico to flannel and etcd

ktsakalozos:
This PR includes a set of scripts that would allow to transition a *single* node from a setup with dqlite+calico to etcd+flannel. This is needed when a node departs a cluster and needs to get back to its prior state. These scripts will be exercised when the user calls `microk8s disable ha-cluster`.

In case the transition to etcd+flannel fails we have a rollback script to bring the current state back.


ktsakalozos:
Thank you @joedborg

Issue #1197, opened 2020-05-15
APIserver complaining after dqlite node is removed


ktsakalozos:
@freeekanayaka after forming an HA cluster and removing one of the nodes the apiserver keeps showing the following exception: https://pastebin.ubuntu.com/p/Xm2s52Gzm5/

*How to reproduce:*
On three machines install MicroK8s from the `edge/ha-dqlite` channel and enable ha:
```
sudo snap install microk8s --classic --channel=edge/ha-preview
sudo microk8s enable ha-cluster
```
Join the three nodes with `sudo microk8s add-node` and `sudo microk8s join ....`.

Wait for all the nodes to become voters. Check that with:
```
sudo microk8s status ha-cluster
```
Remove one of the nodes with:
```
sudo microk8s leave
```
Dqlite now reports:
```
> microk8s.status ha-cluster                                                                                                                                     HA is enabled on this node but an HA cluster is has not formed yet.
Cluster nodes:
  - 172.31.24.191:19001 is a voter node
  - 172.31.25.112:19001 is a spare node
```
This does not align with the log line:
```
time="2020-05-15T04:21:59Z" level=warning msg="dqlite: servers: [{1 172.31.24.191:19001 voter} {1 172.31.25.112:19001 voter}] address=172.31.24.191:19001 attempt=2"
```

I would expect the cluster to keep its leader and keep working without being able to select a new leader. If this is not the case then that would mean we will need four nodes to ensure an HA setup?

Here are the inspection logs of the leader:
[inspection-report-20200515_042228.tar.gz](https://github.com/ubuntu/microk8s/files/4632436/inspection-report-20200515_042228.tar.gz)
and the removed node
[inspection-report-20200515_042339.tar.gz](https://github.com/ubuntu/microk8s/files/4632438/inspection-report-20200515_042339.tar.gz)
The node 172-31-30-18 is removed at around `May 15 04:18:39`

freeekanayaka:
Ok, I think I've found where the problem is. I'll work on a fix.

> This does not align with the log line:
> 
> ```
> time="2020-05-15T04:21:59Z" level=warning msg="dqlite: servers: [{1 172.31.24.191:19001 voter} {1 172.31.25.112:19001 voter}] address=172.31.24.191:19001 attempt=2"
> ```

This information is actually spurious, don't worry about it. It says "voter" but it's because that particular field is unused and doesn't get updated. It's mainly a bug in the logger, it should only mention the address and not the role.

[...]

PR #1210, opened 2020-05-20
Update the dqlite patch and add instructions on producing it

ktsakalozos:
This PR updates the dqlite patch with the fix for https://github.com/ubuntu/microk8s/issues/1197 and instructions for producing the patch.


ktsakalozos:
@joedborg  I am going to merge this since it is blocking the build. Please review at a later stage.

PR #1211, opened 2020-05-20
Eliminate references to the microk8s.upgrade command

ktsakalozos:
The plan is to go with an `microk8s.enable ha-cluster` addon and not an upgrade command.


ktsakalozos:
This does not need a review. this is "dead code" removal.

PR #1212, opened 2020-05-20
Improving the HA UX experience

ktsakalozos:
A few fixes in this PR:
- we make sure dqlites storage directory `${SNAP_DATA}/var/kubernetes/backend/ ` has the right permissions so commands like `microk8s.leave` can reset it.
- when joining a dqlite cluster you are asked to first enable ha "Please, retry after enabling HA on this joining node with 'microk8s enable ha-cluster'."
- stop poluting stdout/stderr from subprocess.check_call

### DIFF
--- a/scripts/cluster/join.py
+++ b/scripts/cluster/join.py
@@ -337,6 +337,7 @@ def reset_current_dqlite_installation():
 
     delete_dqlite_node(my_ep, other_ep)
 
+    print("Generating new cluster certificates.", flush=True)
     shutil.rmtree(cluster_dir, ignore_errors=True)
     os.mkdir(cluster_dir)
     if os.path.isfile("{}/cluster.crt".format(cluster_backup_dir)):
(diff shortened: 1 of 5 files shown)


scripts/cluster/join.py
```diff
@@ -338,34 +338,38 @@ def reset_current_dqlite_installation():
     if len(my_ep) > 0 and "127.0.0.1" not in my_ep[0]:
         for ep in other_ep:
             try:
-                subprocess.check_output("curl -X 'DELETE' https://{}/cluster/{} --cacert {} --key {} --cert {}  -k -s"
+                subprocess.check_output("curl -X DELETE https://{}/cluster/{} --cacert {} --key {} --cert {}  -k -s"
```
joedborg, line 341: Out of interest, why not use `requests`?

ktsakalozos: This comment is addressed in https://github.com/ubuntu/microk8s/pull/1220

Yes, requests looks better. We switched to that.

PR #1213, opened 2020-05-20
Port of NSIS installer

joedborg:
~~WIP to trigger and debug the Action build.~~

This moves the new NSIS based installer out of a private repo and into MicroK8s.  It also adds the framework to building the installer via GitHub actions.

Needless to say, this should be squashed when merged.

PR #1216, opened 2020-05-21
Feature/installer/catch no vm

joedborg:
Following on from https://github.com/ubuntu/microk8s/pull/1213

After testing on MacOS, seems like the VM check was broken.  Not what changed to broke it, but it's now fixed.


ktsakalozos:
LGTM +1

Issue #1223, opened 2020-05-25
Rebuilding snap always fails, must always do snapcraft clean folllowed by snapcraft --use-lxd


balchua:
Lately, I've been experiencing failure when rebuilding the snap.

This is the error I am getting.

```
Get:1 http://security.ubuntu.com/ubuntu xenial-security InRelease [109 kB]                                                                                                                                        
Hit http://archive.ubuntu.com/ubuntu xenial InRelease                                                                                                                                                             
Get:2 http://archive.ubuntu.com/ubuntu xenial-updates InRelease [109 kB]                                                                                                                                          
Get:3 http://archive.ubuntu.com/ubuntu xenial-backports InRelease [107 kB]                                                                                                                                        
Fetched 325 kB in 0s (0 B/s)                                                                                                                                                                                      
Copying needed target link from the system /root/parts/microk8s/install/go/src/github.com/kubernetes/kubernetes/_output/local/bin/linux/amd64
[Errno 21] Is a directory: '/root/parts/microk8s/install/go/src/github.com/kubernetes/kubernetes/_output/local/bin/linux/amd64'
We would appreciate it if you anonymously reported this issue.
No other data than the traceback and the version of snapcraft in use will be sent.
Would you like to send this error data? (Yes/No/Always/View) [no]: view
Traceback (most recent call last):
  File "/snap/snapcraft/4282/bin/snapcraft", line 11, in <module>
    load_entry_point('snapcraft==3.9.1', 'console_scripts', 'snapcraft')()
  File "/snap/snapcraft/4282/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/snap/snapcraft/4282/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/snap/snapcraft/4282/lib/python3.6/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/snap/snapcraft/4282/lib/python3.6/site-packages/snapcraft/cli/_command.py", line 88, in invoke
    return super().invoke(ctx)
  File "/snap/snapcraft/4282/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/snap/snapcraft/4282/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
[...]

PR #1232, opened 2020-05-26
clean up kubernetes _output directory 

balchua:
After it is copied to KUBE_SNAP_BINS, this fixes the issue https://github.com/ubuntu/microk8s/issues/1223


balchua:
@ktsakalozos i tried replacing the `rm`  with `unlink` it works when there is no modification in the source.  But once i modify the source, it starts showing the same error.  
Here are my steps:
1.  add the code to unlink the bin directory.
2.  I do a `snapcraft clean --use-lxd`, to make sure its all clean.
3.  Do a `snapcraft --use-lxd`.
4.  After a successful build, modify some codes.
5.  Run `snapcraft --use-lxd`

It will skip all other parts, and when it comes to the `microk8s` part, it starts complaining about this.

```shell
Copying needed target link from the system /root/parts/microk8s/install/go/src/github.com/kubernetes/kubernetes
[Errno 21] Is a directory: '/root/parts/microk8s/install/go/src/github.com/kubernetes/kubernetes'
```

Issue #1222, opened 2020-05-23
Upgrade Istio to 1.6 (major update)


tringuyen-yw:
Istio has announced a major update [Announcing Istio 1.6](https://istio.io/news/releases/1.6.x/announcing-1.6/) on 2020-05-21
Would appreciate if you can update microk8s accordingly.

balchua:
Using the `istioctl` to install makes it a lot easier.  Allow different profiles to set during addon enabling will be great.

tegk:
Feel free to have a look at my [PR](https://github.com/ubuntu/microk8s/pull/1233) to use v1.6.0

balchua:
Continuing the discussion here @tegk 

I do have some thoughts about using istioctl to install istio, just like what istio documentation says. The same thing with disabling the addon.
Instead of applying the demo profile we can have users set profiles while enabling the addon.
For example:
`microk8s enable istio:profile=default;addonComponents.grafana.enabled=true`

Where these `profile=default;addonComponents.grafana.enabled=true` are then parsed to become

`istioctl install --set profile=default --set addonComponents.grafana.enabled=true`
For reference on parsing arguments when enabling addons, check linkerd or metallb addons.

tegk:
> Continuing the discussion here @tegk
> 
> I do have some thoughts about using istioctl to install istio, just like what istio documentation says. The same thing with disabling the addon.
> Instead of applying the demo profile we can have users set profiles while enabling the addon.
> For example:
> `microk8s enable istio:profile=default;addonComponents.grafana.enabled=true`
> 
> Where these `profile=default;addonComponents.grafana.enabled=true` are then parsed to become
> 
> `istioctl install --set profile=default --set addonComponents.grafana.enabled=true`
> For reference on parsing arguments when enabling addons, check linkerd or metallb addons.

Sorry, I have created a separate issue for this because the discussion could diverge too much from the intention of this issue.

https://github.com/ubuntu/microk8s/issues/1235

tringuyen-yw:
> Feel free to have a look at my [PR](https://github.com/ubuntu/microk8s/pull/1233) to use v1.6.0

Can you show the step how to use this PR?

tegk:
You can build the snap by your self as described here: https://github.com/ubuntu/microk8s/blob/master/docs/build.md

If you are simply interested in trying Istio 1.6, I would just install it by my self with istioctl.

tringuyen-yw:
[...]

PR #1233, opened 2020-05-26
upgrade Istio add-on to v1.6.4

tegk:
Istio v1.6.0 is a major update that should be available to a broad audience via microk8s.

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -159,7 +159,7 @@ fi
 
 if ([ -f "$SNAP_USER_COMMON/istio-auth.lock" ] || [ -f "$SNAP_USER_COMMON/istio-auth.lock" ]) && ! [ -f "$SNAP_DATA/bin/istioctl" ]
 then
-  ISTIO_VERSION="v1.0.5"
+  ISTIO_VERSION="v1.6.0"
   echo "Fetching istioctl version $ISTIO_VERSION."
   ISTIO_ERSION=$(echo $ISTIO_VERSION | sed 's/v//g')
   mkdir -p "${SNAP_DATA}/tmp/istio"
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -159,7 +159,7 @@ fi
 
 if ([ -f "$SNAP_USER_COMMON/istio-auth.lock" ] || [ -f "$SNAP_USER_COMMON/istio-auth.lock" ]) && ! [ -f "$SNAP_DATA/bin/istioctl" ]
 then
-  ISTIO_VERSION="v1.6.0"
+  ISTIO_VERSION="v1.6.3"
   echo "Fetching istioctl version $ISTIO_VERSION."
   ISTIO_ERSION=$(echo $ISTIO_VERSION | sed 's/v//g')
   mkdir -p "${SNAP_DATA}/tmp/istio"
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -159,7 +159,7 @@ fi
 
 if ([ -f "$SNAP_USER_COMMON/istio-auth.lock" ] || [ -f "$SNAP_USER_COMMON/istio-auth.lock" ]) && ! [ -f "$SNAP_DATA/bin/istioctl" ]
 then
-  ISTIO_VERSION="v1.6.3"
+  ISTIO_VERSION="v1.0.5"
   echo "Fetching istioctl version $ISTIO_VERSION."
   ISTIO_ERSION=$(echo $ISTIO_VERSION | sed 's/v//g')
   mkdir -p "${SNAP_DATA}/tmp/istio"
--- a/microk8s-resources/actions/enable.istio.sh
+++ b/microk8s-resources/actions/enable.istio.sh
@@ -9,12 +9,13 @@ echo "Enabling Istio"
 
 if [ ! -f "${SNAP_DATA}/bin/istioctl" ]
 then
-  ISTIO_VERSION="v1.6.4"
+  ISTIO_VERSION="1.6.4"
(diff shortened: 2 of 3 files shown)


balchua:
Thanks for this PR @tegk.

 I do have some thoughts about using `istioctl` to install istio, just like what istio documentation says.  The same thing with disabling the addon.
Instead of applying the `demo` profile we can have users set profiles while enabling the addon.
For example:
`microk8s enable istio "profile=default:addonComponents.grafana.enabled=true"`

Where these `profile=default:addonComponents.grafana.enabled=true` are then parsed to become

`istioctl install --set profile=default --set addonComponents.grafana.enabled=true`

Just my 2 cents.

What do you think?

tegk:
> Thanks for this PR @tegk.
> 
> I do have some thoughts about using `istioctl` to install istio, just like what istio documentation says. The same thing with disabling the addon.
> Instead of applying the `demo` profile we can have users set profiles while enabling the addon.
> For example:
> `microk8s enable istio:profile=default;addonComponents.grafana.enabled=true`
> 
> Where these `profile=default;addonComponents.grafana.enabled=true` are then parsed to become
> 
> `istioctl install --set profile=default --set addonComponents.grafana.enabled=true`
> 
> Just my 2 cents.
> 
> What do you think?

@balchua Sounds like a good idea! Could you open an issue where we can discuss this?

balchua:
We can continue the discussion in the istio upgrade issue https://github.com/ubuntu/microk8s/issues/1222

balchua:
Some few pointers.

When transitioning from the old yamls to the new, we may have to make sure that we disable the old resources together with the new resources created.

Otherwise the old resources will stay there.

tringuyen-yw:
OK to update to [Istio 1.6.3](https://github.com/istio/istio/releases) released on 2020-06-18 ?

ktsakalozos:
@tringuyen-yw, sure. The latest the better.

@balchua @tegk, I guess you do not want me to wait for the istioctl based installation, right?

balchua:
@ktsakalozos im fine with the istio 1.6 PR to push through.  We can work on the `istioctl` based install for later versions.
@tegk what do you think?

tegk:
@tringuyen-yw,I have updated the version to 1.6.3
@ktsakalozos Please merge in case you do not see any problems as this would hugely beneficia

tringuyen-yw:
Thanks very much for taking care of the issue. Just curious, is it that simple to upgrade an add-on version?

tegk:
> Thanks very much for taking care of the issue. Just curious, is it that simple to upgrade an add-on version?

Yes, its trivial. Just change the version number in the files that were changed in this PR.

snap/hooks/configure
```diff
@@ -159,7 +159,7 @@ fi
 
 if ([ -f "$SNAP_USER_COMMON/istio-auth.lock" ] || [ -f "$SNAP_USER_COMMON/istio-auth.lock" ]) && ! [ -f "$SNAP_DATA/bin/istioctl" ]
 then
-  ISTIO_VERSION="v1.0.5"
```
ktsakalozos, line 162: Leave this to 1.0.5 please.

sstubbs:
Hi,

I'm not sure this information will help anyone but just adding it in case.

I've just installed istio manually as I want a minimal installation. The demo profile uses a lot of resources and I would rather not have two copies of prometheus running.

Here is what I've installed.

```
istioctl install \
--set profile=default \
--set values.grafana.enabled=false \
--set values.prometheus.enabled=false \
--set values.gateways.istio-egressgateway.autoscaleEnabled=false \
--set values.gateways.istio-ingressgateway.autoscaleEnabled=false \
--set values.pilot.autoscaleEnabled=false \
--set meshConfig.enablePrometheusMerge=true
```

And then I manually add the grafana charts to prometheus. When I asked the istio commutnity a few months ago I was told the integrated prometheus is not for production purposes and they now provide the charts for other grafana installs https://grafana.com/grafana/dashboards?search=istio.

tringuyen-yw:
@sstubbs 

Q1. can you confirm that by "installed istio manually" you meant installing istio from binaries, outside of microk8s, using the procedure described in https://istio.io/latest/docs/setup/getting-started/ ?

Q2. After your manual install of istio, what is the output of `microk8s status | grep istio`?

sstubbs:
Yes that is how I installed it.

Here are all the addons I enabled before hand for reference.
```
sudo snap install microk8s --classic --channel=latest/edge/ha-preview
microk8s enable ha-cluster dns rbac prometheus
```

When I run `microk8s status | grep istio` I get `istio: enabled`

tringuyen-yw:
`istio: enabled` even though you didn't explicitly enable the add-on via `microk8s enable istio` ???

I don't know how microk8s manages the "foreign" istio and could somehow recognize as its own add-on. I would like the opinion of microk8s experts on this. Thanks

sstubbs:
Yes it is strange. I did have another mirok8s snap installed previously but then I ran.
```
microk8s reset
sudo snap remove microk8s
```
Then I installed this. Could it be cached somewhere from that previous installation.

balchua:
Hi microk8s detect addons present is mostly be based on if a cluster resource is present or if the binary exist in certain locations, for example `${SNAP_DATA}/bin/helm`.
In the case of istio it checks for the pod name `istio`.
The list of addons are kept in this file `https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/addon-lists.yaml`.

tegk:
Istio 1.6.4 is out with security fixes that also affects the version that is currently used in Microk8s

tegk:
We need to use the new istioctl installing process so I am closing this until I have time to actually contribute this.

PR #1247, opened 2020-05-28
Add a pytest for clustering

ktsakalozos:
This will test HA and we can extend it in the future.

We will need to have it called by `test-distro.sh` or call it directly from a Jenkins job.


ktsakalozos:
Thank you @joedborg

Issue #1241, opened 2020-05-27
Python addons should be able to use the python extension


camille-rodriguez:
This line prevents the use of python extension on files for addons. For example, enable.kubeflow.sh is a python addon, but uses the bash extension because of how the wrapper is configured. An option would be to look for addons with either .sh or .py extensions.  https://github.com/ubuntu/microk8s/blob/c256d7a007e2d5a6620196e778839d4dc9499064/microk8s-resources/wrappers/microk8s-enable.wrapper#L38

ktsakalozos:
Thank you for spotting this and suggesting a fix @camille-rodriguez. We could even have a .* to support any extension.

Would you (or anyone else) be interested in a PR?

camille-rodriguez:
@ktsakalozos Happy to help, I will look into it. I'm no expert in bash scripting, but I doubt something as simple as * exists in bash. We'll probably have to loop over the actions to find the correct one, since the following line needs the full path to launch the wrapper.

joedborg:
If it's just py and sh, we could do `.{py,sh}`.  Depends what we want to allow I guess.

joedborg:
Or do a similar to what I had to do with dashboard-proxy, where we have a shell wrapper around Python (useful for injecting the right environment) https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/microk8s-dashboard-proxy.wrapper

camille-rodriguez:
I submitted a PR, let me know what you think @joedborg @ktsakalozos ! :)

anushkrishnav:
I am intrested

krisstern:
Is anyone working on this issue at the moment?

krisstern:
Hi @camille-rodriguez Is it okay for me to take the modifications you have tried to make in #1248 and submit another patch to follow up on this issue? I could submit a PR soon.

ktsakalozos:
@krisstern you can take over this PR. Thank you.

krisstern:
Thanks @ktsakalozos! I will work on a PR for the issue then.

krisstern:
So was able to test install `microk8s` without the suggested changes first using LXD according to https://github.com/ubuntu/microk8s/blob/master/docs/build.md#building-the-snap-using-lxd on my ubuntu for arm64, but seems `multipass` is not happy with this architecture. So will need to take extra time to retry everything using my ubuntu for intel setup, otherwise seems like not an ideal situation.

krisstern:
[...]

PR #1248, opened 2020-05-28
Addon extension

camille-rodriguez:
This is a PR to fix https://github.com/ubuntu/microk8s/issues/1241.

The microk8s wrappers to enable and disable addons expect only .sh extensions. However, new addons are being written in python as well. This PR is to avoid restricting in which code the addons must be coded.


camille-rodriguez:
> LGTM. Should we fix the enable.kubeflow.sh extension in this PR as well, or push that out to another one?

I can add that quickly and let Travis tell us if everything works as expected.

microk8s-resources/wrappers/microk8s-enable.wrapper
```diff
@@ -35,17 +35,21 @@ for addon in "$@"; do
...
+        break
+      else
+        echo "Failed to enable $action"
+        exit 1
+      fi
     fi
-  elif [ -f "${SNAP}/actions/$action.yaml" ]; then
+  done
+
+  if [ -f "${SNAP}/actions/$action.yaml" ]; then
```
ktsakalozos, line 52: This is a difference in behavior.

If we have both `enable.foo.sh` and `foo.yaml` files the `foo.yaml` in skipped. With this PR we run the script and then we also apply the manifest. This is not what we want.

Consider the case on the ingress addon. There is an ingress.yaml that serves as a template, and the enable.ingress.sh parses the template yaml and applies it. We do not want to apply the template yaml.

camille-rodriguez: I see how that is changed. However, I have trouble understanding in which case the yaml would be applied then. In the ingress example, the code will always find first the enable.ingress action, so it would never run the yaml manifest. When is this actually used?

camille-rodriguez: Ok, so from the whole list, only metrics-server.yaml does not have an "enable.metrics-server.sh" equivalent, so that would be the only use case so far for that chunk of code

camille-rodriguez: New PR pushed, it should resolve this.

krisstern:
A follow-up PR #2795 has been submitted to address the previous patches attempted in this PR.

PR #1259, opened 2020-06-01
No not inspect etcd and flannel when in HA mode

ktsakalozos:
Flannel and etcd are not running in HA, they should not be reported as error.


ktsakalozos:
> What if flanneld is enabled via addons but not HA?

What do you mean? There is not addon enabling flannel.

joedborg:
> > What if flanneld is enabled via addons but not HA?
> 
> What do you mean? There is no addon enabling flannel.

Ignore me, it's Monday morning.  I was thinking of Fluentd.

PR #1263, opened 2020-06-02
Verify the cert of the master node to join

ktsakalozos:
We want to verify the identity of the master node (the node we are joining to) by checking its certificate fingerprint.

Proposed changes:
- The connection token is extended with the sha256sum of the certificate fingerprint
- The joining node can skip the certificate check with the `--skip-verify` flag


scripts/cluster/join.py
```diff
@@ -36,6 +38,23 @@
...
+    :param port: the port of the server
+    :return: the md5sum of the server cert
+    """
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.settimeout(1)
+    wrapped_socket = ssl.wrap_socket(sock)
+    wrapped_socket.connect((addr, port))
+    der_cert_bin = wrapped_socket.getpeercert(True)
+    wrapped_socket.close()
+    return hashlib.md5(der_cert_bin).hexdigest()
```
johnsca, line 55: As discussed online, MD5 is insecure.  And SHA256 is only 65 chars vs MD5's 33, so it definitely seems worth switching to SHA256.

ktsakalozos: +1 doing it now. Thank you

ktsakalozos:
@jdstrand @joemcmanus could we have a review of this PR from anyone from the security team?

Thank you.

balchua:
This may break the `microk8s add --token <random token>`.

scripts/cluster/join.py
```diff
@@ -730,6 +752,10 @@ def join_dqlite(connection_parts):
     master_ep = connection_parts[0].split(":")
     master_ip = master_ep[0]
     master_port = master_ep[1]
+
+    if len(connection_parts) > 2 and verify:
```
ktsakalozos, line 756: @balchua if you specify a token that does not include a '/' it would be as if there is no sha to verify the connection and the `verify_server` will not be called.

balchua: Im afraid the message printed out during `add-node` maybe wrong if token is passed.  I think a check is needed prior to executing `openssl` and verify if `token` is passed in as argument.

setharnold, line 756: Spitting on `:` to select IP and port in this fashion probably prevents IPv6 addresses from being used.

scripts/cluster/join.py
```diff
@@ -36,6 +38,23 @@
...
 
+def get_fingerprint(addr, port):
+    """
+    Get the certificate fingerprint of the server at addr:port
+
+    :param addr: the address of the server
+    :param port: the port of the server
+    :return: the digest of the server cert
+    """
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
setharnold, line 49: This appears to be creating a second, short-lived, connection, solely for the purpose of getting a peer certificate and checking that the checksum of the certificate matches a fingerprint supplied from another source.

If so, this introduces a race condition -- the connection that would actually be used by the rest of the application has no relationship the certificate that was retrieved in this step, they come from different connections.

I can understand that the deployment model of microk8s may not allow for *actual* live certificates to be used, so a method to identify peers by a fingerprint alone may be useful, but it at least needs to verify the certificate used on the connection in question.

scripts/cluster/join.py
```diff
@@ -36,6 +38,23 @@
...
+    sock.settimeout(1)
+    wrapped_socket = ssl.wrap_socket(sock)
+    wrapped_socket.connect((addr, port))
+    der_cert_bin = wrapped_socket.getpeercert(True)
+    wrapped_socket.close()
+    return hashlib.sha256(der_cert_bin).hexdigest()
```
setharnold, line 55: If the peer doesn't supply a certificate, this step will fail with exception `TypeError: object supporting the buffer API required` -- please make sure this exception either terminates the application or otherwise prevents the connection to untrusted sources.

scripts/cluster/join.py
ktsakalozos: You are right, we have not looked into IPv6 support yet. Supporting IPv6 will probably need to touch other places as well. I am making a note to revisit this part when we start working on IPv6.

scripts/cluster/join.py
ktsakalozos: If anything goes wrong with the`get_fingerprint` we catch the top level `Exception` and exit gracefully, see https://github.com/ubuntu/microk8s/pull/1263/files#diff-7ffd8f66475be19b5114657ee59733edR789 . I believe this is the behavior you expect.

scripts/cluster/join.py
ktsakalozos: You are correct, we create a short-lived connection just to get the certificate.

Python's `requests` package does not allow getting the certificate details. The work we need to do to is described in https://stackoverflow.com/questions/16903528/how-to-get-response-ssl-certificate-from-requests-in-python/52072170#52072170 

I am looking for alternatives before commit to that approach.

scripts/cluster/join.py
johnsca: I'd recommend using `urlparse` regardless.

scripts/cluster/join.py
johnsca: Discussed in chat, but adding here for reference...

It looks like if you use `streaming=True` as [documented here](https://requests.readthedocs.io/en/master/user/advanced/#body-content-workflow), you can get access to the socket being used by requests and call `getpeercert` on it:

```python
>>> import requests
>>> with requests.get('https://google.com/', stream=True) as resp:
...     print(resp.raw.connection.sock.getpeercert(True))
... 
b'0\x82\x04\xc50\x82\x03\xad\xa0\x03\x02\x01\x02...'
```

ktsakalozos: I am struggling with this and would appreciate any pointers you may be able to offer.

We need to use the same connection to do two things. 1. do the SSL handshake and inspect the host certificate 2. If the certificate fingerprint is acceptable perform a POST request. The closest  I got to this is flow is to keep the tcp connection and perform two HTTP requests similar to [1]. The obstacle I am now facing is that I am asking the connection to not verify the certificates [2] and (I think) this restrict me from getting the SSLsocket form wrap_socket [3] with the error "ssl.SSLError: [SSL: UNKNOWN_PROTOCOL] unknown protocol". Further more when wrapping the socket I suspect the socket gets detached and a new handshake takes place [4].

Any hints you have to offer are much appreciated.

[1] https://stackoverflow.com/questions/2009243/how-to-speed-up-pythons-urllib2-when-doing-multiple-requests
[2] https://docs.python.org/2/library/httplib.html#httplib.HTTPSConnection
[3] https://docs.python.org/3/library/ssl.html#ssl.SSLContext.wrap_socket
[4] https://github.com/python/cpython/blob/3.8/Lib/ssl.py#L1006

scripts/cluster/join.py
ktsakalozos: @balchua the message produced if a user provided token is requested looks like this:
```
root@u1:~# microk8s.add-node --token 12345678901234567890123456789012
From the node you wish to join to this cluster, run the following:
microk8s join 10.228.72.149:25000/12345678901234567890123456789012/d8d3e9e758d011b0b180122d55173cb3ebf426d4332cb9bea5ba0e464560280e

If the node you are adding is not reachable through the default interface you can use one of the following:
 microk8s join 10.1.215.192:25000/12345678901234567890123456789012/d8d3e9e758d011b0b180122d55173cb3ebf426d4332cb9bea5ba0e464560280e
 microk8s join 10.1.100.0:25000/12345678901234567890123456789012/d8d3e9e758d011b0b180122d55173cb3ebf426d4332cb9bea5ba0e464560280e
 microk8s join 10.228.72.149:25000/12345678901234567890123456789012/d8d3e9e758d011b0b180122d55173cb3ebf426d4332cb9bea5ba0e464560280e
```

This looks correct to me as next to the token we add the certificate token. However, you are right the connection string is not only the token provided. We will be able to still join if we use the connection string without the server certificate fingerprint, ie:
```
microk8s join 10.228.72.149:25000/12345678901234567890123456789012
```
Alternatively there is the `--skip-verify` flag
```
microk8s join 10.228.72.149:25000/12345678901234567890123456789012/d8d3e9e758d011b0b180122d55173cb3ebf426d4332cb9bea5ba0e464560280e --skip-verify
```

Would this be ok with you?

microk8s-resources/wrappers/microk8s-add-node.wrapper
```diff
@@ -65,12 +65,14 @@ fi
 default_ip="$(get_default_ip)"
 all_ips="$(get_ips)"
 
-echo "Join node with: microk8s join ${default_ip}:$port/${token}"
+check=$(openssl x509  -md5 -in "$SNAP_DATA"/certs/server.crt -outform der | sha256sum | cut -d' ' -f1)
```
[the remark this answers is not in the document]
ktsakalozos: You are right, `-md5` is removed now.

scripts/cluster/join.py
ktsakalozos: Spent some more time digging into this. We now use `http.client` manuly force the SSL handshake with connect() and if the certificate checks out we do the POST request.

scripts/cluster/join.py
ktsakalozos: Because this is not a url urlparse will not parse it correctly.
```
>>> urlparse('127.0.0.1:2500')
ParseResult(scheme='', netloc='', path='127.0.0.1:2500', params='', query='', fragment='')
```

joedborg (approved):

PR #1268, opened 2020-06-04
Force passthrough of silent install to multipass

joedborg:
User feedback shows that silent isn't passed through bundled installers.


joedborg:
Tested on laptop.  We should look at testing this under automation, but would need bare metal.

ktsakalozos:
LGTM +1

PR #1271, opened 2020-06-05
Build pre-stable releases with its own dqlite patch

ktsakalozos:


ktsakalozos:
Merging this so to see our CI is OK too. @joedborg please review when you have time.

Issue #1272, opened 2020-06-05
Grafana reports "Panel plugin not found: grafana-piechart-panel" on several dashboards


nikolayvinogradov:
Installed microk8s and enabled dns, storage and prometheus add-ons. Connected to grafana via port-forward from external kubectl (localhost:3000). 

On few dashboards (see below) grafana shows red labels with "Panel plugin not found: grafana-piechart-panel" instead of pie charts.

    $ kubectl get cm -n monitoring  | awk '{ print $1; }' | tail -n +2 | while read cm; do ( kubectl describe cm -n monitoring $cm | grep pie ) && echo $cm; done
                "pieType": "donut",
                "type": "grafana-piechart-panel",
                "pieType": "donut",
                "type": "grafana-piechart-panel",
                        "pieType": "donut",
                        "type": "grafana-piechart-panel",
                        "pieType": "donut",
                        "type": "grafana-piechart-panel",
    grafana-dashboard-cluster-total
                "pieType": "donut",
                "type": "grafana-piechart-panel",
                "pieType": "donut",
                "type": "grafana-piechart-panel",
                        "pieType": "donut",
                        "type": "grafana-piechart-panel",
                        "pieType": "donut",
                        "type": "grafana-piechart-panel",
    grafana-dashboard-namespace-by-workload
                "pieType": "donut",
                "type": "grafana-piechart-panel",
                "pieType": "donut",
                "type": "grafana-piechart-panel",
                        "pieType": "donut",
                        "type": "grafana-piechart-panel",
                        "pieType": "donut",
                        "type": "grafana-piechart-panel",
    grafana-dashboard-workload-total

![Screenshot_20200605_151257](https://user-images.githubusercontent.com/32940265/83875019-15476480-a73f-11ea-9fe2-d865fb6090da.png)

PR #1274, opened 2020-06-05
Add pie chart to the grafana of prometheus operator

ktsakalozos:
Fixes: https://github.com/ubuntu/microk8s/issues/1272

### DIFF
--- a/microk8s-resources/actions/prometheus/README.md
+++ b/microk8s-resources/actions/prometheus/README.md
@@ -11,6 +11,12 @@ Then update the deployments of prometheus and alertmanager to have one replica o
 
 There is no reason to automate this because this may not work in the future. This is not the recommended way of installing Prometheus Operator.
 
+Add in the grafana deployment in the grafana container the environment variable:
+```dtd
+      - env:
+        - name: GF_INSTALL_PLUGINS
+          value: grafana-piechart-panel
+```
 
 # References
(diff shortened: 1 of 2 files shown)


ktsakalozos:
Have a look @nikolayvinogradov, this seems to be working.

ktsakalozos:
This approach will cause the grafana pod to crashloop when dns is not available and grafana.com is not resolvable.
```
jackal@aurora:~$ microk8s.kubectl logs -n monitoring   pod/grafana-5976758c65-fzmb6

Error: ✗ Failed to send request: Get https://grafana.com/api/plugins/repo/grafana-piechart-panel: dial tcp: lookup grafana.com on 127.0.0.53:53: read udp 127.0.0.1:56396->127.0.0.53:53: read: connection refused

NAME:
   Grafana cli plugins install - install <plugin id> <plugin version (optional)>

USAGE:
   Grafana cli plugins install [arguments...]
``` 

 Would it be better if we just not use the piechart?

ktsakalozos:
@nikolayvinogradov, I will close this PR and not move forward with it. The reason is that this PR places the requirement pods can reach grafana.com which in general is OK but it will break the deployment in network restricted environments.

We have to find another way to address the issue. We could either:
- document a post deployment step to install the missing addon, or
- patch the manifests to not use the pie chart

What would best suit your needs?

Issue #1194, opened 2020-05-13
Multus does not create a second interface in POD


ednaganon:
I would like to have a second flannel interface  in a pod using Multus. I have few questions
1. First try - since Flannel exists as the default network, i create only multus-daemonset.yml and flannel networkAattachmentDefinition network (but not flannel daemonset). Creating a test pod with the annotation - only eth0 created:

$ **kubectl -n=kube-system get po -o wide**
[sudo] password for eganon:
NAME                                              READY   STATUS    RESTARTS   AGE    IP            NODE          NOMINATED NODE   READINESS GATES
coredns-588fd544bf-sqmh5                          1/1     Running   2          17h    10.1.8.15     gr-05         <none>           <none>
dashboard-metrics-scraper-db65b9c6f-zb6sb         1/1     Running   2          17h    10.1.8.16     gr-05         <none>           <none>
heapster-v1.5.2-58fdbb6f4d-6jrm5                  4/4     Running   11         17h    10.1.8.14     gr-05         <none>           <none>
kube-multus-ds-amd64-9qvz9                        1/1     Running   0          172m   192.168.1.6   192.168.1.6   <none>           <none>
kube-multus-ds-amd64-dv8sl                        1/1     Running   1          172m   192.168.1.5   gr-05         <none>           <none>
kube-multus-ds-amd64-lpk7b                        1/1     Running   0          172m   192.168.1.8   192.168.1.8   <none>           <none>
kubernetes-dashboard-67765b55f5-p22jj             1/1     Running   2          17h    10.1.8.17     gr-05         <none>           <none>
monitoring-influxdb-grafana-v4-6dc675bf8c-hb95v   2/2     Running   4          17h    10.1.8.13     gr-05         <none>           <none>

**$ kubectl -n=kube-system get net-attach-def**
NAME      AGE
flannel   10h

**$ kubectl -n=kube-system describe net-attach-def**
Name:         flannel
Namespace:    kube-system
Labels:       <none>
Annotations:  API Version:  k8s.cni.cncf.io/v1
Kind:         NetworkAttachmentDefinition
Metadata:
  Creation Timestamp:  2020-05-12T21:56:27Z
  Generation:          1
  Managed Fields:
    API Version:  k8s.cni.cncf.io/v1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .:
          f:kubectl.kubernetes.io/last-applied-configuration:
      f:spec:
        .:
[...]

Issue #867, opened 2019-12-15
How to change cni? I want to use macvlan or multus


markus7811:
Hi,

i am new to microk8s but think it is really good for running some iot workload or homeautomation and things like this. Therefor i need to use macvlan.

At the moment it is really hard to get an idea what i have to do to change the default cni (flannel) to something other, like multus (would be my favorit) or macvlan. Please consider to add some information what needs to be done to change the cni.

Thank you a lot.

balchua:
Hi @markus7811 thank you for pointing out the lack of documentation on this front.  Apologies.

To get you started, cni binaries are usually stored in `$SNAP_DATA/bin`, in the case of cilium cni, it is actually in `$SNAP_DATA/opt/cni/bin`

On the cni configuration aspect, again it is stored in `$SNAP_DATA/args/cni-network/`

If you need to change kubelet's configuration, it is located in `$SNAP_DATA/args/kubelet`

If you need to change containerd configuration, you will need to change in `$SNAP_DATA/args/containerd-template.toml`.  please take note that you are modifying the template file and not the actual `containerd.toml`

`$SNAP_DATA` is often located in `/var/snap/microk8s/current` directory.

balchua:
Since multus is now an addon on edge channel, closing the issue.

PR #1276, opened 2020-06-05
Add new add-on for Multus CNI

apnar:
This PR creates a new add-on that installs the Multus CNI plugin.  Multus CNI enables attaching multiple network interfaces to pods in Kubernetes (details here: https://github.com/intel/multus-cni).  The install is transparent and the current CNI plugin is carried forward as the default in Multus.  With Multus installed users can easily add multiple interfaces to pods using varying CNI plugins (like bridge, macvlan, ipvlan, ptp, etc.) all of which wasn't possible previously in microk8s without some hacking. 

I've sorted out enable/disable compatibility with the cilium (the only other CNI add-on) so either one or both may be enabled simultaneously.  It may make sense to make a more structural changes to store CNI plugins in SNAP_DATA in general, but I'll save that for a later PR.

Multus supports installation via their pod or manually, I've opted for the manual install approach in this PR so there is no pod that runs or remains running.  Installation defaults to the latest version at the time of enabling but supports the MULTUS_VERSION env variable if a specific version is required.

I've provided a test case that creates a pod with multiple interfaces but I don't know how the test the test so it may need a closer look.

Open to any suggestions for recommendations for improvement.

### DIFF
--- a/microk8s-resources/actions/disable.cilium.sh
+++ b/microk8s-resources/actions/disable.cilium.sh
@@ -37,15 +37,6 @@ fi
 
 set_service_expected_to_start flanneld
 
-echo "Restarting kubelet"
-refresh_opt_in_config "cni-bin-dir" "\${SNAP}/opt/cni/bin/" kubelet
-run_with_sudo preserve_env snapctl restart "${SNAP_NAME}.daemon-kubelet"
-echo "Restarting containerd"
-if ! grep -qE "bin_dir.*SNAP}\/" $SNAP_DATA/args/containerd-template.toml; then
-  run_with_sudo "${SNAP}/bin/sed" -i 's;bin_dir = "${SNAP_DATA}/opt;bin_dir = "${SNAP}/opt;g' "$SNAP_DATA/args/containerd-template.toml"
-fi
-run_with_sudo preserve_env snapctl restart "${SNAP_NAME}.daemon-containerd"
-
 echo "Restarting flanneld"
 run_with_sudo preserve_env snapctl stop "${SNAP_NAME}.daemon-flanneld"
 
--- a/microk8s-resources/actions/disable.multus.sh
+++ b/microk8s-resources/actions/disable.multus.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+set -e
+
+source "${SNAP}/actions/common/utils.sh"
+
+KUBECTL="${SNAP}/kubectl --kubeconfig=${SNAP_DATA}/credentials/client.config"
+
+if [ -f "${SNAP_DATA}/args/cni-network/00-multus.conf" ]
+then
+  echo "Disabling Multus"
+  echo "Checking if multus daemonset is running"
+  DS_CHECK=$(${KUBECTL} get pods -n kube-system --selector=app=multus -o name)
+  if [ -n "${DS_CHECK}" ]; then
+    echo "Found multus deamonset, removing it and cleaning up OS files"
+    cat "${SNAP}/actions/multus.yaml" | ${SNAP}/bin/sed "s#{{SNAP_DATA}}#${SNAP_DATA}#g" | ${KUBECTL} delete -f - 
+    run_with_sudo rm -f "${SNAP_DATA}/args/cni-network/00-multus.conf"
+    run_with_sudo rm -rf "${SNAP_DATA}/args/cni-network/multus.d"
+    run_with_sudo rm -f "${SNAP_DATA}/opt/cni/bin/multus"
+
+    echo -n "Waiting for multus daemonset to terminate."
+    while [ -n "${DS_CHECK}" ]; do
+      sleep 3
+      DS_CHECK=$(${KUBECTL} get pods -n kube-system --selector=app=multus -o name)
+      echo -n "."
+    done 
+    echo
+    echo "Multus daemonset terminated."
+    echo "Initiating removal on all other nodes."
+    nodes_addon multus disable
+  else
+    echo "Daemonset not found so we are likely a node, cleaning up OS files"
+    run_with_sudo rm -f "${SNAP_DATA}/args/cni-network/00-multus.conf"
+    run_with_sudo rm -rf "${SNAP_DATA}/args/cni-network/multus.d"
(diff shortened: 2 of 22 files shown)


apnar:
I should add that I had to make some changes to the cilium add-on enable/disable so they play well together.  I also improved the clean up a little for the cilium disable.

apnar:
And lastly it should close #867 and close #1194

balchua:
This looks cool.  Don't know much about multus though. 😁

Does it work along with the default `flannel` cni?

apnar:
Yes!  When you install it nothing changes, everything will still use the default Flannel (or cilium if that’s installed). You can then add annotations to specific pods to add additional network interfaces using the same or different CNIs. You can look at the 2 yaml files in the test/template for of this PR for examples of how it works.

apnar:
Rebased to fix recent merge conflicts and cleaned up linting errors.

Is there a way I can trigger travis to do a build test without UNDER_TIME_PRESSURE being true?  Just so I can confirm my test cases work correctly.

ktsakalozos:
Thank you for staying on top of the recent changes.

To have travis run specific tests you will probably need to edit .travis and mark the PR as WIP so we do not review it and merge it.

apnar:
Thanks for the review!  I went ahead and removed the wrapper as it was just informational.  I moved the info so it's displayed when the plugin is enabled instead.  Changes are now reflected here.

ktsakalozos:
Thank you for this PR. It looks great for a single node. What would happen in a cluster setup?

Some addons may need to configure and restart services on all nodes and in some cases they may need to enable addons (or even themselves) on all nodes. For this purpose we have a set of distributed operations [1] paired with respective calls in common/utils. 

In clusters things may get more complicated if you consider the case where a node is added to a cluster well after the addon got enabled. When joining a node we copy the kubelet arguments from the master node [2] but there may be other information that we may need to carry over. Having  multus as a daemonset may be more convenient for this case because k8s would take care of running the daemon in all (new and old) nodes.

Finally, you may have seen that we will be soon shipping HA [3]. It may worth making sure that your patch works for the HA branch [4].

Really appreciate your work. Thank you.

[1] https://github.com/ubuntu/microk8s/blob/master/scripts/cluster/distributed_op.py
[2] https://github.com/ubuntu/microk8s/blob/master/scripts/cluster/join.py#L834
[3] https://github.com/ubuntu/microk8s/issues/1127
[4] https://github.com/ubuntu/microk8s/tree/feature/ha-enable

apnar:
Thanks for the comments.  I hadn't considered the cluster use cases, here are a few thoughts based on my understanding:

- Multus itself is stateless so there should be no additional considerations for a HA cluster vs a regular cluster.

- When enabled on all members of the cluster it should work correctly in its current state.

- If enabled on some members of the cluster but not others there would be issues in its current state.

- As with the cilium add-on there are local changes that are required that are outside what a daemonset would handle, in particular changing the location of the cni-bin-dir from $SNAP to $SNAP_DATA with the associated restarts and making the other CNI plugins available in that new location.  That may theoretically be possible to do via a daemon set but it certainly isn't handled by the one developed by the multus team and I doubt you'd want to add and maintain one here.  The same is true for the cilium add-on so I don't see how it could work in a cluster without its enable script being run on every node.

- The above could possibly be mitigated by a more structural change or two, like shifting the default cin-bin-dir to always reside in $SNAP_DATA so it's in a writable location.  CNI plugins could then either be links back to $SNAP (which is how I handle it in this add-on) or just put them there directly at install time.  That'd also simplify any other CNI add-ons that you might want to add in the future.  Would you guys be open to a PR implementing this?

- I don't think it's an unreasonable burden to tell a user that if they want to use an add-on it must be enabled on all nodes joining a cluster.  I think it'd be fine as documentation but it could also be programmatically verified at the time of node joining.

- For existing clusters I do think the enable/disable should be effective cluster wide though as that'd be the expectation of the user.  For that the distributed operations you mentioned would be a good option.  I don't see anything like this in the cilium add-on.  Do you know of any other add-ons that are leveraging that that I could look at as an example?

TLDR, this add-on supports clustering as well as the other CNI plugin (cilium) does.

apnar:
I thought about this some more and think the best approach is to move the default CNI bin dir to SNAP_DATA directory.  That'll allow the multus add-on to much more easily be deployed across a cluster using native daemonsets.  It also simplifies things and removes a number of other tweaks required when installing other add-ons.

I've moved this to draft and I'll submit a separate PR to move the default CNI bin dir for you to look at.  If that looks good I'll rewrite this to move to a deamonset approach.

apnar:
@ktsakalozos, this is in decent enough form I think it's ready for you to take a look at.  Worked with @sergiomanso to change approach to using a daemonset.  This also moves the default CNI bin directory to SNAP_DATA and takes care of cleaning up all the various places that changed it there and back in all the other add-ons.

It applies cleanly to both master and the ha-enable branch and should be cluster friendly for enabling, usage, and disabling.

You'd mentioned on slack potentially needing to rebaseline, happy to do it just let me know which branch/change you'd like to baseline off of.

ktsakalozos:
Thank you for this work @apnar,  @sergiomanso. This work will be available on the `latest/edge` channel within a day: `sudo snap install microk8s --channel=latest/edge --classic`.

We need two more things to complete the multus work:
- a docs page for this addon describing how it should be used
- an announcement/request for testing in https://discuss.kubernetes.io/

evilnick:
Documentation is in the upstream discourse. I have made a wiki post for Multus, <https://discuss.kubernetes.io/t/add-on-multus/11664> 
Just add any details to that and I will add links when it is ready to go.

evilnick:
@apnar if you can't edit that wiki post (you need to have a certain amount of trust on the kubernetes discourse) you can either reply to it with the details or just link me to some notes here and I will write up the docs.

apnar:
@evilnick thanks, I posted a reply on discourse with some basics on the add-on and how to use it.  Let me know if you think it needs more/less etc.

@ktsakalozos thanks for merging!  and thanks for the assistance.  Do you have a suggestion on the proper spot to post an announcement/request for testing?  I didn't see any other add-on posts in the general Announcement category, should I stick it in "General Discussion" & "microk8s"?

PR #1295, opened 2020-06-11
Import container images from sideloaded snap

joedborg:
PR to code a method, primarily for IoT devices, to deliver containerd images in an offline environment of MicroK8s.  It’s expected that this would be part of the IoT device provisioning process.


joedborg:
Keep open.

PR #1299, opened 2020-06-12
Update nvidia packages

ktsakalozos:
This makes gpu work out of the box at least on ubuntu 20.04


ktsakalozos:
Thank you @mar-kolya for your work in this.

PR #1302, opened 2020-06-12
Testing the containerd update

ktsakalozos:
Updating containerd to 1.3.4 and runc to 1.0.0-rc10


microk8s-resources/actions/common/containerd-config-migrate.py
```diff
@@ -0,0 +1,71 @@
+import shutil
```
ktsakalozos, line 1: @balchua, is the migration needed? Containerd says it is backwards compatible with the previous version, does this include the config it runs with? I think It runs ok with the old config.

balchua: I couldn't get the gpu working with the old config.

ktsakalozos: I think gpu was not working because we did not change the runtime in the initial commit.

I did some testing and it seems fine to me now. Let me know what you think.

balchua: You mean the `$RUNTIME`?

ktsakalozos: > You mean the `$RUNTIME`?

Yes

joedborg:
Keep in mind that 1.4.0 is close too.

ktsakalozos:
> Keep in mind that 1.4.0 is close too.

Indeed 1.4.0 is about to be released. Based on the upgrade path in [1] for compatibility reasons starting from 1.2.Y we should first go through a 1.3.X before trying 1.4.Z.

[1] https://containerd.io/releases/#upgrade-path

ktsakalozos:
Thank you @balchua for your work, sorry it took me so long to merge.

balchua:
No worries! Its all good. 👍

PR #1344, opened 2020-06-23
Start dqlite and calico by default

ktsakalozos:
Start microk8s with calico and dqlite without the need to `microk8s enable ha-cluster`.

`microk8s disable ha-cluster` will get you back to etcd+flannel


apnar:
This would change the default CNI bin dir but doesn't seem to touch all the points where that may be a concern.  For example if one were to enable the cilium add-on then disable it it wouldn't revert back to a working state using calico.

The PR we're working on for the multus add-on also moves the default directory but I believe catches all the various use cases with the other add-ons.  It's still draft but the first commit there in PR  #1276 should be pretty solid for moving the default CNI directory and we should have the rest sorted in the next day or so.

Issue #1342, opened 2020-06-23
microk8s crashed with latest/stable - channel 1.18.4 revision 1499


kinowarrior:
We have:

```
microk8s is running
addons:
dns: enabled
gpu: enabled
registry: enabled
storage: enabled
```

We believe this revision has broken our GPU.   If we roll back to 

```
1.18/stable --> 1422
```

Then things work again, but we have a problem that we need more local registry space.

Unfortunately for us, currently:

1.18/stable --> 1422 revision which has no registry disk expansion option
latest/stable --> 1499 revision which breaks our GPU plugin

In our syslog:

```
Jun 22 22:32:27 systemd[1]: Mounting Mount unit for microk8s, revision 1499...
Jun 22 22:32:27 systemd[1]: Mounted Mount unit for microk8s, revision 1499.
```

Right before our entire cluster crashed.

As well as:

```
microk8s.daemon-kubelet[14122]: E0623 10:00:27.727348   14122 pod_workers.go:191] Error syncing pod 9e5db650-4f9f-4ecc-9518-e8d74b8a2e02 ("hostpath-provisioner-75fdc8fccd-9krs9_kube-system(9e5db650-4f9f-4ecc-9518-e8d74b8a2e02)"), skipping: failed to "CreatePodSandbox" for "hostpath-provisioner-75fdc8fccd-9krs9_kube-system(9e5db650-4f9f-4ecc-9518-e8d74b8a2e02)" with CreatePodSandboxError: "CreatePodSandbox for pod \"hostpath-provisioner-75fdc8fccd-9krs9_kube-system(9e5db650-4f9f-4ecc-9518-e8d74b8a2e02)\" failed: rpc error: code = Unknown desc = failed to create containerd task: OCI runtime create failed: unable to retrieve OCI runtime error (open /var/snap/microk8s/common/run/containerd/io.containerd.runtime.v2.task/k8s.io/bb945655dacdda4e5cc47ffd565d5e02725fb701cbaaeeaf819857de271fc27a/log.json: no such file or directory): nvidia-container-runtime did not terminate sucessfully: unknown"
```

Seeing `kubectl describe node` error, of the form:

```
  Warning  ContainerGCFailed        29s    kubelet, xxx     rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial unix /var/snap/microk8s/common/run/containerd.sock: connect: no such file or directory"
```

[...]

PR #1348, opened 2020-06-23
Revert "Update nvidia packages"

ktsakalozos:
Reverts ubuntu/microk8s#1299

@joedborg, @mar-kolya I am afraid I have to revert this PR. Although it works for me that I am in 20.04 we get reports we are breaking old 18.04 deployments [1]

[1] https://github.com/ubuntu/microk8s/issues/1342

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -437,7 +437,7 @@ parts:
   # b) the version dependencies of each package is met.
   nvidia-runtime:
     plugin: dump
-    source: https://nvidia.github.io/nvidia-container-runtime/ubuntu16.04/amd64/nvidia-container-runtime_3.2.0-1_amd64.deb
+    source: https://nvidia.github.io/nvidia-container-runtime/ubuntu16.04/amd64/nvidia-container-runtime_2.0.0+docker18.06.1-1_amd64.deb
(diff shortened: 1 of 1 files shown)


mar-kolya:
To be honest, I find this wrong because 20.04 is LTS and should be supported by microk8s which is promoted by Canonical. I'm not quite sure I understand why it is OK to not work properly on current LTS, but not OK to not work properly for previous LTS.

Unfortunately with this, and inability of snaps to release different versions for different distributions shows inadequacy of this ecosystem...

ktsakalozos:
@mar-kolya your point is valid. Both 20.04 and 18.04 are LTS releases and we should work on making sure GPU workloads are supported in both. We opted to revert this PR because it is better to not break 18.04 users while we try to properly support the 20.04 users. It is important to respect the time invested by 18.04 users in building their workflow around MicroK8s. At the same time it is also important to figure out how to support the 20.04 users.

Building the nvidia runtime as part of the snap build process [1] may be a solution to the problem. We need to evaluate this approach.

[1] https://github.com/ubuntu/microk8s/blob/feature/jdb/strict/snap/snapcraft.yaml#L647

PR #1371, opened 2020-07-01
Configure RBAC on all HA nodes

ktsakalozos:
When enabling the RBAC addon we ensure we configure all API servers.
When we join a node we pass the authorization mode.


ktsakalozos:
@joedborg merging this, please review when you get the time.

PR #1373, opened 2020-07-01
Handle non-resolvable node names

ktsakalozos:
This should address the comment at https://github.com/ubuntu/microk8s/issues/1127#issuecomment-648415386

Note that this is a partial fix because the first node to which other nodes connect does not set the `--hostname-override` automatically. The admin should do that. The same is expected in the current (non-ha) clustering.


ktsakalozos:
I am merging this @joedborg . Please review when you get some time.

PR #1376, opened 2020-07-02
Test on GH

ktsakalozos:
Tests are running on GH with this PR.

We also run the refresh path test where we start from 1.14 and upgrade all the way to the snap just built.


ktsakalozos:
I am merging this @joedborg please review when you have time. Thanks.

PR #1382, opened 2020-07-02
Strict confinement

joedborg:
Currently WIP to create .snaps in GH action.


oivindoh:
@joedborg trying to get a handle on the status of strict confinement support in microk8s - is this the correct place to follow progress, or has the relevant changes already been merged elsewhere?

joedborg:
Hey @oivindoh,

This is the correct place!  Sadly, we're currently dealing with higher priority tasks at the moment, so this is on hold until we can get some time to come back to it. 

If I recall correctly, the latest build should install and bring up a working MicroK8s.  What's stopping this from becoming production ready is that certain workloads are failing under strict confinement and we're trying to get to the bottom of why.  With that in mind, feel free to test this out and report any more failures you see as this might help out.

Thanks,
Joe

bkuschel:
Hi, @joedborg , Is there any update on this?

joedborg:
Hey @bkuschel 

Unfortunately, the answer is still the same as above.  You can play with the last build to experiment if you want and leave any feedback, but it's not supported as yet.

joedborg:
> You mentioned "certain workloads" in your comment above, do have any specifics about what type of workload would likely fail?

It's hard to say really.  Most of our addons currently work but certain things (that we're collecting data on) will trigger AppArmor into denials.  The end goal is to get AppArmor to understand that containerised workloads should be able to do what they like, within their containment, but this is a way off at the moment.

ktsakalozos:
The PR https://github.com/canonical/microk8s/pull/2996 is to be used from now on.

PR #1384, opened 2020-07-06
Update dqlite patch with weights

ktsakalozos:


ktsakalozos:
Nothing to review here. Just updating the dqlite patch on the HA branch.

PR #1389, opened 2020-07-07
Update wording for add-node

### LINKED ISSUE #1388: add-node message is a bit vague
The message emitted after running 'add-node' says:

"Join node with:
microk8s join ip-172-31-20-243:25000/DDOkUupkmaBezNnMheTBqFYHLWINGDbf"

or similar. For some people following along, this is interpreted  to mean execute that command immediately (i.e. not from a joining node). Suggest a wording change to make it clearer

evilnick:
Fixes #1388

### DIFF
--- a/microk8s-resources/wrappers/microk8s-add-node.wrapper
+++ b/microk8s-resources/wrappers/microk8s-add-node.wrapper
@@ -65,7 +65,8 @@ fi
 default_ip="$(get_default_ip)"
 all_ips="$(get_ips)"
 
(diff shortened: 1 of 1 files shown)


evilnick:
@tvansteenburgh  I believe @ktsakalozos is okay with this, but he wanted your review in case you thought changing the output may negatively impact people running scripts etc.

ktsakalozos:
If we agree this should land with v1.19 we can merge the PR to the `feature/ha-enable` branch.

Issue #1270, opened 2020-06-05
Centos 7.4: microk8s.join throws openssl error


pelojoy:
Running microk8s.join on Centos 7.4 version, getting the following error:

`microk8s.join 192.168.1.33:25000/oAnOGKxcsBCfkRybcDudVCdWkGamOsOf
openssl: relocation error: /snap/core/current/lib/x86_64-linux-gnu/libresolv.so.2: symbol __h_errno, version GLIBC_PRIVATE not defined in file libc.so.6 with link time reference
Traceback (most recent call last):
  File "/snap/microk8s/1253/scripts/cluster/join.py", line 387, in <module>
    update_flannel(info["etcd"], master_ip, master_port, token)
  File "/snap/microk8s/1253/scripts/cluster/join.py", line 132, in update_flannel
    get_etcd_client_cert(master_ip, master_port, token)
  File "/snap/microk8s/1253/scripts/cluster/join.py", line 107, in get_etcd_client_cert
    subprocess.check_call(cmd_cert.split())
  File "/snap/microk8s/1253/usr/lib/python3.5/subprocess.py", line 581, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['openssl', 'req', '-new', '-key', '/var/snap/microk8s/1253/certs/server.key', '-out', '/var/snap/microk8s/1253/certs/server.remote.csr', '-config', '/var/snap/microk8s/1253/certs/csr.conf']' returned non-zero exit status 127`

pelojoy:
Thanks to @Balchu’s hint, I tried to modify `join.py` to use /snap/core/current/usr/bin/openssl. And it seems to work. I need to mount the scripts to a different location to be able to modify it:

`mount --bind -o nodev,ro  /home/cluster  /var/lib/snapd/snap/microk8s/1253/scripts/cluster`

It comes back with a warning. But it does seem to work.
```
[root@zeek8s02 cluster]# microk8s.join 192.168.1.33:25000/ejRzaJDWUIovjoezgDaCUZszowzsotG 
WARNING: can't open config file: /usr/lib/ssl/openssl.cnf
```

ktsakalozos:
@pelojoy thank you for reporting the issue an taking the extra step to provide a fix.

Could you please try to set in the `check_call` the following environment variables and keep using the openssl found in the MicroK8s snap:
```
SNAP="/snap/microk8s/current"
[...]

PR #1392, opened 2020-07-08
Fix path to openssl

balbirthomas:
Running microk8s on certain distributions such as Centos 7.4, resulted
in failure to execute openssl. This happened because the path to
openssl was not being correctly set within the snap. As a result
either no openssl was found or the wrong one was selected leading
to linker relocation errors.

This commit fixes the issue by specifying the full path to openssl.

Fixes: https://github.com/ubuntu/microk8s/issues/1270


ktsakalozos:
@balbirthomas thank you for your work on this issue.

Could you use black to fix the code style issues reported? Thank you.

balbirthomas:
> @balbirthomas thank you for your work on this issue.
> 
> Could you use black to fix the code style issues reported? Thank you.

Done. Checks are running now.

PR #1406, opened 2020-07-15
Stopping control plane on HA

ktsakalozos:
Only the dqlite voters will run the controller manager and the scheduler. Running three instances of those services should be enough.

### DIFF
--- a/microk8s-resources/wrappers/control-plane-kicker
+++ b/microk8s-resources/wrappers/control-plane-kicker
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -eu
+
+export PATH="$SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH"
+ARCH="$($SNAP/bin/uname -m)"
+export IN_SNAP_LD_LIBRARY_PATH="$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/$ARCH-linux-gnu:$SNAP/usr/lib/$ARCH-linux-gnu"
+export PYTHONNOUSERSITE=false
+
+source $SNAP/actions/common/utils.sh
+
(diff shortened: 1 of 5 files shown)


balchua:
I'd like to understand a bit more about dqlite voters.  Say i have 5 node clusters.  How many are voters and how many are standby?

ktsakalozos:
> Say i have 5 node clusters. How many are voters and how many are standby?

3 nodes are voter nodes
2 nodes are stand-by nodes
the rest are spare nodes

Voter nodes replicate the DB and form the cohort in DB maintenance
Stand-by nodes replicate the DB but do not take part in the DB maintenance
Spare nodes neither replicate nor maintain the DB

balchua:
Thanks. Is the 3 node voters and 2 node spares are the maximum it will have? What happens to 7 node cluster?
What do spares do? Do the spares become voters or standbys when some nodes become unhealthy/unresponsive for a prolonged period?
Thanks again for your response.

ktsakalozos:
> What happens to 7 node cluster?

A cluster of n nodes with n > 5 will have 3 voters, 2 standbys and (n-5) spare nodes.

In case a "standby" node leaves the cluster a "spare" one takes its place. Similarly, when a "voter" leaves the cluster a "standby" node will take its place.

balchua:
Great. Thanks for the explanation.  As i understand in this pr, the `controller-manager` and `scheduler` are only needed on nodes which are voters while `apiserver` must run on all nodes.

It would be nice to have the explanation of what the voters, standbys and spares in the documentation.  Just like you did, who takes over who.

Regarding this PR, if it is a single node cluster, the `control-plane-kicker` is running, but doesn't stop the 2 services (`controller-manager` and `scheduler`).

ktsakalozos:
In a cluster at any time there has to be at least one scheduler and controller-manager. This PR makes sure the two services run on nodes acting as dqlite voters. As a result we run three schedulers and controller-managers in total.

balchua:
Thanks @ktsakalozos for the explanation.  This is great for constrained environments.

ktsakalozos:
thank you for the review and feedback.

PR #1412, opened 2020-07-16
Add multiple bundle options for Kubeflow

knkski:
Adds the ability to set `KUBEFLOW_BUNDLE` to one of `full`, `lite`, or `edge`. This will let people kick the tires on Kubeflow on their laptops, without having to have a beast of a laptop.

### DIFF
--- a/microk8s-resources/actions/enable.kubeflow.sh
+++ b/microk8s-resources/actions/enable.kubeflow.sh
@@ -106,7 +106,7 @@ def get_hostname():
 
 def main():
     password = os.environ.get("KUBEFLOW_AUTH_PASSWORD") or get_random_pass()
-    bundle = os.environ.get("KUBEFLOW_BUNDLE") or "cs:kubeflow-195"
+    bundle = os.environ.get("KUBEFLOW_BUNDLE") or "cs:kubeflow-206"
     channel = os.environ.get("KUBEFLOW_CHANNEL") or "stable"
     no_proxy = os.environ.get("KUBEFLOW_NO_PROXY") or None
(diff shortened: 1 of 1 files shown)


microk8s-resources/actions/enable.kubeflow.sh
```diff
@@ -106,7 +106,7 @@ def get_hostname():
 
 def main():
     password = os.environ.get("KUBEFLOW_AUTH_PASSWORD") or get_random_pass()
-    bundle = os.environ.get("KUBEFLOW_BUNDLE") or "cs:kubeflow-195"
+    bundle = os.environ.get("KUBEFLOW_BUNDLE") or "cs:kubeflow-206"
```
ktsakalozos, line 109: Could we use the syntax we have for passing arguments in addons (`microk8s enable kubeflow:bundle=lite`)?

Arguments (eg "bundle=lite") are passed as the first argument on the executable script [1].

[1] https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/microk8s-enable.wrapper#L39

ktsakalozos:
Very happy with this work. It means we can now turn on the kubeflow test we have disabled.

We will also need to update the docs.

microk8s-resources/actions/enable.kubeflow.sh
knkski: Given that the rest of the arguments for kubeflow are env vars, can we merge as-is, and open up a new PR that deprecates the env var arguments and switches to the string args?

ktsakalozos: Sounds good. Lets work this on a new PR.

PR #1413, opened 2020-07-17
Removing multus from arm64

ktsakalozos:


ktsakalozos:
@apnar @sergiomanso the validation of the multus addon is failing on an ARM64 on AWS.

The test pod is failing with:
```
  Warning  FailedCreatePodSandBox  2m25s (x3 over 2m53s)   kubelet, ip-172-31-8-16  (combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbo
x "9b61f39172bbeec230b0807d9edf94ca80be181c75d78678eae1eb929da06071": [default/multus-alpine:net1]: error adding container to network "net1": failed to allocate for range 0: requested IP address 10.111.111.111 is
not available in range set 10.111.111.1-10.111.111.254
```

I am disabling it for now until we figure out what is wrong.

apnar:
@ktsakalozos this likely isn't an issue with ARM64, most likely related to a previous failed or interrupt test run.  I'm using the "host-local" ipam plugin for this test which by default stores the previously assigned IP addresses in /var/lib/cni/networks/$NETWORK_NAME.  If all tests complete and the pod is taken down gracefully the IP will be released and any subsequent tests will be fine, but if something kills the run then the file is left there and the IP is still marked as used and on the next run will error out like the above since it thinks the IP is already in use (FYI, I'm intentionally using a specific static IP so I can verify that it was assigned correctly later in the test).

Note that ipam isn't the only thing to use the /var/lib/cni/networks/ directory, flannel also stores config there in the default config so I wasn't overly concerned about leaving the behavior.  It was on my list to at some point mention that this should be changed (likely to living in $SNAP_DATA).

So I'd suggest the following:
- if you wipe the /var/lib/cni/networks directory between runs you should be good
- adapt clean up scripts to clean up that directory when microk8s is removed
- I can adapt the test to store the IP files in a better directory which is already cleaned up
- Ideally all CNI would be compiled with defaults that use a directory in SNAP_DATA

apnar:
I submitted PR #1417 which does the third point above.

Issue #726, opened 2019-10-15
Running the microk8s.join command without any parameter returns an exception as opposed to a human-readable error


CalvinHartwell:
Hi all, 

When running the microk8s.join command without any parameters, it returns a python exception as opposed to a human-readable error: 

```
ubuntu@battleship:~$ microk8s.join 
Traceback (most recent call last):
  File "/snap/microk8s/973/scripts/cluster/join.py", line 352, in <module>
    if args[0] == "reset":
IndexError: list index out of range
```

My assumption is that this should give a warning or error, I.E: 

"Microk8s Error: no microk8s end point specified, please run microk8s .add-node on one node to generate the correct microk8s.join endpoint for additional cluster nodes."

Cheers,

- Calvin

evilnick:
Connected to this, you also get an exception if you enter an incorrect address:
```
Traceback (most recent call last):
  File "/snap/microk8s/982/usr/lib/python3/dist-packages/urllib3/connection.py", line 137, in _new_conn
    (self.host, self.port), self.timeout, **extra_kw)
  File "/snap/microk8s/982/usr/lib/python3/dist-packages/urllib3/util/connection.py", line 91, in create_connection
    raise err
  File "/snap/microk8s/982/usr/lib/python3/dist-packages/urllib3/util/connection.py", line 81, in create_connection
    sock.connect(sa)
OSError: [Errno 113] No route to host

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/snap/microk8s/982/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 560, in urlopen
    body=body, headers=headers)
  File "/snap/microk8s/982/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 346, in _make_request
    self._validate_conn(conn)
  File "/snap/microk8s/982/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 787, in _validate_conn
    conn.connect()
  File "/snap/microk8s/982/usr/lib/python3/dist-packages/urllib3/connection.py", line 217, in connect
    conn = self._new_conn()
  File "/snap/microk8s/982/usr/lib/python3/dist-packages/urllib3/connection.py", line 146, in _new_conn
    self, "Failed to establish a new connection: %s" % e)
[...]

PR #1416, opened 2020-07-17
check for no args in 'join' command

justinmclark:
[#726 ](https://github.com/ubuntu/microk8s/issues/726)

Question on [commented issue](https://github.com/ubuntu/microk8s/issues/726#issuecomment-543285601) below

### DIFF
--- a/scripts/cluster/join.py
+++ b/scripts/cluster/join.py
@@ -4,6 +4,7 @@ import random
 import string
 import subprocess
 import os
+import platform
 import getopt
 import sys
 import time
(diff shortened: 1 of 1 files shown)


justinmclark:
@ktsakalozos , I changed this to a draft pull request because I'd like to understand the Travis failure if it is truly an error. 

I would also like to add a test if you think it's necessary, though the change seems trivial.

justinmclark:
**Proposed solutions for commented issue regarding unreachable IP addresses:**

1. We could wrap the issue with a try/except (I'm not sure I like this solution)
2. Add a ping function to test the connection. Since I'm still learning the internals of microk8s this might be a bad idea, but I've added a commit that resolves the issue mentioned in the comment.
3. There might be a better solution I'm not thinking of.

PR #1417, opened 2020-07-17
Bug/fix multus test issue

apnar:
This should correct the issue you ran into running the test on ARM64 (although the issue wasn't directly related to ARM64).  It also includes a reversion of your previous commit removing ARM64 support.

Note, I don't have access to test on ARM64 so this is a "should fix" :)


apnar:
superseded by #1872

PR #1423, opened 2020-07-21
Added support for user CA cert and key

balbirthomas:
This commit adds an option (-r/--refresh) to the
microk8s-refresh-certs script to support user supplied
CA certificate and key.

### DIFF
--- a/microk8s-resources/wrappers/microk8s-refresh-certs.wrapper
+++ b/microk8s-resources/wrappers/microk8s-refresh-certs.wrapper
@@ -80,38 +80,35 @@ function revert {
   $SNAP/microk8s-start.wrapper
 }
 
-function refresh_ca {
+function backup_ca_crt_and_key {
   echo "Backing up CA cert and key under $BACKUP"
-  run_with_sudo mkdir -p $SNAP_DATA/var/log/ca-backup/
-  if [ -e ${SNAP_DATA}/certs/ca.crt ]
+  run_with_sudo mkdir -p "${BACKUP}"
+
+  if [ -e "${SNAP_DATA}/certs/ca.crt" ]
   then
-      run_with_sudo cp ${SNAP_DATA}/certs/ca.crt $BACKUP
-  fi
-  if [ -e ${SNAP_DATA}/certs/ca.key ]
-  then
-      run_with_sudo cp ${SNAP_DATA}/certs/ca.key $BACKUP
+      run_with_sudo cp "${SNAP_DATA}/certs/ca.crt" "$BACKUP"
   fi
 
-  echo "Creating new CA certificate and key"
-  run_with_sudo rm -f ${SNAP_DATA}/certs/ca.crt
-  run_with_sudo rm -f ${SNAP_DATA}/certs/ca.key
-
-  if [ -e ${CAPATH}/ca.crt ]
+  if [ -e "${SNAP_DATA}/certs/ca.key" ]
   then
-      run_with_sudo cp ${CAPATH}/ca.crt ${SNAP_DATA}/certs/ca.crt
+      run_with_sudo cp "${SNAP_DATA}/certs/ca.key" "$BACKUP"
+  fi
+}
+
+function refresh_ca {
+  if [ -e "${CAPATH}/ca.crt" ] && [ -e "${CAPATH}/ca.key" ]
+  then
+      backup_ca_crt_and_key
+
+      echo "Removing old CA certificate and key"
+      run_with_sudo rm -f "${SNAP_DATA}/certs/ca.crt"
+      run_with_sudo rm -f "${SNAP_DATA}/certs/ca.key"
+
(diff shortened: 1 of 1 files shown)


balchua:
Just my 2 cents, i do find it better if user can do either:
* To generate certs with user supplied ca certs. `microk8s refresh-certs -i /path/to/custom/certs` 
* To generate certs with microk8s self signed ca certs. `microk8s refresh-certs -i`

This is just me. 😁

microk8s-resources/wrappers/microk8s-refresh-certs.wrapper
```diff
@@ -73,6 +80,43 @@ function revert {
   $SNAP/microk8s-start.wrapper
 }
 
+function refresh_ca {
+  echo "Backing up CA cert and key under $BACKUP"
```
ktsakalozos, line 84: Start this function by checking if the user provided the CA files instead of taking a back up and then restoring the old CA from the backup. Something like:
```
if ! [ -e ${CAPATH}/ca.crt ]
then
  fail fast
fi
```

microk8s-resources/wrappers/microk8s-refresh-certs.wrapper
```diff
@@ -73,6 +80,43 @@ function revert {
...
+  then
+      run_with_sudo cp ${CAPATH}/ca.key ${SNAP_DATA}/certs/ca.key
+  else
+      echo "Could not find ca.key in $CAPATH, aborting cert update"
+      run_with_sudo cp ${BACKUP}/ca.crt ${SNAP_DATA}/certs/ca.crt
+      run_with_sudo cp ${BACKUP}/ca.key ${SNAP_DATA}/certs/ca.key
+      exit 1
+  fi
+
+  update_configs
```
ktsakalozos, line 118: With this update_configs we skip the call to `produce_certs` [1]. However the `produce_certs` is using the CA to issue other certificates, see for example [2] where we sign the server certificate with the CA. If we update the CA the rest of the certificates need to be re-issued.

My suggestion is to add the CA files as an optional argument to `produce_certs`. If no argument is provided then `produce_certs`  will have the original behavior. If we provide the CA to the `produce_certs` then we will re-issue the certificates following the execution path of a `refresh` but instead of creating the CA files in `produce_certs`, use the CA files provided. 

[1] https://github.com/ubuntu/microk8s/pull/1423/files#diff-0805575b10502a2e22b62c908aeedf1fL88
[2] https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/common/utils.sh#L369

microk8s-resources/wrappers/microk8s-refresh-certs.wrapper
balbirthomas: Ok will do.

microk8s-resources/wrappers/microk8s-refresh-certs.wrapper
balbirthomas: Ok will investigate and change as requested.

balbirthomas: I have added a line to `refersh_ca` that re-signs the server certificate by calling `gen_server_cert`. This seems to be working well, and fixes the issue of having to restart host system (Thank you). It needs further testing and some investigation on what type of validation and verification of CA cert and key must be done before ingesting them.

balbirthomas:
> This is great!! Thanks!
> 
> Just my 2 cents, i do find it better if user can do either:
> 
>     * To generate certs with user supplied ca certs. `microk8s refresh-certs -i /path/to/custom/certs`
> 
>     * To generate certs with microk8s self signed ca certs. `microk8s refresh-certs -i`
> 
> 
> This is just me. grin

The current task/ticket aims to implement the functionality in the backend. It may be exposed in the front end when stable and well tested.

ktsakalozos:
> Just my 2 cents, i do find it better if user can do either:
> 
> * To generate certs with user supplied ca certs. `microk8s refresh-certs -i /path/to/custom/certs`
> * To generate certs with microk8s self signed ca certs. `microk8s refresh-certs -i`

What you suggest here @balchua makes sense. However, having a space right after `-i` becomes an issue for `getopt` see [1].

I think what we should do is:
```
# to install a new CA:
microk8s refresh-certs -i /path/to/custom/certs
# to refresh the current CA with an autogenerated one:
microk8s refresh-certs -r
```

[1] https://stackoverflow.com/questions/26431682/why-cant-i-have-a-space-between-option-and-optional-argument-using-getopt

balchua:
@ktsakalozos, so the path should follow the `-i` immediately instead of space.  I guess that is acceptable.
Or use python script to avoid confusion.

ktsakalozos:
Closing this PR in favor of https://github.com/ubuntu/microk8s/pull/1486

PR #1424, opened 2020-07-21
Fix LXC deployments, umount kubelet when removing microk8s

ktsakalozos:


ktsakalozos:
@joedborg please review when you have time.

PR #1428, opened 2020-07-23
NO_PROXY info in containerd-env comments

ktsakalozos:


ktsakalozos:
@chris-sanders what do you think about this? Thanks.

chris-sanders:
I like it. Even of someone doesn't read the docs this should get them where they need.

ktsakalozos:
Thank you.

Issue #1173, opened 2020-05-01
feature request: Standard way for microk8s to communicate about a local registry


nicks:
Many Kubernetes clusters support a local registry, to make pushing images directly to the cluster fast. But every cluster documents it with manual shell scripts:

- https://kind.sigs.k8s.io/docs/user/local-registry/
- https://microk8s.io/docs/registry-built-in
- https://github.com/rancher/k3d/blob/master/docs/registries.md#using-a-local-registry
- https://docs.openshift.com/container-platform/4.3/registry/accessing-the-registry.html#registry-accessing-directly_accessing-the-registry

Because every cluster is slightly different, this makes it hard for tools to automate this work for users.

## The big idea

Clusters should use annotations on the `kube-system` namespace to document how to configure a registry

## Detailed design

I propose 3 new "standard" annotations:

`x-k8s.io/registry`: the host of the registry (hostname and port), as seen from outside of the cluster
`x-k8s.io/registry-from-cluster`: the host of the registry (hostname and port), as seen from inside the cluster
`x-k8s.io/registry-help`: The URL of documentation where users can read more about how to set a registry up. 

Tools that would benefit from local registries should read this annotation and display it to the user

## What this means for microk8s

All microk8s clusters would have a `registry-help` annotation pointing users to https://microk8s.io/docs/registry-built-in

When you run `microk8s enable registry`, microk8s would automatically add the `x-k8s.io/registry` annotation to nodes in the cluster.

## Prior Art

This proposal is based on what we've been recommending to people in Tilt, with a lot of success: https://docs.tilt.dev/choosing_clusters.html#custom-clusters

*update 5/4* - changed the proposal to use the `kube-system` namespace instead of node annotations

nicks:
I also filed this feature request here https://github.com/kubernetes-sigs/kind/issues/1543 and here https://github.com/rancher/k3d/issues/234, and would love other ideas on places I should post about this

nicks:
Just to keep you all in the loop, sig-cluster-lifecycle has suggested I open a KEP for this to kick the tires more on the proposal.

But I also recognize that microk8s falls slightly outside the KEP process. Are the microk8s maintainers ok with that?

ktsakalozos:
Hi @nicks 

Thank you for the suggestion and apologies for the late response.

I am +1 to the idea of formalizing on the registry use. I see your proposal this is under discussion. As soon as we have something concrete we can test and ship we will gladly do so.

[...]

PR #1437, opened 2020-07-27
add support for local registry discovery

nicks:
Fixes https://github.com/ubuntu/microk8s/issues/1173

### DIFF
--- a/microk8s-resources/wrappers/apiservice-kicker
+++ b/microk8s-resources/wrappers/apiservice-kicker
@@ -16,6 +16,7 @@ then
 fi
 
 restart_attempt=0
+installed_registry_help=0
 while true
 do
     if [ $restart_attempt -ge 5 ]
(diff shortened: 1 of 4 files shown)


microk8s-resources/wrappers/apiservice-kicker
```diff
@@ -45,4 +46,17 @@ do
             restart_attempt=0
         fi
     fi
+
+    # If no local-registry-hosting documentation has been installed,
+    # install a help guide that points to the microk8s registry docs.
+    if [ $installed_registry_help -eq 0 ] &&
```
nicks, line 52: I put this code in the apiservice-kicker, because it needs to happen after the apiservice comes up. 

This seemed like the best fit for that kind of a life-cycle hook in the current code (even if the fit isn't that great)

But I'm completely open to suggestions on better places to put it!

ktsakalozos: Indeed we do not have a place to put any maifests we want to apply at boot time.

As you said, it is good enough for now. We will explore alternatives if this causes problems.

ktsakalozos:
@nicks is there any doc/blog post we could supplement the 1.19 release announcement regarding this feature and push to our social channels? Thanks.

nicks:
@ktsakalozos Not a lot right now aside from https://docs.tilt.dev/choosing_clusters.html#discovering-the-registry and the KEP itself (https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/generic/1755-communicating-a-local-registry)

I'm hoping to get a few more implementations in the field and then write about it more on our blog

PR #1441, opened 2020-07-28
Regular Kubeflow argument handling

knkski:
Updates it to match how other addons handle arguments, instead of environment variables. Keeps the environment variables around for now.


ktsakalozos:
Where are we going to document these arguments? Should we create a page for the kubeflow addon to mention the arguments? https://microk8s.io/docs/addons#heading--list

knkski:
@ktsakalozos: Sure, how do I go about creating that page?

PR #1452, opened 2020-07-30
Add message in joining

ktsakalozos:
When joining an HA cluster we print a  `Contacting cluster at <node>` right at the beginning so we do not leave the user wondering if the command works or not.

```
$ sudo microk8s join 10.228.72.149:25000/c30b882c33b126b388af31c61573d4f6
Contacting cluster at 10.228.72.149
Waiting for node to join the cluster. .. .. 
```


ktsakalozos:
@knkski this came from a feedback you provided some time ago.

ktsakalozos:
Done @knkski. Thank you for the review.

Issue #1447, opened 2020-07-29
[HA] enabling / disabling cilium addon


balchua:
Enabling Cilium on HA cluster does not disable Calico.
This makes it 2 cni running in the cluster.

In short enabling Cilium addon must stop calico while disabling Cilium must enable calico.

balchua:
Pr merged

PR #1453, opened 2020-07-30
Fix cilium in an ha cluster

ktsakalozos:
Enabling and disabling cilium works on a single node cluster with this PR. I think that was always the case.

Fixes: https://github.com/ubuntu/microk8s/issues/1447

### DIFF
--- a/microk8s-resources/actions/disable.cilium.sh
+++ b/microk8s-resources/actions/disable.cilium.sh
@@ -35,7 +35,7 @@ then
   run_with_sudo $SNAP/sbin/ip link delete "cilium_vxlan"
 fi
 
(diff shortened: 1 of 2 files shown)


ktsakalozos:
> Why do we need to restart flannel in this case?

Deployments following the latest/stable will not be automatically upgraded to an HA setup where calico is used. These deployments will keep using flannel. The restart in https://github.com/ubuntu/microk8s/pull/1453/files#diff-9a6aa719fb63279e55dac696e77f3bd7R45 is because the deployment may have been on of those.

balchua:
Ok so in a multi node non HA, cilium addon is not seamless.  One has to manually remove the vxlan created by flannel.
But in HA mode, Cilium addon doesn't need to remove any vxlan.
Overall this LGTM.

PR #1457, opened 2020-07-31
Bugfix status 

balchua:
This PR fixes the bug when `microk8s status -wait-ready` is executed.  It waits for the kubernetes node to report `Ready` state.


balchua:
Suddenly `pytest` isn't available anymore.  That is why it is failing.

PR #1458, opened 2020-07-31
update fluentd addon includes kibana and elasticsearch.

balchua:
Upgraded the following:
* fluentd to v3.0.2
* elasticsearch to v7.4.2
* kibana to v7.2.0


strigona-worksight:
FYI it is recommended to run the same minor version of Kibana as Elasticsearch. https://www.elastic.co/guide/en/kibana/7.4/setup.html#elasticsearch-version

balchua:
@strigona-worksight thanks for the link.  I took the addon yamls from upstream kubernetes. 
here is the link to the [repo](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/fluentd-elasticsearch/es-image).
The elasticsearch image used is also `7.4.2`
https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-elasticsearch/es-image/Dockerfile#L20, same version as kibana.  I will update the description above.

strigona-worksight:
Ah, fair enough. Odd, it looks like it was probably just an oversight from upstream. From the same documentation, kibana 7.2 will work with elasticsearch 7.4, but not the other way around. Generally speaking, I think most people upgrade them together.

balchua:
It is rather strange that i couldn't find v7.4.2 at `docker.elastic.co/kibana/kibana-oss`.  For now i will leave it as the same as from upstream.  FYI. the way elastic is installed is not recommended for production setup.

strigona-worksight:
I did notice that kibana image wasn't listed there, but I was able to pull it. I also made a note to upstream about the kibana/elasticsearch discrepancy so perhaps they'll fix it in the near future.

Yes, I also agree that it's not a recommended production setup, though it's probably good enough. It'd probably be better to stand up a cluster using the Elasticsearch Operator, granted I'm sure that would open up a whole can of worms.

PR #1463, opened 2020-08-01
Upgrade prometheus operator

balchua:
Upgraded components:
* prometheus-operator: v0.40.0
* grafana - 7.1.0
* prometheus - v2.20.0


ktsakalozos:
Thank you for this work @balchua . 

With this PR the prometheus addon will start deploying what is released from the kube-prometheus project. Which release of kube-prometheus are we putting here? Release 0.6.0?

kube-prometheus is really great and is of great benefit to the community. What worries me a bit is the maturity of the end product. I see for example images coming from developer accounts [1, 2].

It makes me wonder if this is the right time to make the switch to kube-prometheus. What are your thoughts?

[1] https://github.com/ubuntu/microk8s/pull/1463/files#diff-bd3850cfcf23201318dbac3020c867d3R28
[2] https://github.com/ubuntu/microk8s/pull/1463/files#diff-358e176fdeca4a1e05d59d81699da5ebR27 (this comes directly from the prometheus operator)

ktsakalozos:
Thank you for this work @balchua . 

With this PR the prometheus addon will start deploying what is released from the kube-prometheus project. Which release of kube-prometheus are we putting here? Release 0.6.0?

kube-prometheus is really great and is of great benefit to the community. What worries me a bit is the maturity of the end product. I see for example images coming from developer accounts.

It makes me wonder if this is the right time to make the switch to kube-prometheus. Your thoughts?

balchua:
Good catch. Well the kube-prometheus project is a combo of all things prometheus.  But i honestly find it too complicated.  I do prefer to grab the prometheus and grafana addon from the typhoon project.
https://github.com/poseidon/typhoon/tree/master/addons.

Having that aside, we probably need to follow the instructions on kube-prometheus on how to build the manifests using jsonnet. From there we can pin the kube-prometheus version according to our preference.

balchua:
Another alternative is we take the release versioned source and extract the `manifest` folder.  This way we have a consistent version in microk8s. Wdyt?

ktsakalozos:
> Another alternative is we take the release versioned source and extract the `manifest` folder

Yes this is what we should be doing. We should take the manifests from the latest released kube-prometheus at https://github.com/prometheus-operator/kube-prometheus/tree/release-0.6/manifests .

A correction on my part. We were always shipping kube-prometheus as the prometheus addon.

balchua:
Ok its clear now.  I will update the PR and the readme on the process we follow to update the addon.

ktsakalozos:
Just asked in https://github.com/prometheus-operator/kube-prometheus/discussions/642 about the images from developers accounts.

balchua:
Ok as i understand we will use the one published in quay.

ktsakalozos:
> Ok as i understand we will use the one published in quay.

No. Lets use the same images upstream kube-prometheus put in the 0.6 release. I would rather have a tested image from a dev than an untested from an official org.

balchua:
I've made some changes to how the addon is updated and enabled.

Issue #1397, opened 2020-07-10
Kubeflow enable checks deadlock


ktsakalozos:
We have two checks for checking the status of kubeflow, the ambassador operator [1] and the uk8s model [2]. If the uk8s model exists but the operator is not running the use addon can neither be enabled nor disabled.

A way to reproduce this is:
```
sudo microk8s enable kubeflow
sudo microk8s kubectl delete ns kubeflow
``` 
[1] https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/addon-lists.yaml#L143
[2] https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/enable.kubeflow.sh#L183

Issue #958, opened 2020-02-11
Failed to enable kubeflow


titsuki:
[inspection-report-20200211_171636.tar.gz](https://github.com/ubuntu/microk8s/files/4184636/inspection-report-20200211_171636.tar.gz)

+ Error Message:
```
$ microk8s.enable kubeflow
Enabling dns...
Enabling storage...
Enabling dashboard...
Enabling ingress...
Enabling rbac...
Enabling juju...
Kubeflow could not be enabled:
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
 22  116M   22 25.8M    0     0  51231      0  0:39:39  0:08:48  0:30:51 71301
curl: (18) transfer closed with 94828385 bytes remaining to read

Command '('microk8s-enable.wrapper', 'juju')' returned non-zero exit status 1
Failed to enable kubeflow
```

+ microk8s version:
```
$ sudo snap find microk8s
Name      Version  Publisher   Notes    Summary
microk8s  v1.17.2  canonical✓  classic  Kubernetes for workstations and appliances
```

Maybe this issue is same as #943 but I'm a microk8s newbie and cannot make a 
 correct judgement about it.

mattk08:
I'm having a similar problem enabling kubeflow, slightly different error

```
microk8s.enable kubeflow
Enabling dns...
Enabling storage...
Enabling dashboard...
Enabling ingress...
Enabling rbac...
Enabling juju...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (77) Problem with the SSL CA cert (path? access rights?)

Command '('microk8s-enable.wrapper', 'juju')' returned non-zero exit status 1
```

```
snap find microk8s
Name      Version  Publisher   Notes    Summary
microk8s  v1.17.2  canonical✓  classic  Kubernetes for workstations and appliances
```

mattk08:
https://github.com/ubuntu/microk8s/issues/943#issuecomment-581424578

suggested

```
microk8s.disable kubeflow
microk8s.enable kubeflow
```

but that did not help

andrewcheny:
My error message looks like this:

[...]

PR #1467, opened 2020-08-03
Switch microk8s enable wrapper to Python script

### LINKED ISSUE #1446: Switch to unix-style argument handling for addon enabling
This is a proposal to switch to regular unix-style argument handling when enabling addons. Briefly, instead of this:

    microk8s.enable kubeflow:"bundle=lite"

switch to this:

    microk8s.enable kubeflow --bundle lite

The big issue to solve is that right now, users can enable multiple addons at once like this:

    microk8s.enable gpu kubeflow

So, we run into an issue of how to handle argument parsing with multiple addons at once. The easiest thing to do would be to just disallow enabling multiple addons at once when using unix-style argument handling. The only issue with this is that an addon may want positional arguments. For example, say the `kubeflow` addon allowed a positional argument of which bundle size to deploy, like this:

    microk8s.enable kubeflow lite

If the user types `microk8s.enable kubeflow gpu dns`, do we parse that as the user wanting 3 addons enabled, or only `kubeflow` and `dns` enabled, and they mistakenly typed `gpu` instead of `full`?

However, if we stick to only parameter flags such as `--foo bar`, it should be possible to implement unix-style argument handling in a backwards-compatible way. In pseudocode, we could have logic like this:

```python
if all(a.split(':')[0] in ADDONS for a in ARGS):
    # We have a command line made entirely of addon names, with or without params
    # e.g. `microk8s.enable foo bar:"123" baz`
    for addon in ARGS:
        enable(addon)
else:
    # Treat as unix-style arguments, where enabling multiple addons at once is disallowed
    # Positional arguments are also disallowed
    ...
```

As a further proposal, I think we should move to [click](https://click.palletsprojects.com/en/7.x/), which will make it easy to create a cohesive CLI experience, as it automatically handles stuff such as argument parsing and `--help` flags.

knkski:
For now, it's just a clone of the old bash script, with the ability to handle unix-style flag arguments. Using click however, will allow expanding the enable scripts to be full click subcommands, instead of just bash scripts.

Fixes #1446


microk8s-resources/wrappers/microk8s-enable.wrapper
```diff
@@ -1,64 +1,60 @@
...
-    exit 2
-  fi
-done
-exit $result
+    if all(a.split(':')[0] in ADDONS for a in addons) and len(addons) > 1:
+        for addon in addons:
+            addon, *args = addon.split(':')
+            run([PARENT / ('actions/enable.%s.sh' % addon)] + args.split(';'))
+    else:
+        addon, *args = addons[0].split(':')
```
johnsca, line 26: Seems like this would blow up with an IndexError if the user calls this without any args. Should probably have a check for an empty addons list and report a reasonable error.

knkski: Added `required=True` which forces at least one argument to be passed, otherwise help text is displayed.

balchua:
@knkski do we drop the checking for addons if they are already running?  If yes, i think this is required, sometimes users manually edit the resource, and when the user enabled it again, it loses their previous changes.

knkski:
@balchua: The use case for dropping that check is a result of #1397. If a user enables kubeflow and it either fails or the `kubeflow` namespace is later deleted, a user will be unable to either enable or disable kubeflow.

It would be nice to have `microk8s.enable kubeflow` support redeploying in case of a failed enable attempt, but maybe there's a better way to handle that. @ktsakalozos, @balchua: thoughts? cc @RFMVasconcelos, who I think also had opinions on that.

balchua:
Thanks @knkski for the info. Instead of dropping that altogether, maybe we can consider adding `idempotent: true` or `false` into the `addon-list.yaml`?  If `idempotent` is `true` then we dont reapply otherwise we reapply.
Another option i can think of is an option `--force` that users can pass to the enable action.  Example:
`microk8s enable --force kubeflow`
These are just options on top of my head.

ktsakalozos:
> sometimes users manually edit the resource, and when the user enabled it again, it loses their previous changes.

The enable script used to reapply the manifests without checking if the addon was already enabled. This was causing some confusion [2] and we had to fix it [1].

@knkski kubeflow enables the dns addon but I also think it expects the forward dns to be correctly set. Will the kubeflow deploy in an environment where 8.8.8.8 is not reachable?

[1] https://github.com/ubuntu/microk8s/pull/927
[2] https://github.com/ubuntu/microk8s/issues/958

balchua:
So does it mean that `enable` command do not need to check whether it is enabled or not?

ktsakalozos:
Sorry I was not clear. We need to keep the current behavior of not re-applying the addon manifests. We have to respect any configuration users have on the running workloads.

balchua:
Aaa ok. Thanks for clarifying.

knkski:
I've updated this PR so that it keeps the current behavior of not enabling already-enabled addons, for everything but Kubeflow, which has it's own logic for handling that. Namely, it will allow disabling Kubeflow even if Kubeflow is already disabled, which lets us fix #1397. 

I didn't create a `idempotent: true` section in `addon-list.yaml`, opting instead for a hardcoded check for the kubeflow addon, but I could add that if desired.

balchua:
Thanks @knkski.  Im good with that.  The unix style args is nice.

microk8s-resources/wrappers/microk8s-disable.wrapper
```diff
@@ -1,76 +1,14 @@
...
-    echo
-    actions="$(find "${SNAP}/actions" -name '*.yaml' -or -name 'disable.*.sh')"
-    actions="$(echo "$actions" | sed -e 's/.*[/.]\([^.]*\)\..*/\1/' | sort | uniq)"
-    for action in $actions; do
-        echo "  $action"
-    done
-    exit 0
-fi
-
-exit_if_stopped
```
ktsakalozos, line 33: I think we should keep this check. If microk8s is stopped we do not want to try to enable anything, right?

ktsakalozos:
@knkski there are a few conflicts in this PR. Do you think you could address them so we get this PR ready for 1.20? Thank you.

scripts/wrappers/common/utils.py
```diff
@@ -211,3 +225,64 @@ def set_service_expected_to_start(service, start=True):
...
+    else:
+        addon, *args = addons[0].split(':')
+
+        if addon in xabled_addons and addon != 'kubeflow':
+            click.echo("Addon %s is already %sd." % (addon, action))
+            sys.exit(0)
+
+        if addon not in existing_addons:
+            click.secho("Addon `%s` not found." % addon, fg='red', err=True)
+            click.echo("The available addons are:\n - %s" % '\n - '.join(existing_addons), err=True)
```
ktsakalozos, line 260: I am not so sure about this. If we print the list of addons here I easily see a request to put descriptions next to the addon names.

scripts/wrappers/common/utils.py
```diff
@@ -102,13 +113,10 @@ def get_dqlite_info():
...
-    clusterLockFile = os.path.expandvars("${SNAP_DATA}/var/lock/clustered.lock")
-    if os.path.isfile(clusterLockFile):
-        print(
-            "This MicroK8s deployment is acting as a node in a cluster. "
-            "Please use the microk8s status on the master."
-        )
-        exit(0)
+    if (snap_data() / 'var/lock/clustered.lock').exists():
+        click.echo('This MicroK8s deployment is acting as a node in a cluster.')
+        click.echo('Please use `microk8s enable` on the master.')
```
ktsakalozos, line 118: Based on the error message, I assume method was used by `microk8s.status`.

scripts/wrappers/common/utils.py
```diff
@@ -211,3 +225,64 @@ def set_service_expected_to_start(service, start=True):
...
+
+        if addon not in existing_addons:
+            click.secho("Addon `%s` not found." % addon, fg='red', err=True)
+            click.echo("The available addons are:\n - %s" % '\n - '.join(existing_addons), err=True)
+            sys.exit(1)
+
+        if args and addons[1:]:
+            click.secho(
+                click.style(
+                    "Can't pass string arguments and flag arguments simultaneously!\n", fg='red'
```
ktsakalozos, line 266: I like what you are doing here, error messages in red. Could we do this across all microk8s in a separate PR? If we do this coloring only here it will look odd.

ktsakalozos:
Just reverted this PR, there are a few things (probably minor) that will need a bit more attention.

On enable/disable we request the status of each addon [1]. We should first make sure the API server is running or else the get status fails with:
```
14:55:29 2020-11-27 at 12:55:29 | INFO + /snap/microk8s/1829/actions/enable.cilium.sh
14:55:29 2020-11-27 at 12:55:29 | INFO Traceback (most recent call last):
14:55:29 2020-11-27 at 12:55:29 | INFO File "/snap/microk8s/1829/scripts/wrappers/enable.py", line 32, in <module>
14:55:29 2020-11-27 at 12:55:29 | INFO enable(prog_name='microk8s enable')
14:55:29 2020-11-27 at 12:55:29 | INFO File "/snap/microk8s/1829/usr/lib/python3/dist-packages/click/core.py", line 716, in __call__
14:55:29 2020-11-27 at 12:55:29 | INFO return self.main(*args, **kwargs)
14:55:29 2020-11-27 at 12:55:29 | INFO File "/snap/microk8s/1829/usr/lib/python3/dist-packages/click/core.py", line 696, in main
14:55:29 2020-11-27 at 12:55:29 | INFO rv = self.invoke(ctx)
14:55:29 2020-11-27 at 12:55:29 | INFO File "/snap/microk8s/1829/usr/lib/python3/dist-packages/click/core.py", line 889, in invoke
14:55:29 2020-11-27 at 12:55:29 | INFO return ctx.invoke(self.callback, **ctx.params)
14:55:29 2020-11-27 at 12:55:29 | INFO File "/snap/microk8s/1829/usr/lib/python3/dist-packages/click/core.py", line 534, in invoke
14:55:29 2020-11-27 at 12:55:29 | INFO return callback(*args, **kwargs)
14:55:29 2020-11-27 at 12:55:29 | INFO File "/snap/microk8s/1829/scripts/wrappers/enable.py", line 25, in enable
14:55:29 2020-11-27 at 12:55:29 | INFO enabled_addons, _ = get_status(get_available_addons(get_current_arch()), True)
14:55:29 2020-11-27 at 12:55:29 | INFO File "/snap/microk8s/1829/scripts/wrappers/status.py", line 157, in get_status
14:55:29 2020-11-27 at 12:55:29 | INFO kube_output = kubectl_get("all")
14:55:29 2020-11-27 at 12:55:29 | INFO File "/snap/microk8s/1829/scripts/wrappers/common/utils.py", line 169, in kubectl_get
14:55:29 2020-11-27 at 12:55:29 | INFO return run("kubectl", kubeconfig, "get", cmd, "--all-namespaces", die=False)
14:55:29 2020-11-27 at 12:55:29 | INFO File "/snap/microk8s/1829/scripts/wrappers/common/utils.py", line 39, in run
14:55:29 2020-11-27 at 12:55:29 | INFO result.check_returncode()
14:55:29 2020-11-27 at 12:55:29 | INFO File "/snap/microk8s/1829/usr/lib/python3.5/subprocess.py", line 659, in check_returncode
14:55:29 2020-11-27 at 12:55:29 | INFO self.stderr)
14:55:29 2020-11-27 at 12:55:29 | INFO subprocess.CalledProcessError: Command '('kubectl', '--kubeconfig=/var/snap/microk8s/1829/credentials/client.config', 'get', 'all', '--all-namespaces')' returned non-zero exit status 1
```

Similar case is when we enable/disable addons. We used to wait for up to 30 seconds for the API server to get ready [2]. This way if an addon enables/disables the apiserver we would not fail in [3].

The last thing that needs some mending is the `--help`. Most of the addons do not have a `--help` so we are pushing the users to this wrong path:
```
ubuntu@ip-172-31-40-63:~$ microk8s.disable --help
Usage: microk8s disable [OPTIONS] ADDONS...

  Disables one or more MicroK8s addons.

  For a list of available addons, run `microk8s status`.

  To see help for individual addons, run:

      microk8s disable ADDON -- --help

Options:
  --help  Show this message and exit.
ubuntu@ip-172-31-40-63:~$ microk8s.disable dns -- --help
Disabling DNS
Reconfiguring kubelet
Removing DNS manifest
serviceaccount "coredns" deleted
configmap "coredns" deleted
deployment.apps "coredns" deleted
service "kube-dns" deleted
clusterrole.rbac.authorization.k8s.io "coredns" deleted
clusterrolebinding.rbac.authorization.k8s.io "coredns" deleted
DNS is disabled
```
We should get back the old help message.


[1] https://github.com/ubuntu/microk8s/pull/1467/files#diff-44f8a77098734624eb59a3c59e17024d622e15fb9927eb4d6f3583ac724ffe25R25 
[2] https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/microk8s-enable.wrapper#L31
[3] https://github.com/ubuntu/microk8s/pull/1467/files#diff-cea8894ac12cb68765874a03fba91e79d22c596d86633886b5c02abe52829448R246

Issue #196, opened 2018-11-19
Cant push image to local registry 


lethevimlet:
I enabled registry with
`mircrok8s.enable registry`

then I successfully build my image with a Dockerfile.
`microk8s.docker build -t localhost:32000/test:1 . `

But I can't push a custom image to the local registry
`microk8s.docker push localhost:32000/test:1`

It will stay stuck with the following message
`The push refers to a repository [localhost:32000/test]`

Am I missing something? any ideas of what could be wrong with the registry or my setup?

ktsakalozos:
Hi @lethevimlet ,

Could you share the report produced by `microk8s.inspect`? Would you also be able to enable debug logs on dockerd and the docker client? https://github.com/ubuntu/microk8s#configuring-microk8s-services 

Thanks

lethevimlet:
Unfortunately, the results produced by microk8s.inspect yield sensitive information that I can't share since it's against our privacy policy but I can tell you more information about the setup if that helps to replicate the issue.

I'm using Ubuntu Server  16.04.5 LTS 
Microk8s was installed with the following command
`snap install microk8s --classic`

inspect shows the following error for docker:
nov 19 18:05:51 kubernetes microk8s.daemon-docker[13518]: time="2018-11-19T18:05:51.402240293+01:00" level=info msg="failed to mount layer sha256:------------------------------------------------------ (sha256:------------------------------------------------------ ) from docker.io/library/node: errors:\ndenied: requested access to the resource is denied\nunauthorized: authentication required\n"

ktsakalozos:
Do you think you could share a docker image I could build that would cause this error?

Just to give you some context, the docker registry is set to be insecure: https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/default-args/docker-daemon.json and this is how we test the registry before every release https://github.com/ubuntu/microk8s/blob/master/tests/validators.py#L182

ktsakalozos:
I am basically asking for some instructions on how to reproduce the error.

ktsakalozos:
Just to double check, you do not have a second dockerd installed on your system, right?

ktsakalozos:
On a cluster that looks like this:
```
NAMESPACE            NAME                                       READY   STATUS    RESTARTS   AGE
container-registry   pod/registry-5f6c6bf97f-ds8wd              1/1     Running   0          20m
kube-system          pod/hostpath-provisioner-98d6db847-kpkvh   1/1     Running   0          20m
[...]

PR #1470, opened 2020-08-04
enabling the registry should fail if localhost can resolve to an ipv6 address

nicks:
Fixes https://github.com/ubuntu/microk8s/issues/196

I tested this a little bug a little bit with different microk8s channels. This comment (https://github.com/ubuntu/microk8s/issues/196#issuecomment-590403234) was particularly helpful in confirming that there's some underlying kubernetes networking issue, rather than something specific to `docker push`.

I can definitely reproduce the problem on --channel=1.17/stable. It might be fixed on --channel=1.18/stable, but supposedly ipv6 support on v1.18 is still in beta, so it seems safer to error on v1.18 too.

### DIFF
--- a/microk8s-resources/actions/enable.registry.sh
+++ b/microk8s-resources/actions/enable.registry.sh
@@ -4,6 +4,14 @@ set -e
 
 source $SNAP/actions/common/utils.sh
 
+hostv6=$(getent ahostsv6 localhost | head -n1 | cut -d' ' -f1)
+if [ "$hostv6" = "::1" ]; then
+  echo "In-cluster registry does not work with ipv6"
+  echo "Remove the mapping from ::1 to localhost in /etc/hosts and try again"
+  echo "For more details, see https://github.com/ubuntu/microk8s/issues/196"
+  exit 1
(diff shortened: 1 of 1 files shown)


nicks:
oooh the test failure is interesting...lemme think about this. i wonder if it makes sense to
a) fix the test to use ipv4, or
b) use an environment variable to override the ipv6 check, or
c) both

nicks:
OK, here's where I currently am:

- the registry works fine with ipv6 on master (k8s v1.18)
- the registry fails with ipv6 on all previous k8s releases (k8s v1.17)

does that mean it makes sense to backport this fix to the microk8s v1.17 channel? i'm not sure. is there a general policy for when it makes sense to fix things for old kubernetes versions in microk8s?

ktsakalozos:
Thank you for this PR @nicks and for the effort of testing other versions.

We normally back port security fixes (not features) to the past three k8s releases (currently 1.18, 1.17, 1.16).

As 1.19 is a couple of weeks away I do not think we would be back porting this to 1.16.

PR #1482, opened 2020-08-12
Add flake8 to CI checks

knkski:
black doesn't catch some things such as unused variables or trailing spaces in docstrings, so add flake8 to the CI process.

### DIFF
--- a/installer/vm_providers/_multipass/_multipass.py
+++ b/installer/vm_providers/_multipass/_multipass.py
@@ -15,7 +15,6 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import logging
-import os
 import sys
 from typing import Dict, Optional, Sequence
 
--- a/installer/vm_providers/_multipass/_multipass.py
+++ b/installer/vm_providers/_multipass/_multipass.py
@@ -47,8 +47,6 @@ class Multipass(Provider):
         return "multipass"
 
     def run(self, command: Sequence[str], hide_output: bool = False) -> Optional[bytes]:
-        env_command = self._get_env_command()
-
         cmd = ["sudo"]
         cmd.extend(command)
         self._log_run(cmd)
--- a/microk8s-resources/actions/enable.kubeflow.sh
+++ b/microk8s-resources/actions/enable.kubeflow.sh
@@ -211,7 +211,7 @@ def main():
             args[arg] = strtobool(args[arg])
 
     with open("/proc/meminfo") as f:
-        memtotal_lines = [l for l in f.readlines() if "MemTotal" in l]
+        memtotal_lines = [line for line in f.readlines() if "MemTotal" in line]
 
     try:
         total_mem = int(memtotal_lines[0].split(" ")[-2])
--- a/scripts/cluster/agent.py
+++ b/scripts/cluster/agent.py
@@ -9,6 +9,7 @@ import string
 import subprocess
 import sys
 import time
+from typing import List
 
 import yaml
 
--- a/scripts/cluster/agent.py
+++ b/scripts/cluster/agent.py
@@ -24,7 +24,7 @@ from .common.utils import (
     try_initialise_cni_autodetect_for_clustering,
 )
 
-from flask import Flask, jsonify, request, abort, Response
+from flask import Flask, jsonify, request, Response
 
 app = Flask(__name__)
 CLUSTER_API = "cluster/api/v1.0"
--- a/scripts/cluster/agent.py
+++ b/scripts/cluster/agent.py
@@ -292,8 +292,8 @@ def join_node_etcd():
 
     if is_node_running_dqlite():
         msg = (
-            "Failed to join the cluster. This is an HA dqlite cluster. \n"
-            "Please, retry after enabling HA on this joining node with 'microk8s enable ha-cluster'."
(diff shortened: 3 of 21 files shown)


knkski:
Added `flake8` to the CI process without fixing any of the errors. If the lints look good to actually enforce, I'll go ahead and fix them in this PR.

installer/vm_providers/_multipass/_multipass.py
```diff
@@ -48,8 +47,6 @@ def _get_provider_name(cls):
         return "multipass"
 
     def run(self, command: Sequence[str], hide_output: bool = False) -> Optional[bytes]:
-        env_command = self._get_env_command()
```
knkski, line 51: This didn't look like it had any side effects, so seemed to just remove.

scripts/cluster/agent.py
```diff
@@ -9,6 +9,7 @@
 import subprocess
 import sys
 import time
+from typing import List
```
knkski, line 12: This is necessary due to `# type: List[str]` typing comment. It seems like we might want to just annotate `get_dqlite_voters()` instead.

scripts/cluster/join.py
```diff
@@ -7,6 +7,7 @@
 import getopt
 import sys
 import time
+from typing import Dict
```
knkski, line 10: This is necessary due to `# type: Dict[str, str]` typing comment. TBH, might make sense to just remove the typing comment instead. I at least don't get much out of seeing that `json()` would return a dict of strings.

tests/test-addons.py
```diff
@@ -142,10 +141,13 @@ def test_cilium(self):
         Sets up and validates Cilium.
         """
         print("Enabling Cilium")
-        p = Popen(
-            "/snap/bin/microk8s.enable cilium".split(), stdout=PIPE, stdin=PIPE, stderr=STDOUT
+        run(
```
knkski, line 144: Switching these two `Popen` calls to `run` seem equivalent, might someone else might want to double check on that.

tests/validators.py
```diff
@@ -378,7 +381,6 @@ def validate_linkerd():
         timeout_insec=300,
     )
     print("Linkerd proxy injector up and running.")
-    ### Disabling this test because the deletion of the namespace get stuck.
```
knkski, line 381: It looks like this test is not disabled, so just removed the comment instead of adjusting formatting.

microk8s-resources/actions/enable.kubeflow.sh
```diff
@@ -207,7 +207,7 @@ def main():
             args[arg] = strtobool(args[arg])
 
     with open("/proc/meminfo") as f:
-        memtotal_lines = [l for l in f.readlines() if "MemTotal" in l]
+        memtotal_lines = list(filter(lambda l: "MemTotal" in l, f.readlines()))
```
knkski, line 210: Kind of silly one is allowed and the other isn't

scripts/cluster/common/utils.py
```diff
@@ -20,7 +20,7 @@ def try_set_file_permissions(file):
     os.chmod(file, 0o660)
     try:
         shutil.chown(file, group='microk8s')
-    except:
+    except LookupError:
```
ktsakalozos, line 23: Why is this a LookupError? https://docs.python.org/3/library/shutil.html#shutil.chown

scripts/cluster/common/utils.py
```diff
@@ -169,7 +169,7 @@ def is_same_server(hostname, ip):
         hname, _, _ = socket.gethostbyaddr(ip)
         if hname == hostname:
             return True
-    except:
+    except socket.error:
```
ktsakalozos, line 172: Are we sure this will raise only a socket.error? https://docs.python.org/3/library/socket.html#socket-families

tests/validators.py
```diff
@@ -278,7 +281,7 @@ def validate_forward():
             resp = requests.get("http://localhost:5123")
             if resp.status_code == 200:
                 break
-        except:
+        except requests.RequestException:
```
ktsakalozos, line 284: Couldn't there be a ConnectionError https://requests.readthedocs.io/en/master/_modules/requests/exceptions/ ?

ktsakalozos:
@knkski I like this PR. Couldn't we not deal with the flow changes in the exception handling now? It is going to take some time to verify we got everything right.

tests/validators.py
knkski: `ConnectionError` inherits from `RequestException`, so catching `RequestException` should catch that as well, along with all the other exceptions in that page.

scripts/cluster/common/utils.py
knkski: The other exceptions that I see on that page (`socket.herror`, `socket.gaierror`, and `socket.timeout`) all subclasses from `OSError`, which `socker.error` is an alias of, so this should catch any errors that we care about.

scripts/cluster/common/utils.py
knkski: The group not existing will trigger a `LookupError` (based on testing this myself). The file not existing will create a `FileNotFoundError`. That should never happen though, since this function is only used in code that creates the file first. It would be useful to have `FileNotFoundError` then crash the program, since that would be an actual bug.

ktsakalozos (approved):
LGTM +1 thank you @knkski

PR #1488, opened 2020-08-13
Build K8s binaries in separate build step

knkski:
Since these don't rely on local source code changes other than what's under build-scripts/, if we build these in a separate snapcraft part, we won't have to rebuild them when making changes in most of the codebase. Building these binaries currently takes the majority of the time spent doing incremental rebuilds.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -161,6 +161,7 @@ parts:
       - libuv1
     build-packages:
       - libuv1-dev
+      - pkg-config
     organize:
       usr/lib/: lib/
       include/: usr/include/
@@ -226,6 +227,53 @@ parts:
       flanneld: opt/cni/bin/
     stage:
       - opt/cni/bin/flanneld
+
+  k8s-binaries:
+    after: [dqlite]
+    build-snaps: [go]
+    plugin: dump
+    build-attributes: [no-patchelf]
+    source: build-scripts/
+    build-packages:
+      - build-essential
+      - curl
+      - git
+    override-build: |
+      set -eux
+      . ./set-env-variables.sh
+
+      # if "${KUBE_SNAP_BINS}" exist we have to use the binaries from there
+      # if "${KUBE_SNAP_BINS}" does not exist but it is set we will put the k8s binaries there
+      # if "${KUBE_SNAP_BINS}" does not exist and it is not set we do not need to keep the created binaries
+      if [ ! -e "${KUBE_SNAP_BINS}" ]; then
+        if [ -z "${KUBE_SNAP_BINS}" ]; then
+          . ./set-env-binaries-location.sh
+        fi
+        echo "Downloading binaries"
+        . ./fetch-other-binaries.sh
+        echo "Building k8s binaries"
+        . ./build-k8s-binaries.sh
+      else
+        echo "Binaries provided in $KUBE_SNAP_BINS"
+      fi
+      mkdir bins/
+      cp build/kube_bins/$KUBERNETES_TAG/$KUBE_ARCH/* bins/
+
+      # Add bash completion for microk8s kubectl.
+      bins/kubectl completion bash | sed "s/complete -o default -F __start_kubectl kubectl/complete -o default -F __start_kubectl microk8s kubectl/g" | sed "s/complete -o default -o nospace -F __start_kubectl kubectl/complete -o default -o nospace -F __start_kubectl kubectl/g" > kubectl.bash
+      bins/kubectl completion bash | sed "s/complete -o default -F __start_kubectl kubectl/complete -o default -F __start_kubectl microk8s.kubectl/g" | sed "s/complete -o default -o nospace -F __start_kubectl kubectl/complete -o default -o nospace -F __start_kubectl kubectl/g" > kubectl.bash
+
+      snapcraftctl build
+    organize:
+      bins/*: .
+    stage:
+      - kube-apiserver
+      - kube-controller-manager
+      - kube-proxy
+      - kube-scheduler
+      - kubectl
+      - kubectl.bash
+      - kubelet
 
   libnftnl:
     plugin: autotools
@@ -342,7 +390,7 @@ parts:
     - python3-click
     - python3-dateutil
   microk8s:
-    after: [containerd, dqlite]
+    after: [containerd, dqlite, k8s-binaries]
     plugin: dump
     build-attributes: [no-patchelf]
     build-packages:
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -156,13 +156,30 @@ parts:
       - libuv1
     build-packages:
       - libuv1-dev
+      - pkg-config
```
knkski, line 159: Not sure why this works when building the full snap, but when only building the parts relevant to the K8s binaries, this package isn't installed.

snap/snapcraft.yaml
```diff
@@ -156,13 +156,30 @@ parts:
...
+    build-attributes: [no-patchelf]
+    source: build-scripts/
+    build-packages:
+      - git
+      - build-essential
+    override-build: |
+      set -eux
+      rm -rf go/
+      snapcraftctl build
+      bash wrapper.sh
```
knkski, line 181: Had to put this into a script explicitly run by bash instead of sh. Not sure how this section of script worked when it was sitting in the `microk8s` part, but sh can't do `. foo.sh` or `source foo.sh`, which this script does.

snap/snapcraft.yaml
```diff
@@ -301,26 +318,9 @@ parts:
...
-        fi
-        echo "Downloading binaries"
-        . build-scripts/fetch-other-binaries.sh
-        echo "Building k8s binaries"
-        . build-scripts/build-k8s-binaries.sh
-      else
-        echo "Binaries provided in $KUBE_SNAP_BINS"
-      fi
+      . build-scripts/set-env-variables.sh
+      KUBE_SNAP_BINS="/root/parts/k8s-binaries/build/build/kube_bins/$KUBE_VERSION"
```
knkski, line 323: It seems like some of this script should be moved over to the `k8s-binaries` part. Is there any reason not to move over the bits related to copying the K8s binaries into the snap (and maybe switching to an `organize` directive while we're at it)?

snap/snapcraft.yaml
```diff
@@ -425,9 +425,6 @@ parts:
       ./kubectl completion bash | sed "s/complete -o default -F __start_kubectl kubectl/complete -o default -F __start_kubectl microk8s kubectl/g" | sed "s/complete -o default -o nospace -F __start_kubectl kubectl/complete -o default -o nospace -F __start_kubectl kubectl/g" > kubectl.bash
       ./kubectl completion bash | sed "s/complete -o default -F __start_kubectl kubectl/complete -o default -F __start_kubectl microk8s.kubectl/g" | sed "s/complete -o default -o nospace -F __start_kubectl kubectl/complete -o default -o nospace -F __start_kubectl kubectl/g" > kubectl.bash
 
-      if $REMOVE_KUBE_SNAP_BINS; then
```
knkski, line 428: @ktsakalozos: i'm not sure what this code does exactly, but it relies on an env var that the new `wrapper.sh` script creates now in a different part. I removed it for now to make the build work, but is it still necessary?

snap/snapcraft.yaml
johnsca: `. foo.sh` is [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#dot) so does work in `sh`.  The path locations can get confusing during the snapcraft build process, especially when splitting things into separate parts, so if I had to guess, I think you might have run up against that.

build-scripts/wrapper.sh
```diff
@@ -0,0 +1,22 @@
...
+
+REMOVE_KUBE_SNAP_BINS=false
+echo "Reached here"
+# if "${KUBE_SNAP_BINS}" exist we have to use the binaries from there
+# if "${KUBE_SNAP_BINS}" does not exist but it is set we will put the k8s binaries there
+# if "${KUBE_SNAP_BINS}" does not exist and it is not set we do not need to keep the created binaries
+if [ ! -e "${KUBE_SNAP_BINS}" ]; then
+  if [ -z "${KUBE_SNAP_BINS}" ]; then
+    # "${KUBE_SNAP_BINS}" not set, we will remove the binaries after we build them.
+    REMOVE_KUBE_SNAP_BINS=true
```
ktsakalozos, line 13: I think we can remove this too.

snap/snapcraft.yaml
ktsakalozos: This absolute path does not look right. There must be a way to express this relative to one of the varibales in https://snapcraft.io/docs/parts-environment-variables and/or move the binary placement (eg `cp $KUBE_SNAP_BINS/$KUBE_ARCH/kube-apiserver .`) in the new part.

knkski: Good catch. I put that in to just make it work while I investigated moving these binaries into their own steps as well. I don't think we'll need it as soon as that can happen.

snap/snapcraft.yaml
ktsakalozos: I wouldn't mind placing the contents of the wrapper.sh here. We already run a couple of shell commands eitherway.

knkski: @johnsca: I am able to reproduce `sh` not understanding `.` in the shell:

```
# Bash works fine
snapcraft-microk8s # cd /root/parts/etcd/build
snapcraft-microk8s ../parts/etcd/build# ls
build-certs.sh	build-k8s-binaries.sh  fetch-k8s-binaries.sh  fetch-other-binaries.sh  patches	pre-patches  set-env-binaries-location.sh  set-env-variables.sh
snapcraft-microk8s ../parts/etcd/build# . set-env-variables.sh
 ... snip working output ...

# sh doesn't
snapcraft-microk8s ../parts/etcd/build# sh
\h ../parts/etcd/build# pwd
/root/parts/etcd/build
\h ../parts/etcd/build# ls
build-certs.sh	build-k8s-binaries.sh  fetch-k8s-binaries.sh  fetch-other-binaries.sh  patches	pre-patches  set-env-binaries-location.sh  set-env-variables.sh
\h ../parts/etcd/build# . set-env-variables.sh
sh: 3: .: set-env-variables.sh: not found
```

knkski: It turns out this is due to `dash` being the default for `sh`, and `dash` doesn't like `. foo.sh`. It instead wants `. ./foo.sh`. Still not really sure why it worked previously, other than guessing that some other package in another step switches `sh` to point at `/bin/bash`.

ktsakalozos:
@balchua I think you will like this PR.

Remember a few months ago when we were ended up to always rebuilding the k8s binaries?

balchua:
Oh yes!!!

snap/snapcraft.yaml
knkski: This line won't be necessary as soon as https://github.com/ubuntu/microk8s/pull/1490 gets merged

ktsakalozos:
LGTM +1 thank you @knkski

balchua:
@knkski thanks for this, builds are way faster now!

PR #1490, opened 2020-08-14
Move other binaries to separate snapcraft part

knkski:
Moves pulling in of a few extra non-k8s binaries to their own separate snapcraft steps. This will additionally speed up rebuilds when changing any files other than those under build-scripts/.


ktsakalozos (approved):
LGTM +1 thank you @knkski

Issue #1495, opened 2020-08-18
Multus may block microk8s refreshes


ktsakalozos:
In the configure hook if we detect the multus addon enabled we disable and re-enable it [1]. This operation has to be performed within 5 minutes or else the configure script will fail (see below). Do we need to disable/enable multus? @apnar, is there a better way to achieve this?

```
15:06:45 2020-08-18 at 12:06:45 | INFO error: cannot perform the following tasks:
15:06:45 2020-08-18 at 12:06:45 | INFO - Run configure hook of "microk8s" snap if present (run hook "configure":
15:06:45 2020-08-18 at 12:06:45 | INFO -----
15:06:45 2020-08-18 at 12:06:45 | INFO ++++ /snap/microk8s/1622/sbin/ip -o -4 addr list cni0
15:06:45 2020-08-18 at 12:06:45 | INFO ++++ /snap/microk8s/1622/usr/bin/gawk '{print $4}'
15:06:45 2020-08-18 at 12:06:45 | INFO ++++ /snap/microk8s/1622/usr/bin/cut -d/ -f1
15:06:45 2020-08-18 at 12:06:45 | INFO ++++ head -1
15:06:45 2020-08-18 at 12:06:45 | INFO +++ CNI_IP=10.1.80.1
15:06:45 2020-08-18 at 12:06:45 | INFO +++ local ips=
15:06:45 2020-08-18 at 12:06:45 | INFO +++ for ip in $IP_ADDR
15:06:45 2020-08-18 at 12:06:45 | INFO +++ '[' 172.17.0.1 '!=' 10.1.80.1 ']'
15:06:45 2020-08-18 at 12:06:45 | INFO +++ ips+=172.17.0.1
15:06:45 2020-08-18 at 12:06:45 | INFO +++ for ip in $IP_ADDR
15:06:45 2020-08-18 at 12:06:45 | INFO +++ '[' 10.1.80.1 '!=' 10.1.80.1 ']'
15:06:45 2020-08-18 at 12:06:45 | INFO +++ for ip in $IP_ADDR
15:06:45 2020-08-18 at 12:06:45 | INFO +++ '[' 10.236.94.121 '!=' 10.1.80.1 ']'
15:06:45 2020-08-18 at 12:06:45 | INFO +++ ips+=' 10.236.94.121'
15:06:45 2020-08-18 at 12:06:45 | INFO +++ for ip in $IP_ADDR
[...]

PR #1496, opened 2020-08-19
Do not re-enable multus on configure hook

ktsakalozos:
Unblock the release.

See https://github.com/ubuntu/microk8s/issues/1495


ktsakalozos:
@apnar I am removing the re-enabling of multus from the configure hook so as to unblock the release process.

Looking forward to your investigation if this approach has any defect.

Issue #1497, opened 2020-08-19
Metallb libssl_conf.so missing(?)


natdan:
On arm64 (aarch64) snap versions:
  latest/candidate: v1.18.8  2020-08-14 (1615) 187MB classic
  latest/edge:      v1.18.8  2020-08-19 (1625) 186MB classic

BTW latest/stable (v1.18.6) is fine...

```
$ microk8s enable metallb
Enabling MetalLB
Enter each IP address range delimited by comma (e.g. '10.64.140.43-10.64.140.49,192.168.0.105-192.168.0.111'): 192.168.5.10-192.168.5.99
Error configuring OpenSSL
548145813264:error:25066067:DSO support routines:DLFCN_LOAD:could not load the shared library:dso_dlfcn.c:187:filename(libssl_conf.so): libssl_conf.so: cannot open shared object file: No such file or directory
548145813264:error:25070067:DSO support routines:DSO_load:could not load the shared library:dso_lib.c:233:
548145813264:error:0E07506E:configuration file routines:MODULE_LOAD_DSO:error loading dso:conf_mod.c:271:module=ssl_conf, path=ssl_conf
548145813264:error:0E076071:configuration file routines:MODULE_RUN:unknown module name:conf_mod.c:212:module=ssl_conf
Applying Metallb manifest
namespace/metallb-system created
error: error validating "STDIN": error validating data: unknown object type "nil" in Secret.data.secretkey; if you choose to ignore these errors, turn validation off with --validate=false
Failed to enable metallb
```

ktsakalozos:
Thank you for following the candidate channel and for reporting this issue @natdan. What distribution is this on?

@balchua I think we missed the OPENSSL_CONF variable [1].

[1] https://github.com/ubuntu/microk8s/pull/985/files

natdan:
LOL - I was, kind of, avoiding to say immediately! O: )

It is on Raspberry Pi  64bit PiOS (formerly known as Raspbian)...

I'm looking forward giving it a go as soon as your PR goes through.
Thanks for the quick fix, too! :)

natdan:
Now I am confused. @ktsakalozos your PR is adding 
```
export OPENSSL_CONF="/snap/microk8s/current/etc/ssl/openssl.cnf"
```
but that line is already in the master! :( (and there are other changes)...

natdan:
[...]

PR #1498, opened 2020-08-19
Set openssl conf path on metallb

ktsakalozos:
Fixes: https://github.com/ubuntu/microk8s/issues/1497

### DIFF
--- a/microk8s-resources/actions/enable.metallb.sh
+++ b/microk8s-resources/actions/enable.metallb.sh
@@ -4,6 +4,8 @@ set -e
 
 source $SNAP/actions/common/utils.sh
 
+export OPENSSL_CONF="/snap/microk8s/current/etc/ssl/openssl.cnf"
+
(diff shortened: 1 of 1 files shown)


natdan:
Shouldn't change be 
```
export OPENSSL_CONF=$SNAP/etc/ssl/openssl.cnf
```
instead of using absolute path `/snap/microk8s/current`?

I understand that the same change was added on other two places as PR you linked before: (https://github.com/ubuntu/microk8s/pull/985/files) - but maybe even there it should use $SNAP env var.

ktsakalozos:
> export OPENSSL_CONF=$SNAP/etc/ssl/openssl.cnf

@natdan I probably agree with you. For this PR I will go with the old fix which I do not remember what made me use the full path.

We can investigate if the $SNAP will work and change it in a followup PR.

Thank you for reporting the issue and validating the fix.

Issue #824, opened 2019-11-25
Exposing Nginx-ingress to external traffic


moredhel:
I am wanting to expose my traffic to the internet,

When I enable ingress and define an ingress route, I have two services running which pick up on it:

- external-dns
- cert-manager

due to the way Nginx-ingress is configured, it does not pick up on the public ip of the running host, but instead defaults to `127.0.0.1`.

It would be useful to enable this as an option to allow serving traffic & automatically registering dns.

in order to do this, I copied the Nginx-ingress-controller yaml file in this repo, and updated the `--publish-status-address` to the public ip of my node.

I have found no documentation on this procedure, so was curious about whether or not this was a knob that had been intentionally hidden, and if so what the motivation behind it was.

Thanks

ktsakalozos:
@moredhel thank you for reporting this. I think it is worth at least document your approach. Could you offer a self contained example on exactly what you are doing? Something that I can reproduce? Could you share the yaml manifests you are applying, eg the rules and a demo service? Thank you.

moredhel:
@ktsakalozos sure thing,

I ended up cloning the microk8s repo & making the change there. The following can be run from the root directory of master:

```
sed -i 's/127.0.0.1/$(curl -4 ifconfig.io)/' microk8s-resources/actions/enable.ingress.sh
```

Then, to test:

1. enable ingress
2. install & configure external-dns
```
        helm upgrade \
                --version=2.10.2 \
                --namespace=kube-system \
                -i external-dns \
                -f values.yaml \
                stable/external-dns
```

sample _values.yaml_:
```
image:
  repository: bitnami/external-dns
  tag: 0.5.17-debian-9-r25

## The DNS provider where the DNS records will be created (options: aws, google, inmemory, azure )
provider: cloudflare

## This controls which types of resource external-dns should 'watch' for new
## DNS entries.
sources:
  - ingress

## RBAC parameteres
## https://kubernetes.io/docs/reference/access-authn-authz/rbac/
##
rbac:
  create: true


resources:
  limits:
    memory: 50Mi
  requests:
    memory: 50Mi
    cpu: 10m

cloudflare:
  proxied: false
  apiToken: "<><><><>"
  email: "cloudflare@example.com"

## Modify how DNS records are sychronized between sources and providers (options: sync, upsert-only )
policy: upsert-only

## Prometheus Exporter / Metrics
##
metrics:
  enabled: true
```
[...]

PR #1499, opened 2020-08-19
Configure nginx ingress with host IP

panlinux:
If possible, use the host IP when setting the nginx ingress "public" IP.

Fixes issue #824

### DIFF
--- a/microk8s-resources/actions/enable.ingress.sh
+++ b/microk8s-resources/actions/enable.ingress.sh
@@ -23,7 +23,11 @@ echo "Enabling Ingress"
 
 ARCH=$(arch)
 TAG="0.33.0"
-EXTRA_ARGS="- --publish-status-address=127.0.0.1"
+HOST_IP=$(get_default_ip)
+if ! valid_ip "${HOST_IP}"; then
+  HOST_IP="127.0.0.1"
+fi
+EXTRA_ARGS="- --publish-status-address=${HOST_IP}"
 DEFAULT_CERT="- ' '"
 
 if [ ! -z "$CERT_SECRET" ]


ktsakalozos:
Thank you for this PR @panlinux. Looks great!

As the 1.19 release is scheduled for next week I am going to hold off this PR for a bit.

I appreciate your patience. Cheers.

panlinux:
Oh, I thought the edge channel would have something like a daily build, exactly to test changes like these, but looks like it's behind.

ktsakalozos:
> to test changes like these

You can get the built snap as an artifact of the github workflow/test.

panlinux:
ok, built the arm64 snap with this change, and the other one suggested in https://github.com/ubuntu/microk8s/issues/824#issuecomment-688491339, but still need to test them.
This one here I did a quick deploy and saw the public address being the one from the host. What I would like to do next is change the ip of the host, and see what happens after a reboot. I suspect it will break things, which is where the second suggested change should be better, but I have to check.

ktsakalozos:
We also need to figure out how we will going to handle the change of the default IP (eg a laptop changing networks).

panlinux:
The pod would have to be restarted I guess.
Stupid question: would this work?
```
--publish-status-address=0.0.0.0
```

patsevanton:
Can i help for resolve this PR?

amirhs712:
Hey, can we resolve this PR, please?

typoworx-de:
IP address should be configureable and fallback should be Host-IP or 127.0.0.1.

jniclas:
Hey, I had the problem too now by trying to use ingress with external-dns, where it picket the 127.0.01. Is it posibble to still integrate that?

Issue #1347, opened 2020-06-23
InvalidDiskCapacity in a fresh install of microk8s on Ubuntu 20.04 on Raspberry Pi 4 8GB 


nuinattapon:
Please run `microk8s inspect` and attach the generated tarball to this issue.

We appreciate your feedback. Thank you for using microk8s.
I install microk8s fresh in my new Raspberry PI 4 8GB model today on top of UBUNTU 20.04LTS.

And I found that my K8s node is always not ready because of "invalid capacity 0 on image filesystem". 
I believe that this is a bug. Pls advise.

`NAME     STATUS     ROLES    AGE   VERSION   INTERNAL-IP     EXTERNAL-IP   OS-IMAGE           KERNEL-VERSION     CONTAINER-RUNTIME
ubuntu   NotReady   <none>   13m   v1.18.3   192.168.1.110   <none>        Ubuntu 20.04 LTS   5.4.0-1012-raspi   containerd://1.2.5
df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            3.8G     0  3.8G   0% /dev
tmpfs           782M  4.4M  777M   1% /run
/dev/mmcblk0p2   29G  8.2G   20G  30% /
tmpfs           3.9G   29M  3.8G   1% /dev/shm
tmpfs           5.0M  4.0K  5.0M   1% /run/lock
tmpfs           3.9G     0  3.9G   0% /sys/fs/cgroup
/dev/loop0       86M   86M     0 100% /snap/core/9290
/dev/loop1       49M   49M     0 100% /snap/core18/1708
/dev/loop3       24M   24M     0 100% /snap/snapd/7267
/dev/loop4       62M   62M     0 100% /snap/lxd/14808
/dev/loop7       49M   49M     0 100% /snap/core18/1753
/dev/loop6       26M   26M     0 100% /snap/snapd/8147
/dev/loop5       64M   64M     0 100% /snap/lxd/15685
/dev/mmcblk0p1  253M  107M  146M  43% /boot/firmware
tmpfs           782M   20K  782M   1% /run/user/1000
/dev/loop2      165M  165M     0 100% /snap/microk8s/1423`

[k_describe_node.txt](https://github.com/ubuntu/microk8s/files/4820171/k_describe_node.txt)
[inspection-report-20200623_214123.tar.gz](https://github.com/ubuntu/microk8s/files/4820177/inspection-report-20200623_214123.tar.gz)

ktsakalozos:
[...]

PR #1501, opened 2020-08-19
Add memory cgroup check

exceptorr:
Adding an additional check to ensure that memory cgroup is explicitly enabled if running on ARM architecture (RPi or any other), as it's not enabled on ARM systems by default (this is explicitly noted in the microk8s docs: https://microk8s.io/docs/install-alternatives#heading--arm and causing issues like https://github.com/ubuntu/microk8s/issues/1347)


ktsakalozos (approved):
LGTM +1 thank you @exceptorr

PR #1502, opened 2020-08-20
configure multus to use symlinked current SNAP_DATA locations

apnar:
Fixes issue #1495.

Changes the multus daemon set to use the "current" path symlinks for its mounts and arguments.  This should allow it to survive refreshes and upgrades without needing to be disabled/re-enabled.


ktsakalozos (approved):
LGTM +1 thank you @apnar for the PR and for being available in such a short notice. Really appreciate this.

PR #1509, opened 2020-08-22
Apply cni.yaml in nodes with the apiserver

ktsakalozos:
When refreshing a non HA node or a node with the apiserver not running we should not apply the cni.yaml


ktsakalozos:
No needed.

Issue #784, opened 2019-11-08
knative HelloWorld Serving Code Example


datadot:
Deploy a clean microk8s snap deployment:

`snap install microk8s --current`

Enable DNS, Istio + kNative

`sudo microk8s.enable dns istio knative`

Deploy HelloWorld go service example from:

```
apiVersion: serving.knative.dev/v1alpha1 # Current version of Knative
kind: Service
metadata:
  name: helloworld-go # The name of the app
  namespace: default # The namespace the app will use
spec:
  template:
    spec:
      containers:
        - image: gcr.io/knative-samples/helloworld-go # The URL to the image of the app
          env:
            - name: TARGET # The environment variable printed out by the sample app
              value: "Go Sample v1"
```
The pods will create and fail due to below error:

`Readiness probe errored: rpc error: code = Unknown desc = failed to exec in container: failed to start exec "fc02dba843b84f907e3054501f078791474de71dce1d68e37734af3ef30fcf22": OCI runtime exec failed: exec failed: container_linux.go:345: starting container process caused "apparmor failed to apply profile: write /proc/self/attr/exec: operation not permitted": unknown`

Result from kubectl get ksvc is "RevisionMissing".

Digging further but has anyone else experienced this?

datadot:
Ok I have done some further testing, this works with clean 18.04.3 LTS (bionic) however issue exists in 19.10 (eoan).

rbt:
I'm getting a similar error attempting to exec -ti into a pod, something I did regularly during development while under minikube (attempting to switch). The pod has the status of "Running" and has a single container.

```
/snap/bin/microk8s.kubectl exec -t -i web-6d64dd86f-dzb6q  -- bash
error: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec "78186bf219f66adc86f6255eb98bbbe4683f6875c58022feb6cdced38b30f594": OCI runtime exec failed: exec failed: container_linux.go:345: starting container process caused "apparmor failed to apply profile: write /proc/self/attr/exec: operation not permitted": unknown
```

ktsakalozos:
@rbt can you please attach the tarball from `microk8s.inspect`. I would like to see the error in `dmesg` you are probably seeing.

rbt:
1.16/stable: [inspection-report-20191111_173728.tar.gz](https://github.com/ubuntu/microk8s/files/3833486/inspection-report-20191111_173728.tar.gz)

[...]

PR #1515, opened 2020-08-26
Apply runc patch to fix nnp issues

ktsakalozos:
Fixes: https://github.com/ubuntu/microk8s/issues/784

### DIFF
--- a/build-scripts/build-k8s-binaries.sh
+++ b/build-scripts/build-k8s-binaries.sh
@@ -25,7 +25,7 @@ git clone --depth 1 https://github.com/kubernetes/kubernetes $GOPATH/src/github.
     PATCHES="pre-patches"
   fi
 
-  for patch in "${SNAPCRAFT_PROJECT_DIR}"/build-scripts/"$PATCHES"/*.patch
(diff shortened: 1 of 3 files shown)


ktsakalozos:
@joedborg @jdstrand while investigating the issues @giner reported on this patch I did a rebase to see how this PR goes with v1.19. Unfortunately, calico will not play well. Although the calico-node comes up, the calico-controller (and any other pod I tried) is crashlooping. This may be related to the strict confinement work. The `latest/edge/runc-nnp` branch is updated with the failing snap if you want to have a look.

jdstrand:
Not applying this is problematic and I strongly suggest that the issues with calico be looked at more closely before NAKing this. containerd is not working correctly today and this patch makes it work right for many cases. Without applying this patch, https://github.com/ubuntu/microk8s/issues/784 remains. Furthermore, as mentioned in https://github.com/ubuntu/microk8s/issues/784#issuecomment-675714924, this patch is needed for strict *and* classic because without it, containerd isn't always going to operate as intended and it won't be able to apply security restrictions in the manner it is being told to do so.

I unfortunately don't have the time to look at this now, but have confirmed with @joedborg that calico is not working correctly in strict and devmode. I think someone needs to look at strace/etc to determine where things are failing if this is failing with classic.

PR #1529, opened 2020-08-31
Adding start wrapper

joedborg:
Not sure how this flew under the radar for so long. 

`start` now added to the multi-OS wrapper and the VM will be re-launched if someone tries to run a command on a stopped cluster.


ktsakalozos (approved):
LGTM +1 thank you @joedborg

PR #1533, opened 2020-09-01
Bump Multipass download to 1.4.0

joedborg:
Tested on the dev laptop and sent to IS for signing.


ktsakalozos (approved):
LGTM +1 thank you @joedborg

PR #1535, opened 2020-09-02
Fix upgrade from non-HA to HA

ktsakalozos:
This PR addresses a few bugs on the transition of a cluster from non-HA to HA.

- If one starts an HA transition and never finishes it (calls `microk8s.enable ha-cluster` on the master node but does not transition the nodes to HA) any refreshes on the nodes will fail if they are following a 1.19+ track.
- The upgrade process was not triggered on all nodes.
- If the node we call `microk8s enable ha-cluster` was already on a non-HA cluster that gets upgraded to HA it may have etcd disabled because it may have already transitioned to dqlite. In this case we should not try to take a backup of etcd.

The workaround for all the above is after we remove the node from the cluster via `microk8s.leave` we `snap remove microk8s` and reinstall so we join a clean node.

### DIFF
--- a/scripts/wrappers/upgrade.py
+++ b/scripts/wrappers/upgrade.py
@@ -142,7 +142,7 @@ def get_nodes_info(safe=True):
                         if host not in nodes.decode():
                             print("Node {} not present".format(host))
                             continue
-                        node_info = [(parts[0], parts[1])]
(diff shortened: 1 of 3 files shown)


ktsakalozos:
I am merging this @joedborg so we give it a good test. Please review when you have time.

PR #1537, opened 2020-09-02
Feature/installer/exit codes

joedborg:
To be merged after https://github.com/canonical/multipass-private/pull/259 is.

This scraps the use of the registry for leaving callbacks to the MicroK8s installer and replaces it with exit codes.

Tested with the built Multipass branch.


ktsakalozos:
Looks ok to me. But was not able to verify the affected paths.

PR #1555, opened 2020-09-11
Check current UID and permissions before enabling Kubeflow

### LINKED ISSUE #1396: Juju file permissions after adding user to microk8s group
If a user enables kubeflow with sudo, juju creates state files as the root user. If then the user adds himself to the microk8s group enabling kubeflow is failing with:
```
$ microk8s.enable kubeflow
Enabling dns...
Enabling storage...
Enabling dashboard...
Enabling ingress...
Enabling metallb:10.64.140.43-10.64.140.49...
Waiting for DNS and storage plugins to finish setting up
Deploying Kubeflow...
Kubeflow could not be enabled:
ERROR cannot load ssh client keys: open /var/snap/microk8s/1503/juju/share/juju/ssh: permission denied

Command '('microk8s-juju.wrapper', 'bootstrap', 'microk8s', 'uk8s')' returned non-zero exit status 2
Failed to enable kubeflow
```

To reproduce try:
```
sudo microk8s enable kubeflow
sudo microk8s disable kubeflow
sudo usermod -a -G microk8s $USER"
newgrp microk8s
microk8s enable kubeflow
```

A command like `sudo chown -R  $USER /var/snap/microk8s/current/juju/share/juju/` could work but the safest approach for now is to always `sudo microk8s enable kubeflow`.

knkski:
Disallows enabling Kubeflow as root, and also attempts to fix permissions issues caused by previously attempting to enable Kubeflow as root. If the permissions fix fails, prints out message instructing user how to do it themselves.

Fixes #1396

### DIFF
--- a/microk8s-resources/actions/enable.kubeflow.sh
+++ b/microk8s-resources/actions/enable.kubeflow.sh
@@ -12,8 +12,9 @@ import textwrap
 import time
 from distutils.util import strtobool
 from itertools import count
+from pathlib import Path
 from urllib.error import URLError
-from urllib.parse import urlparse, ParseResult
+from urllib.parse import ParseResult, urlparse
 from urllib.request import urlopen
 
 MIN_MEM_GB = 14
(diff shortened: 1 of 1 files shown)


ktsakalozos (approved):
LGTM +1

betsegaw:
@knkski and @ktsakalozos, what is the expected workflow for users who only have the root user?

PR #1562, opened 2020-09-14
Check machine hostname during inspection

ktsakalozos:
Check if the hostname does not container capital letters or underscores.


evilnick (approved):
:cake:

PR #1579, opened 2020-09-18
Run containerd unconfined

ktsakalozos:
Running containerd unconfined allows us to handle AllowPrivilegeEscallation flag correctly.

Fixes https://github.com/ubuntu/microk8s/issues/784

### DIFF
--- a/microk8s-resources/wrappers/run-containerd-with-args
+++ b/microk8s-resources/wrappers/run-containerd-with-args
@@ -2,6 +2,11 @@
 
 set -ex
 
+# Re-exec outside of apparmor confinement
+if [ -d /sys/kernel/security/apparmor ] && [ "$(cat /proc/self/attr/current)" != "unconfined" ]; then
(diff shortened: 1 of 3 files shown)


microk8s-resources/wrappers/run-containerd-with-args
```diff
@@ -2,6 +2,11 @@
 
 set -ex
 
+# Re-exec outside of apparmor confinement
+if [ -d /sys/kernel/security/apparmor ] && [ "$(cat /proc/self/attr/current)" != "unconfined" ]; then
+    exec aa-exec -p unconfined -- "$0" "$@"
+fi
```
jdstrand, line 8: Note, this is not a proper solution for strict mode.

ktsakalozos: Yes, for strict confinement we need containerd confined.

joedborg (approved):

PR #1582, opened 2020-09-19
fix the kubepublic name to kube-public

balchua:
api-service-kicker check for the local-registry if installed in `kube-public`


ktsakalozos (approved):
LGTM +1 thank you @balchua

PR #1583, opened 2020-09-19
upgrade cilium to version 1.8.3

balchua:


balchua:
@joestringer here's updated cilium.  I've decided to use the same pattern of setup as before.  Except reduce the operator replicas to 1.  Since on single nodes the other one will not come up due to anti affinity.
When you have time, your review is most appreciated.

microk8s-resources/actions/enable.cilium.sh
```diff
@@ -73,6 +74,7 @@ else
       --set global.cni.customConf=true \
       --set global.containerRuntime.integration="containerd" \
       --set global.containerRuntime.socketPath="$SNAP_COMMON/run/containerd.sock" \
```
joestringer, line 76: You should also be able to do `--set global.cni.confPath=$SNAP_DATA` above here, and then the `mkdir`, `cp`, `sed` below can be removed.

joestringer (approved):
LGTM. I posted a note below about the other change you can make to help clean up the file.

microk8s-resources/actions/enable.cilium.sh
balchua: Thanks for the review. That setting is already there. The commands below is just to move the `cni.yaml` to the right place and not the tmp.

joestringer: My point is that they shouldn't be necessary if we just write the config correctly from Helm the first time :-) Originally when I introduced this, the v1.6 YAMLs didn't include an option to configure this option so this is why the script does the copy/sed tricks below.

Any way, this is an unrelated cleanup, up to you on whether you think it makes sense to change it.

balchua: Oh i know what you meant now. 😁 Sorry kindda slow. Haha thanks for the pointers.  Will update the pr.

microk8s-resources/actions/enable.cilium.sh
```diff
@@ -36,7 +37,7 @@ echo "Enabling Cilium"
 
 read -ra CILIUM_VERSION <<< "$1"
 if [ -z "$CILIUM_VERSION" ]; then
-  CILIUM_VERSION="v1.6"
+  CILIUM_VERSION="v1.8.3"
```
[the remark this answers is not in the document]
balchua: Cool!! Will be defaulting it to `1.8`. Also realized that when enabling cilium, one can choose a version. It wasn't so obvious as it is not indicated in the docs.

balchua:
Travis failure due to timeout.  Github action build succeeds.

ktsakalozos:
This update will come with 1.20. I am holding back the review/merge until 1.19 gets branched.

ktsakalozos (approved):
LGTM +1 thank you @balchua for the work and @joestringer for the review.

ktsakalozos:
I am afraid I have to revert this. The upgrade process (from 1.6 to 1.8) is failing.

If you deploy the latest/beta snap (`sudo snap install microk8s --classic --channel=latest/beta) , enable cilium (`sudo microk8s enable cilium`) you will not be able to refresh to a snap revision with the new cilium (`sudo snap install ./microk8s*.snap --classic --dangerous`). The error I get is:
```
+ '[' -L /var/snap/microk8s/1832/bin/cilium ']'
+ echo 'Cilium is enabled we need to reconfigure it.'
Cilium is enabled we need to reconfigure it.
+ rm -rf /var/snap/microk8s/1832/bin/cilium /var/snap/microk8s/1832/bin/cilium-1.6
+ /snap/microk8s/1832/actions/enable.cilium.sh
Addon helm3 is already enabled.
Restarting kube-apiserver
Restarting kubelet
Deleting old cilium_vxlan link
Enabling Cilium
Fetching cilium version v1.8.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   119  100   119    0     0    514      0 --:--:-- --:--:-- --:--:--   515
100 27.3M    0 27.3M    0     0  7400k      0 --:--:--  0:00:03 --:--:-- 7562k
Deploying /var/snap/microk8s/1832/actions/cilium.yaml. This may take several minutes.
configmap/cilium-config configured
serviceaccount/cilium unchanged
serviceaccount/cilium-operator unchanged
clusterrole.rbac.authorization.k8s.io/cilium configured
clusterrole.rbac.authorization.k8s.io/cilium-operator configured
clusterrolebinding.rbac.authorization.k8s.io/cilium unchanged
clusterrolebinding.rbac.authorization.k8s.io/cilium-operator unchanged
deployment.apps/cilium-operator configured
The DaemonSet "cilium" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"k8s-app":"cilium"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
-----)
```

It is caused by this line https://github.com/ubuntu/microk8s/blob/master/snap/hooks/configure#L380
During a snap refresh we try to re-enable the cilium addon. This means the cilium enable script should take care of the upgrade process or maybe not upgrade.

I suspect it to be an easy fix.

PR #1599, opened 2020-09-26
fix disabling prometheus.

### LINKED ISSUE #1590: Failed to disable prometheus: error: the path "/var/snap/microk8s/1668/kube-prometheus/manifests/" does not exist
after upgrading microk8s prometheus somehow is not possible to disable anymore

$ microk8s disable prometheus
Disabling old Prometheus
Disabling Prometheus
error: the path "/var/snap/microk8s/1668/kube-prometheus/manifests/" does not exist
Failed to disable prometheus

balchua:
Fixes #1590 

Add back the old kube-prometheus so it can be disable when refreshed to 1.19.  Also separate the prometheus addon validation, so it can be disabled individually.


ktsakalozos (approved):
LGTM +1 thank you @balchua

Issue #1575, opened 2020-09-16
microk8s.helm3 completion bash isn't working


vijaynagbs:
Please run `microk8s inspect` and attach the generated tarball to this issue.
[inspection-report-20200916_172134.tar.gz](https://github.com/ubuntu/microk8s/files/5231947/inspection-report-20200916_172134.tar.gz)

`source <(microk8s.helm3 completion bash)` 
executing the above command has no effect. 

`microk8s.helm3 version`
version.BuildInfo{Version:"v3.0.2", GitCommit:"19e47ee3283ae98139d98460de796c1be1e3975f", GitTreeState:"clean", GoVersion:"go1.13.5"}

balchua:
The `microk8s helm3` command is a wrapper.  So i usually do this.
```shell
source <(microk8s helm3 completion bash | sed "s/complete -o default -F __start_helm helm/complete -o default -F __start_helm microk8s helm3/g")
```

PR #1610, opened 2020-09-30
Add Helm3 completion script #1575

johnp41:
addition of completion script for helm3


balchua:
@johnp41 many thanks for this PR.  I suppose there are changes needed on the `snapcraft.yaml`?

PR #1625, opened 2020-10-05
Add resolv.conf to forward list in coredns

### LINKED ISSUE #1606: [coredns] default to host configuration not google public
Deploying microk8s on an environment that does not allow access to public domain servers I discovered that the DNS add on is deploying coredns configured with google dns servers hard coded as the default forwarders. It would seem a better option would be to use the hosts DNS configuration rather than a public server that may or may not be allowed on the system.

Referencing https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/#coredns this default is what I as a user had expected from coredns. While I can edit the configmap to make changes myself, for a better first time user experience consider using a configuration like this by default.

RiyaJohn:
Fixes #1606

### DIFF
--- a/microk8s-resources/actions/coredns.yaml
+++ b/microk8s-resources/actions/coredns.yaml
@@ -32,7 +32,7 @@ data:
           fallthrough in-addr.arpa ip6.arpa
         }
         prometheus :9153
-        forward . 8.8.8.8 8.8.4.4 /etc/resolv.conf
+        forward . {{nameservers}}
         cache 30
         loop
         reload
--- a/microk8s-resources/actions/coredns.yaml
+++ b/microk8s-resources/actions/coredns.yaml
@@ -32,7 +32,7 @@ data:
           fallthrough in-addr.arpa ip6.arpa
         }
         prometheus :9153
-        forward . {{nameservers}}
+        forward . $NAMESERVERS
         cache 30
         loop
         reload
--- a/microk8s-resources/actions/coredns.yaml
+++ b/microk8s-resources/actions/coredns.yaml
@@ -32,7 +32,7 @@ data:
           fallthrough in-addr.arpa ip6.arpa
         }
         prometheus :9153
-        forward . 8.8.8.8 8.8.4.4
+        forward . 8.8.8.8 8.8.4.4 /etc/resolv.conf
         cache 30
         loop
         reload
--- a/microk8s-resources/actions/enable.dns.sh
+++ b/microk8s-resources/actions/enable.dns.sh
@@ -7,16 +7,6 @@ KUBECTL="$SNAP/kubectl --kubeconfig=${SNAP_DATA}/credentials/client.config"
 # Apply the dns yaml
 # We do not need to see dns pods running at this point just give some slack
 echo "Enabling DNS"
-echo "Applying manifest"
-ALLOWESCALATION="false"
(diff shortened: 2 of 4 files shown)


balchua:
Thanks @RiyaJohn for the PR.  Really appreciate the contribution.

In this particular feature, I am imagining more to be like this.

When a user wanted to enable dns, we give them option what upstream dns to use.  For example: `microk8s enable dns:8.8.8.8,8.8.4.4` or using cloudflare `microk8s enable dns:1.1.1.1`.
If the user didn't specify an upstream dns it defaults to `/etc/resolv.conf`.

There are ways to extract the "parameters" when enabling an addon. As an example `metallb` addon, wherein the user can specify the IP ranges to use when enabling it.  Such as `microk8s enable metallb:192.168.2.0/16`.
Wdyt?

You can have a look at the implementation [here](https://github.com/ubuntu/microk8s/blob/c6b23f587043fc3a33709e4c672de171830e2b4c/microk8s-resources/actions/enable.metallb.sh#L25).

To build MicroK8s here's the docs https://github.com/ubuntu/microk8s/blob/master/docs/build.md

Thanks for the contribution!! 😊

RiyaJohn:
Not sure if an option to add upstream dns is really a use-case/ customer requirement, since I am not really into the container world. But if its a use-case then yes its a good idea to add that option. Thanks to your links, I will work on adding the option

microk8s-resources/actions/coredns.yaml
```diff
@@ -32,7 +32,7 @@ data:
           fallthrough in-addr.arpa ip6.arpa
         }
         prometheus :9153
-        forward . 8.8.8.8 8.8.4.4
+        forward . 8.8.8.8 8.8.4.4 /etc/resolv.conf
```
chris-sanders, line 35: You can't add resolvers that produce different results like this. 
Ref: https://coredns.io/plugins/forward/

The default policy for forward is random, you're going to mix resolve.conf with 8.8.8.8 which would be bad. I think you're better off sticking to just resolv.conf and accepting an argument to replace it with other resolvers.

chris-sanders (changes requested):
Just a comment on the implementation, you're unintentionally mixing resolvers which I'm sure you don't want.

microk8s-resources/actions/coredns.yaml
RiyaJohn: Thank you, yes this will change to add an option for user to pass dns else resolv.conf will be picked by default, like @balchua commented https://github.com/ubuntu/microk8s/pull/1625#issuecomment-703910093 suggested

RiyaJohn:
@balchua Can you take a look at this PR

balchua:
Thanks @RiyaJohn LGTM 👍

RiyaJohn:
@ktsakalozos can you review this and merge it if it's good to go

RiyaJohn:
When I tried to resolve conflicts, I see that the build and CI fails, doesn't look like its because of my code change. But let me know if I am missing anything

ktsakalozos:
Thank you for the PR @RiyaJohn and apologies for the late review/merge.

As a last minute change we decided not to change the default forward DNS servers but to allow users to set `/ets/resolv.conf` if they wish to. The reason is mostly because it is easier to spot and debug cases where you do not reach the public DNS servers than to spot error in local host configurations.

Issue #719, opened 2019-10-13
Microk8s on armhf architecture


CalvinHartwell:
Hi all, 

The armhf binary is missing and not available right now, which means that some users cannot install microk8s on Ubuntu. For example, if you use the armhf image for Raspberry Pi, you cannot install microk8s: 

> ubuntu@battlecruiser:~$ sudo snap install microk8s --classic
> error: snap "microk8s" is not available on stable for this architecture (armhf)
>        but exists on other architectures (amd64, arm64, ppc64el).

It would be really good if we could also get the build compiled for this architecture and make officially available. 

Cheers,

- Calvin

recklessop:
this is a duplicate issue of https://github.com/ubuntu/microk8s/issues/228

CalvinHartwell:
@recklessop you're correct yes, my apologies. I have hardware available we could use to address the problem.

iam-TJ:
Does the detail at:

https://microk8s.io/docs/install-alternatives#arm

"... This can be remedied on the Rasberry Pi by editing the boot parameters: ..."

imply that  this is now solved?

ktsakalozos:
> imply that this is now solved?

No the issue is not resolved. The instructions are for arm64 not for armhf. We currently do not build for this architecture. You can try to create a build yourself by git cloneing the repo and running  `snapcraft` possibly in a remote build as described here https://forum.snapcraft.io/t/preview-snapcraft-remote-build/11541 Note that we have not tested this so any patches you can provide are welcomed.

bragdonjm:
Running into this now myself.

stevieray8450:
running into this is as well.

I've tried using `snapcraft remote-build` as well. I get this error:

> Remote-build does not support the use of `version-script`.
> Please use `snapcraftctl set-version` part scriptlet with `adopt-info`, or set `version` statically.

erasmus74:
Also running into this, dietpi on RPI4 all recent builds as of this date

ktsakalozos:
@erasmus74 I am fairly certain RPi4 is on ARM64. Could you double check the OS you have installed?

erasmus74:
You're correct, I was on armhf/32-bit, switching to 64bit worked like a charm

forzagreen:
Same issue on Raspberry PI 3 B.
I ended up using [k3s](https://k3s.io/).

angelnu:
[...]

PR #1632, opened 2020-10-08
Try s390x build with Ubuntu's etcd

xnox:
etcd upstream doesn't build/provide s390x builds.

I've modified the part, to download/unpack/use Ubuntu's ectd to create s390x snap of microk8s. It built, but I don't know if it works at all. See the snap at https://code.launchpad.net/~xnox/+snap/xnox-microk8s-test/+build/1142929

Maybe we can cross-build exact same etcd release as a part on Ubuntu? Or is slightly different etcd ok for microk8s?

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -200,9 +200,22 @@ parts:
     plugin: dump
     source: build-scripts/
     override-build: |
-      . ./set-env-variables.sh
-      curl -LO https://github.com/etcd-io/etcd/releases/download/${ETCD_VERSION}/etcd-${ETCD_VERSION}-linux-$KUBE_ARCH.tar.gz
-      tar -xzvf etcd-*.tar.gz --strip-components=1
+      set -eu
+      ARCH=$(dpkg --print-architecture)
+      # No upstream etcd s390x builds yet
+      if [ "${ARCH}" = "s390x" ]
+      then
+        curl -LO https://launchpad.net/ubuntu/+source/etcd/3.2.26+dfsg-8/+build/19387496/+files/etcd-client_3.2.26+dfsg-8_s390x.deb
+        curl -LO https://launchpad.net/ubuntu/+source/etcd/3.2.26+dfsg-8/+build/19387496/+files/etcd-server_3.2.26+dfsg-8_s390x.deb
+        dpkg-deb -x etcd-client_3.2.26+dfsg-8_s390x.deb etcd-client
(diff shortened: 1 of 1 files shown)


ktsakalozos:
Hi @xnox, thank you for this PR.

I would much prefer to have a part similar to [1] as the etcd snap is maintained by our team or use the etcd snap as a stage snap [2]. We would also need to find a s390x machine to test this before merging it.

[1] https://github.com/juju-solutions/etcd-snaps/blob/v3.4/snap/snapcraft.yaml
[2] https://forum.snapcraft.io/t/stage-snaps/11001

xnox:
> Hi @xnox, thank you for this PR.
> 
> I would much prefer to have a part similar to [1] as the etcd snap is maintained by our team or use the etcd snap as a stage snap [2]. We would also need to find a s390x machine to test this before merging it.
> 
> [1] https://github.com/juju-solutions/etcd-snaps/blob/v3.4/snap/snapcraft.yaml
> [2] https://forum.snapcraft.io/t/stage-snaps/11001

I like stage snap. Would you want that for all arches? or just s390x (and the rest continue to reuse upstream builds from github)?

xnox:
s390x machine access is easy for me, i just don't know what to test.

ktsakalozos:
> I like stage snap. Would you want that for all arches? or just s390x

Lets take one step at a time and use the stage-snap only for s390x.



> s390x machine access is easy for me, i just don't know what to test.

As soon as you build the snap you should be able to `snap install ./yournap.snap --classic --dangerous` and you should be able to call `microk8s.kubectl get all -A`. There you should see the calico CNI crashing because the calico docker image does not exist for s390x. This is the next step!

angelnu:
@xnox - I am working in support for armhf so my work at #719 might help - there I am building etcd on not supported platforms so it matches the exact version used by microk8s.

xnox:
s390x is available in travis too, if you have that connected for CI here (at least on s390x).

ktsakalozos:
@xnox we merged a PR to build etcd from source for architecture where prebuild binaries do not exist. Building from source has the benefit of not pining us to a specific version. Does building from source work for the s390x case?

xnox:
> @xnox we merged a PR to build etcd from source for architecture where prebuild binaries do not exist. Building from source has the benefit of not pining us to a specific version. Does building from source work for the s390x case?

YES!

ktsakalozos:
> YES!

Great, I am closing this PR then.

PR #1635, opened 2020-10-09
Test Kubeflow addon in Github actions

knkski:


knkski:
@ktsakalozos: Currently failing due to the apiserver restarting while enabling addons. That seems a lot like the troubles we were having with dqlite causing the apiserver to restart.

ktsakalozos (approved):
LGTM +1 thank you @knkski

PR #1641, opened 2020-10-12
Fix help option for inspect command

bowers:
_microk8s inspect --help_   doesn't invoke the help function.  The arguments aren't passed to the inspect script by the microk8s wrapper.


ktsakalozos (approved):
Thank you @bowers LGTM +1

PR #1650, opened 2020-10-15
for portainer add-on

balasu:


tests/test-addons.py
```diff
@@ -324,6 +325,16 @@ def test_multus(self):
...
+    
+    def test_portainer(self):
+        """
+        Sets up and validates Portainer.
+        """
+        print("Enabling Portainer")
+        microk8s_enable("portainer")
+        print("Validating Portainer")
+        validate_portainer()
+        print("Disabling Portainer")
```
ktsakalozos, line 337: I think you are missing the disable command here.

ktsakalozos (changes requested):
Thank you for this PR @balasu . You will also need to add your addon at https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/addon-lists.yaml

balchua (commented):
We may want to keep `$ARCH` here `cdkbot/microbot-amd64`, i think the team ran the tests on multiple arch too.

ktsakalozos (approved):
LGTM +1 thanky ou for the PR @balasu

PR #1655, opened 2020-10-16
apiservice-kicker: fix a race condition

nicks:
before this change, the apiservice-kicker would sometime reset the registry configmap even when the registry existed


nicks:
thanks @ktsakalozos ! :pray:

PR #1687, opened 2020-10-26
for enabling traefik ingress controller

balasu:
Hi
This is for enabling traefik as an ingresscontroller running on 8080 on traefik namespace.


balchua:
This PR contains portainer addons. I think it is best to keep them separate.  Since you already have another PR for portainer.

ktsakalozos (approved):
LGTM +1 Thank you @balasu

microk8s-resources/actions/traefik.yaml
```diff
@@ -0,0 +1,152 @@
...
+        - --providers.kubernetesingress=true
+        # you need to manually set this IP to the incoming public IP
+        # that your ingress resources would use. Note it only affects
+        # status and kubectl UI, and doesn't really do anything
+        # It could even be left out https://github.com/containous/traefik/issues/6303
+        - --providers.kubernetesingress.ingressendpoint.ip=127.0.0.1
+        ## uncomment these and the ports above and below to enable
+        ## the web UI on the host NIC port 8080 in **insecure** mode
+        # - --api.dashboard=true
+        # - --api.insecure=true
```
ngbrown, line 58: Uncommenting these lines to enable the dashboard causes the pod to crash with a port conflict:

```
traefik.go:76: command traefik error: error while building entryPoint traefik: error preparing server: error opening listener: listen tcp :8080: bind: address already in use
```

Also, shouldn't the CustomResourceDefinitions have been included in this service, or is there a reason to not use them?

balasu: Hi
enabling this parameters for dashboard is insecure, hence by default its disabled, yes CRD to this service is the right way to enable dashboard which available recently . traefik version used here is 2.23, if user wants dashboard to know about services/pods/containers they can use portainer.

PR #1688, opened 2020-10-27
Allow for setting the failure domain

ktsakalozos:


joedborg (approved):
LGTM thanks

ktsakalozos:
@tobiasmcnulty this failure domain option has nothing to do with the k8s failure domain labels. This option lets the datastore (dqlite) know what the failure domains are so it can distribute the database copies accordingly.

Issue #1644, opened 2020-10-13
v1.18.9 flanneld segfault on groovy


panlinux:
```
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2972]: PrevNode.Value: {"Network": "10.1.0.0/16", "Backend": {"Type": "vxlan"}}
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2978]: {"Network": "10.1.0.0/16", "Backend": {"Type": "vxlan"}}
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2912]: fatal error: unexpected signal during runtime execution
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2912]: [signal SIGSEGV: segmentation violation code=0x1 addr=0xe5 pc=0x7f27a28a10c4]
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2912]: runtime stack:
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2912]: runtime.throw(0x19a3a14, 0x2a)
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2912]: #011/usr/local/go/src/runtime/panic.go:616 +0x81
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2912]: runtime.sigpanic()
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2912]: #011/usr/local/go/src/runtime/signal_unix.go:372 +0x28e
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2912]: goroutine 1 [syscall, locked to thread]:
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2912]: runtime.cgocall(0x14432e0, 0xc420079c50, 0x19a184f)
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2912]: #011/usr/local/go/src/runtime/cgocall.go:128 +0x64 fp=0xc420079c08 sp=0xc420079bd0 pc=0x402304
Oct 13 20:03:03 ubuntu microk8s.daemon-flanneld[2912]: os/user._Cfunc_mygetpwuid_r(0x0, 0xc4200a6d20, 0x3002460, 0x400, 0xc4200b0030, 0x0)
[...]

Issue #892, opened 2020-01-04
microk8s.start fails without sudo


dannyyy:
I added my user to the microk8s group to avoid writing sudo all the time.
```
sudo usermod -a -G microk8s <username>
```

If i have to restart the microk8s cluster and do it without sudo, then the stopping part is working fine. But the starting part will not bring up flanned.

```
$ microk8s.start
Started.
Enabling pod scheduling
error: USAGE: uncordon NODE [flags]
See 'kubectl uncordon -h' for help and examples
```

```
$ microk8s.status
microk8s is not running. Use microk8s.inspect for a deeper inspection
```

```
$ microk8s.inspect
Inspecting services
  Service snap.microk8s.daemon-cluster-agent is running
 FAIL:  Service snap.microk8s.daemon-flanneld is not running
For more details look at: sudo journalctl -u snap.microk8s.daemon-flanneld
  Service snap.microk8s.daemon-containerd is running
  Service snap.microk8s.daemon-apiserver is running
  Service snap.microk8s.daemon-apiserver-kicker is running
  Service snap.microk8s.daemon-proxy is running
  Service snap.microk8s.daemon-kubelet is running
  Service snap.microk8s.daemon-scheduler is running
  Service snap.microk8s.daemon-controller-manager is running
  Service snap.microk8s.daemon-etcd is running
  Copy service arguments to the final report tarball
Inspecting AppArmor configuration
Gathering system information
  Copy processes list to the final report tarball
  Copy snap list to the final report tarball
  Copy VM name (or none) to the final report tarball
  Copy disk usage information to the final report tarball
  Copy memory usage information to the final report tarball
  Copy server uptime to the final report tarball
  Copy current linux distribution to the final report tarball
  Copy openSSL information to the final report tarball
  Copy network configuration to the final report tarball
Inspecting kubernetes cluster
  Inspect kubernetes cluster

Building the report tarball
  Report tarball is at /var/snap/microk8s/1107/inspection-report-20200104_113942.tar.gz
```

```
$ sudo journalctl -u snap.microk8s.daemon-flanneld
-- Logs begin at Fri 2020-01-03 13:36:06 UTC, end at Sat 2020-01-04 11:39:52 UTC. --
Jan 03 13:38:36 df-k8s-mikro systemd[1]: Started Service for snap application microk8s.daemon-flanneld.
[...]

PR #1713, opened 2020-11-05
Get flannel to use libc from core

ktsakalozos:
Fixes: https://github.com/ubuntu/microk8s/issues/1644 https://github.com/ubuntu/microk8s/issues/892 for 1.19


joedborg (approved):
LGTM :+1:

Issue #1741, opened 2020-11-16
"microk8s help" command not working on MacOS


Credentive-Sec:
Attempting to run "microk8s help" from the command line on the MacOS version. I receive the following error:

```
> microk8s help              
sudo: microk8s.help: command not found
An error occurred when trying to execute 'sudo microk8s.help' with 'multipass': returned exit code 1.
```

After executing "multipass shell microk8s-vm", I have identified the issue. There is no "microk8s.help" command in the vm. Executing "microk8s help" at the command line of the vm returns the expected result.

Either the microk8s.help command should be added or the command "microk8s help" should be passed as is to the vm.

joedborg:
Hi @Credentive-Sec, thanks for raising the issue. 

I believe microk8s hasn't installed inside the VM correctly.  Could you shell back into the VM and run the following command:

```
sudo snap install microk8s --classic
```

And let me know what happens?

Thanks,
Joe

Credentive-Sec:
ubuntu@microk8s-vm:~$ sudo snap install microk8s --classic
snap "microk8s" is already installed, see 'snap help refresh'

Credentive-Sec:
I want to be careful to follow your instructions to the letter rather than making assumptions - but should I uninstall and reinstall the snap?

joedborg:
@Credentive-Sec ah, I've just realised it's because we don't actually have a dedicated help section, which is why this isn't working.  Instead, you should run
```
microk8s --help
```
I agree this should be handled better though, so I'll have a think about how that could be.

Credentive-Sec:
microk8s --help works for now. Thanks for your help!

PR #1742, opened 2020-11-16
Add help subcommand for wrappers

joedborg:
This helps with the MacOS and Windows installers:https://github.com/ubuntu/microk8s/issues/1741

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -18,6 +18,8 @@ base: core
 apps:
   microk8s:
     command: microk8s.wrapper
+  help:
+    command: microk8s.wrapper --help
   daemon-etcd:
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -18,6 +18,8 @@ base: core
 apps:
   microk8s:
     command: microk8s.wrapper
+  help:
```
ktsakalozos, line 21: This does not look right. If we add the help command, there should be a wrapper for `microk8s help` and the `microk8s.wrapper` should be calling this wrapper.

ktsakalozos: Also if we are to add the help command we should work e bit on the presentation. What do you thing?

Issue #1752, opened 2020-11-19
Echo text refers to dqlite in 002-switch-to-flannel-etcd upgrade-scripts


mercimat:
At the beginning of the following scripts, the "echo" command mentions "dqlite upgrade" while the scripts are about switching to flannel-etcd. See:
https://github.com/ubuntu/microk8s/blob/master/upgrade-scripts/002-switch-to-flannel-etcd/commit-master.sh#L4
https://github.com/ubuntu/microk8s/blob/master/upgrade-scripts/002-switch-to-flannel-etcd/rollback-master.sh#L4

This seems to be a copy/paste from similar scripts in upgrade-scripts/001-switch-to-dqlite.

I also noticed a small typo here that could be fixed at the same time:
https://github.com/ubuntu/microk8s/blob/master/upgrade-scripts/002-switch-to-flannel-etcd/rollback-master.sh#L44

I created a PR with a suggestion of fix. Please let me know in case you think there are further changes to be made.

ktsakalozos:
Thank you for the fixes @mercimat.

PR #1753, opened 2020-11-19
update echo text in flannel-etcd upgrade scripts

mercimat:
At the beginning of these scripts, the "echo" command mentions "dqlite upgrade" while they are about switching to flannel-etcd.
This seems to be a copy/paste from similar scripts in upgrade-scripts/001-switch-to-dqlite.
I also noticed a small typo at the end of `rollback-master.sh` that I fixed at the same time.

See https://github.com/ubuntu/microk8s/issues/1752

I used `flannel-etcd` instead of dqlite to update the text, in an attempt to keep it consistent with the other files. But please let me know in case there are better descriptions that would apply here.

### DIFF
--- a/upgrade-scripts/002-switch-to-flannel-etcd/commit-master.sh
+++ b/upgrade-scripts/002-switch-to-flannel-etcd/commit-master.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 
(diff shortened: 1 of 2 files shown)


ktsakalozos (approved):
Thank you for spotting those errors @mercimat

PR #1776, opened 2020-11-26
Fix pytest dependency for xenial testing

chris-sanders:
The reason pytest isn't running on xenial is because importlib-metadata doesn't support python 3.5 on pypy any longer. I've pinned the version here which fixes the issue. It's worth considering using apt packages rather than pypy when available for this very reason. I didn't make that change here, just fixed the existing setup.


ktsakalozos (approved):
LGTM +1 thank you @chris-sanders

PR #1779, opened 2020-11-28
[Fix] Enable linkerd cli on arm architecture

balchua:
This is a bug fix to allow linkerd cli on arm architectures.
Linkerd on arm is looking good.  I have also enabled the test on arm.


ktsakalozos (approved):
LGTM +1 thank you @balchua

PR #1782, opened 2020-12-01
Switch microk8s enable/disable wrappers to Python script 

knkski:
Followup to #1467, handles some scenarios better and adds tests for them in CI


tests/test_addon_args.py
```diff
@@ -0,0 +1,28 @@
...
+import pytest
+import yaml
+
+
+def test_invalid_addon():
+    with pytest.raises(sh.ErrorReturnCode_1):
+        sh.microk8s.enable.foo()
+
+
+def test_help_text():
```
ktsakalozos, line 13: Please add this in the test_addons suite so we run the test eitherway.

scripts/wrappers/disable.py
```diff
@@ -0,0 +1,43 @@
...
+    """Disables one or more MicroK8s addons.
+
+    For a list of available addons, run `microk8s status`.
+
+    To see help for individual addons, run:
+
+        microk8s disable ADDON -- --help
+    """
+
+    if print_help(addons):
```
ktsakalozos, line 28: Every time we disable an addon we try to print its help?

scripts/wrappers/enable.py
```diff
@@ -1,20 +1,43 @@
...
+        microk8s enable ADDON -- --help
+    """
+
+    if print_help(addons):
+        return
+
+    is_cluster_locked()
+    exit_if_no_permission()
+    ensure_started()
+    wait_for_ready()
```
ktsakalozos, line 34: The behavior we had until now is to wait for 30 seconds not wait forever. This is because it is better to say, "tried to enable the addon and failed" than to hang.

scripts/wrappers/disable.py
```diff
@@ -0,0 +1,43 @@
...
+        return
+
+    is_cluster_locked()
+    exit_if_no_permission()
+    ensure_started()
+    wait_for_ready()
```
ktsakalozos, line 34: Please wait for 30 seconds.

scripts/wrappers/common/utils.py
```diff
@@ -211,3 +224,66 @@ def set_service_expected_to_start(service, start=True):
...
+
+    # Backwards compatibility with enabling multiple addons at once, e.g.
+    # `microk8s.enable foo bar:"baz"`
+    if all(a.split(':')[0] in existing_addons for a in addons) and len(addons) > 1:
+        for addon in addons:
+            if addon in xabled_addons and addon != 'kubeflow':
+                click.echo("Addon %s is already %sd." % (addon, action))
+            else:
+                addon, *args = addon.split(':')
+                wait_for_ready()
```
ktsakalozos, line 254: Please keep the old behavior and wait for 30 seconds.

scripts/wrappers/disable.py
knkski: Updated the function name to clarify its purpose, and added a docstring. The function will check to see if a `--help` or `-h` flag was passed in for an addon. If the addon doesn't handle the flag itself, it will print a generic help message and return `True`, meaning that the addon had its help text printed, and should not actually be enabled/disabled. This means that we can migrate the addons as necessary over to handling a `--help` flag.

tests/test_addon_args.py
knkski: I moved it over to `test-addons.py`, but am marking it as `addon_args` so that we can run it separately. I think we'd want to run these tests against a blank-slate MicroK8s install, otherwise addons may be in unusual states that will break these tests.

ktsakalozos:
This PR is failing because libco was remove from the canonical repo. This problem is addressed in another PR. Merging.

PR #1797, opened 2020-12-08
Limit scope of containerd part

knkski:
Otherwise, it pulls in a version of files such as `enable.kubeflow.sh` that conflicts with what is sitting in the git repo.

For example:

```
Failed to stage: Parts 'containerd' and 'microk8s' have the following files, but with different contents:
    microk8s-resources/actions/enable.kubeflow.sh

Snapcraft offers some capabilities to solve this by use of the following keywords:
    - `filesets`
    - `stage`
    - `snap`
    - `organize`

To learn more about these part keywords, run `snapcraft help plugins`.
```

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -317,15 +317,14 @@ parts:
   containerd:
     build-snaps: [go]
     after: [iptables]
-    source: .
-    source-type: git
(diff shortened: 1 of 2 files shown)


ktsakalozos (approved):
LGTM +1

PR #1800, opened 2020-12-09
Fix Kubeflow hostname handling

knkski:
Ambassador is no longer used in the bundle


ktsakalozos (approved):
LGTM +1 thank you @knkski

PR #1820, opened 2020-12-14
Fix brew cask deprecation warning

kylemclaren:
During initial install on macOS, if 'multipass' is not installed:

```
➜  ~ microk8s install
Support for 'multipass' needs to be set up. Would you like to do that it now? [y/N]: y
Updating Homebrew...
Warning: Calling brew cask install is deprecated! Use brew install [--cask] instead.
```


joedborg (approved):

joedborg:
Thanks Kyle!

PR #1824, opened 2020-12-15
Reverting Kubeflow bundle to older version

knkski:
Until Juju bug #1907860 is fixed, which prevents usage of Istio


ktsakalozos (approved):
LGTM +1 Sorry for the late review, just came back.

PR #1825, opened 2020-12-15
Add pytest based nodetest

chris-sanders:
This brings install and upgrade testing up to parity with the bash based
version. Reorganize the testnode code into a module instead of
one large file. Finally fixed linting as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ tests/__pycache__
 /installer/**/__pycache__
 /installer/dist/
 /installer/build/
+.tox_env/
+__pycache__/
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,0 +1,6 @@
+kubernetes
+jinja2
+pylxd
+pytest
+pytest-xdist
+pyyaml
--- a/tests/test_distros.py
+++ b/tests/test_distros.py
@@ -1,8 +1,8 @@
 #!/bin/env python3
 
+import os
 from distutils.util import strtobool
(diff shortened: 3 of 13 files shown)


joedborg:
A few points for 2021 us:

* These don't run on EC2 probably due to a weird bug with LXD.
* Timeouts seem a bit low, e.g. my laptop can't bring up DNS in time.
* Docker is not running inside the LXD container so tests fail on docker pull.

chris-sanders:
I've added a global test coefficient. I have no idea what was going on with that EC2 instance, but I suspect the current bash scripts fail in the same way since spin up LXD containers with this same profile and install the same snaps.

I don't follow the docker issue though, I don't experience that. I think you might be hitting a snap in lxd on zfs issue. If you can reproduce but I don't hit that particular issue.

I do wonder if some of the timeouts should be increased for individual things that are being cut a little to close. I was setting these based on passing on my laptop but it was just a starting point. We should play with global vs individual tests to land them wherever seems the most appropriate.

chris-sanders:
With the last push all tests are passing again for me, try the environment variable to adjust timeouts and see if this runs for you now @joedborg

joedborg:
@chris-sanders https://github.com/ubuntu/microk8s/commit/466df0a2f7d9ab345f8818009632d395197842f0 needs merging into this PR

chris-sanders:
@joedborg added

joedborg (approved):
LGTM!

Issue #1831, opened 2020-12-18
Unable to install 1.20/stable


balchua:
I am unable to install channel `1.20/stable`.  Fresh install of MicroK8s.

I keep getting this error.

```
thor@norse:~$ sudo snap install microk8s --channel 1.20 --classic
error: cannot perform the following tasks:
- Run configure hook of "microk8s" snap if present (run hook "configure": 
-----
++ local 'IP_ADDRESSES=192.168.2.15 10.211.92.1 172.17.0.1 '
++ cp /var/snap/microk8s/1864/certs/csr.conf.template /var/snap/microk8s/1864/certs/csr.conf.rendered
++ '[' '192.168.2.15 10.211.92.1 172.17.0.1 ' == 127.0.0.1 ']'
++ '[' '192.168.2.15 10.211.92.1 172.17.0.1 ' == none ']'
++ local ips= sep=
++ local -i i=3
+++ echo '192.168.2.15 10.211.92.1 172.17.0.1 '
++ for IP_ADDR in $(echo "$IP_ADDRESSES")
++ ips+='IP.3 = 192.168.2.15'
++ sep='\n'
++ for IP_ADDR in $(echo "$IP_ADDRESSES")
++ ips+='\nIP.4 = 10.211.92.1'
++ sep='\n'
++ for IP_ADDR in $(echo "$IP_ADDRESSES")
++ ips+='\nIP.5 = 172.17.0.1'
++ sep='\n'
++ /snap/microk8s/1864/bin/sed -i 's/#MOREIPS/IP.3 = 192.168.2.15\nIP.4 = 10.211.92.1\nIP.5 = 172.17.0.1/g' /var/snap/microk8s/1864/certs/csr.conf.rendered
++ '[' -f /var/snap/microk8s/1864/certs/csr.conf ']'
++ local force
++ /snap/microk8s/1864/usr/bin/cmp -s /var/snap/microk8s/1864/certs/csr.conf.rendered /var/snap/microk8s/1864/certs/csr.conf
++ force=false
++ false
++ '[' '!' -f /var/snap/microk8s/1864/certs/front-proxy-client.crt ']'
+++ /snap/microk8s/1864/usr/bin/openssl x509 -noout -issuer
++ '[' 'issuer= /CN=front-proxy-ca' == 'issuer=CN = 127.0.0.1' ']'
++ echo 0
+ '[' 0 == 1 ']'
[...]

PR #1834, opened 2020-12-18
Use apiserver readyz endpoint 

balchua:
In the  `configure` hook of snap, use the `readyz` endpoint to if apiserver is ready.
Fixes: #1831


ktsakalozos (approved):
LGTM +1 thank you @balchua. I was not able to reproduce the error you are seeing but I do not see anything wrong with having an extra check on the availability of the apiserver.

PR #1860, opened 2021-01-01
Fix #719 - support for armhf and remote builds

### LINKED ISSUE #719: Microk8s on armhf architecture

angelnu:


angelnu:
tested with KUBE_VERSION=v.1.20.0 and it built sucesfully

ktsakalozos:
Thank you for this PR @angelnu 

Note that it will take us a bit more time to go through it given there is an extra overhead in supporting a new architecture (hardware, CI, testing, things you do not see etc).

ktsakalozos:
It might worth splitting the remote-builds patch in a separate PR.

angelnu:
Yes, I can do the split.

I also noticed as I am doing my test that the commands to enable the addons are already not working correctly - most of them affect all nodes in the cluster so the relevant aspect before enabling is if the cluster contains only compatible architectures. Example: multus need to be enabled on x86 but it then works on arm/arm64.

angelnu:
I will re-submit with just the armrf part now that the remote build got merged

PR #1872, opened 2021-01-06
Re-add ARM64 support for multus (reverting #1413)

apnar:
This re-adds ARM64 support for multus.  Support was originally removed in #1413 due to some testing issues.  Those testing issues have now been resolved in #1571 and #1630.  It also supersedes open PR #1417 which I just closed.

### DIFF
--- a/microk8s-resources/actions/multus.yaml
+++ b/microk8s-resources/actions/multus.yaml
@@ -136,3 +136,62 @@ spec:
       - name: cnibin
         hostPath:
           path: "{{SNAP_DATA}}/opt/cni/bin/"
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: kube-multus-ds-arm64
+  namespace: kube-system
+  labels:
+    tier: node
+    app: multus
+    name: multus
+spec:
+  selector:
+    matchLabels:
+      name: multus
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        tier: node
+        app: multus
+        name: multus
+    spec:
+      hostNetwork: true
+      nodeSelector:
+        kubernetes.io/arch: arm64
+      tolerations:
+      - operator: Exists
+        effect: NoSchedule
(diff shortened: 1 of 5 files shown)


ktsakalozos (changes requested):
Thank you for the PR @apnar 

There are a few things that need a bit more care.
First, in `snapcraft.yaml` we should not be removing the the enable/disable.multus.sh scripts or else the enable command failes with `Nothing to do for multus`.
Second what I run the tests the multus-alpine was not created and it was failing with: http://paste.ubuntu.com/p/Df2f8DNdpS/

apnar:
Thanks for the quick review @ktsakalozos!  Apologies for missing the snapcraft lines, not sure how I missed including that in the PR (and don't actually have an ARM64 box so didn't catch it).  I've fixed that and repushed the PR.

As to the test, it seems to be failing because it can't find an interface on the box called "eth0", note line 362 of the output you pasted:

` error adding container to network "net1": failed to lookup master "eth0": Link not found`

This is one of reasons I originally chose to use "ptp" type interfaces instead of "macvlan" for the test since I don't have visibility into your test boxes or what their interface names are.  It looks like you switched from ptp to macvlan in #1630 which I presume you had a good reason to do.  Possible solutions to fixing the test that I can think of:

- switch back to ptp type interfaces for the test
- change the name or alias the interface on the arm64 test box to be eth0
- try to determine the name of the primary interface at test time and modify the multus-networks.yaml accordingly
- run different tests for amd64 and arm64 and set arm64 test to use the name of the interface of the arm64 test box 

The first and the third option are likely the cleanest.  The first would be easiest but without knowing the issues you ran into with ptp I wouldn't want to do that.

georgenicoll:
Is there anything happening with this one?  Be interested to have multus on rPi with microk8s.

neoaggelos:
This is completed with https://github.com/canonical/microk8s-community-addons/pull/56 and https://github.com/canonical/microk8s-community-addons/pull/60, and will be available starting from MicroK8s 1.25

PR #1876, opened 2021-01-07
pyupgrade --py36-plus **/*.py

cclauss:
Upgrade syntax for newer versions of Python.  https://github.com/asottile/pyupgrade


joedborg:
Thanks for the PR @cclauss, sadly we still support MicroK8s on 16.04, which comes with Py3.5, meaning we must remain backwards compatible.  We are looking at bundling a fixed version of Python with MicroK8s, at which point, we can use this PR.

cclauss:
`pyupgrade --py3-plus` should work for that scenario.

PR #1881, opened 2021-01-11
Create CONTRIBUTING.md

evilnick:
Add a contributing page. Will link from main readme when added


chris-sanders:
It would be nice to add or link to something that mentions running tests locally before submitting them for a PR. Ideally, that would just be via a tox command which runs the tests we consider 'default' for all PRs. Until we land #1825 there isn't a tox, so either we land that and adjust tox as necessary or document what the current expectation is and update it when we do have tox setup.

evilnick:
@ktsakalozos do you want to add whatever testing needs to be done here, or explain it to me so I can finish this up?

ktsakalozos:
@evilnick @chris-sanders just added a section on how to run the tests currently with pytest, we can revise this as soon as the tox based testing is merged.

Also added a PR template.

evilnick:
@ktsakalozos Thanks! :+1:

PR #1904, opened 2021-01-15
Bundle kubectl with Windows installer and set kubectl as brew requirement for MacOS

joedborg:
To address the complaint that file manipulation (e.g. `microk8s kubectl apply -f ./pod.yaml`) is difficult with the installers.

This difficultly is caused by files being referenced on the host, which are not then made available inside the VM running MicroK8s.  The most pragmatic fix for this is appears to be running kubectl on the host pointing at the MicroK8s kubeconfig file, which is written to the host every time `microk8s install` is run.  This does not overwrite or merge with `.kube/config` in order to give the "black box" feel of MicroK8s on Linux.  When `microk8s kubectl` is run, the command is no longer passed into the VM, rather run on the host with `--kubeconfig` to point to the aforementioned written file.

For Windows, we bundle `kubectl` with the NSIS installer and place it into the MicroK8s install directory, within a subdirectory to allow user choice of being added to PATH during install.

For MacOS, `kubernetes-cli` is added as a requirement to the brew file.  I think it's fair to assume that this would usually already be installed in most cases but, if not, brew will install it along side MicroK8s.

### DIFF
--- a/.github/workflows/build-installer.yml
+++ b/.github/workflows/build-installer.yml
@@ -43,6 +43,12 @@ jobs:
         file-url: https://github.com/canonical/multipass/releases/download/v1.5.0/multipass-1.5.0+win-win64.exe
         file-name: multipass.exe
         location: ${{ github.workspace }}/installer/windows
+    - name: Download kubectl
+      uses: carlosperate/download-file-action@v1.0.3
+      with:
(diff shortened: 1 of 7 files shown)


ktsakalozos (approved):
Thank you @joedborg LGTM +1

PR #1918, opened 2021-01-20
Kubelite, consolidate k8s services

ktsakalozos:
With this PR we start kube-proxy, scheduler, controller-manager, apiserver, kubelet from the same server. As a result the memory footprint and the snap size are reduced.

The old services are stopped via a dummy service, see `run-null-daemon`.

The patch to kubernetes is at https://github.com/canonical/kubernetes-dqlite/commit/7136e0646e50d26d40db49ae0f6606775d11a56a

=========================================================

*Also verify you have:*
* [x] Read the [contributions](https://github.com/ubuntu/microk8s/CONTRIBUTING.md) page.


balchua:
I was thinking to have an extra argument to `kubelite` indicating which components to start.
As i recall, in an HA setup on certain nodes we dont run controller manager and scheduler.
Wdyt?

balchua (commented):
If there's a way to further reduce the cpu/memory by not starting certain components when not needed like controller manager and scheduler, that will be great for resource constrained environment like raspberry pi.

ktsakalozos:
> If there's a way to further reduce the cpu/memory by not starting certain components when not needed like controller manager and scheduler, that will be great for resource constrained environment like raspberry pi.

This was what the control-plane-kicker daemon was doing, so I do see the benefit of not starting extra daemons. However, now that we have a single daemon we will have to take into account a few more facts. First, the memory gain of not starting the scheduler and the controller-manager are not that great since we will be loading the go runtime either way. Second, we would want to monitor the dqlite nodes and collocate the control plane with the dqlite-voters (at least this is what we used to do in the control-plane-kicker daemon). As we do not want tot restart all daemons we would need to have logic in the kubernetes patch to make it stop and start specific daemons (now go routines). I would like not to do this at this point. Eventually, we would like to keep the patch minimal and maybe try to merge it upstream.

docs/dqlite.md
```diff
@@ -25,10 +25,10 @@ specific tag.
 ```
 git reset c3c94660a58998dde628de7c716a63b695327016^  --soft 
 git stash                                                    # put the patch on the stash
-git reset 52c56ce7a8272c798dbc29846288d7cd9fbae032 --hard    # Go tho the 1.18.2 release commit
+git reset 52c56ce7a8272c798dbc29846288d7cd9fbae032 --hard    # Go to the 1.18.2 release commit
 git stash apply
 git commit -m "Apply the dqlite patch"
-```
+    ```
```
joedborg, line 31: Is this a mistake?

scripts/cluster/agent.py
```diff
@@ -412,25 +413,23 @@ def configure():
     '''
 
     if "service" in configuration:
-        for service in configuration["service"]:
-            print("{}".format(service["name"]))
-            if "arguments_update" in service:
+        for srv in configuration["service"]:
```
joedborg, line 416: Why the refactor of this?

scripts/cluster/agent.py
```diff
@@ -502,15 +501,15 @@ def update_dqlite_ip(host):
...
+    service('stop', 'apiserver')
     time.sleep(10)
 
     cluster_dir = "{}/var/kubernetes/backend".format(snapdata_path)
     # TODO make the port configurable
     update_data = {'Address': "{}:{}".format(host, dqlite_port)}
     with open("{}/update.yaml".format(cluster_dir), 'w') as f:
         yaml.dump(update_data, f)
-    subprocess.check_call("snapctl start microk8s.daemon-apiserver".split())
+    service('start', 'apiserver')
```
joedborg, line 512: I think this will fail the black formatter?

joedborg (changes requested):

scripts/cluster/agent.py
ktsakalozos: We called a module `service` so this variable was clashing with it.

scripts/cluster/agent.py
ktsakalozos: I think it is passing on these GH actions, right?

docs/dqlite.md
ktsakalozos: Fixed, thanks

Issue #1917, opened 2021-01-20
Internal error occurred: failed to allocate a serviceIP: range is full


rdeepc:
Can't deploy more than **252**(around) Services because **service-cluster-ip-range 10.152.183.0/24** by default, which is **254** hosts.
So it runs **out of ip address** for the services.

To fix the problem I had to use **service-cluster-ip-range=10.152.0.0/16** in **microk8s-resources/default-args/kube-apiserver** to increase the number of ip address. Also set IP according to the range in some other files. 

It would be helpful if it is **service-cluster-ip-range=10.152.0.0/16** by default.

**microk8s inspect**
[inspection-report-20210120_043107.tar.gz](https://github.com/ubuntu/microk8s/files/5840179/inspection-report-20210120_043107.tar.gz)

balchua:
@rdeepc thanks for this one.  Are you interested on putting up a PR for this?  Its just right that you get credited for this work.
Thanks!

mac2000:
`/var/snap/microk8s/current/args/kube-apiserver` is the path for configuration file that needs to be modified

PR #1921, opened 2021-01-21
Increase serviceIP Range (#1917)

rdeepc:
Updated service cluster IP range from 10.152.183.0/24 to 10.152.0.0/16, which increases the number of service IPs

### Thank you for making MicroK8s better

*Also verify you have:*
* [x] Read the [contributions](https://github.com/ubuntu/microk8s/CONTRIBUTING.md) page.

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -370,7 +370,7 @@ produce_certs() {
 
     # Generate apiserver CA
     if ! [ -f ${SNAP_DATA}/certs/ca.crt ]; then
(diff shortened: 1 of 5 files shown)


ktsakalozos:
Thank you for this PR @rdeepc 

Have you tested to see what would happen if you have a cluster with 10.152.183.0/24 and you join it with a node of 10.152.0.0/16?

I wonder if we need to update the joining process [1] to propagate this information so that already existing clusters are not affected.

[1] https://github.com/ubuntu/microk8s/blob/master/scripts/cluster/join.py#L906

rdeepc:
Hi @ktsakalozos 

I have tested as you mentioned, it joins without any problem, but if the master is in 10.152.183.0/24 , the IP range problem does not solve

PR #1925, opened 2021-01-22
Better connectivity check

knkski:
Uses image with curl baked in, and prints out more troubleshooting information.


ktsakalozos (approved):
LGTM +1 thank you @knkski

PR #1963, opened 2021-02-02
Enforce YAML formatting

joedborg:
Draft PR to start a discussion around this.  I believe our YAML should be enforced the same way that our Python is.


joedborg:
Using [Prettier](https://prettier.io) as it's very opinionated and integrates with most editors.

knkski (approved):
LGTM, other than squashing commits before merging

balchua:
LGTM. 👍

ktsakalozos:
Still failing @joedborg . Maybe we should not enforce yaml style in the manifests?

joedborg:
@ktsakalozos all passing now

PR #1871, opened 2021-01-05
Use Nvidia operator

joedborg:
This branch works on nodes without any Nvidia drives installed.  We are waiting on https://github.com/NVIDIA/gpu-operator/issues/126 to allow this to work on nodes already provisioned (essential for most MicroK8s users).

The API remains the same:

```
sudo snap install microk8s --classic
sudo microk8s enable gpu
```

This should fix issues we've been seeing with users on odd combinations of OS and driver version and has also allowed us to strip a bunch of build steps, which is nice.

PR #1964, opened 2021-02-02
Correct containerd systemd settings

joedborg:
Pass correct  settings to the containerd systemd unit.  This is broken off from https://github.com/ubuntu/microk8s/pull/1871.


ktsakalozos (approved):
LGTM +1 thank you @joedborg

PR #1985, opened 2021-02-08
Fixes in kubelite startup script

ktsakalozos:
1. If we do not have `br_netfilter` loaded try to load it.
2. Fix iptables-nft rule lost in code merges
3. Remove duplicate volumes declaration in coredns manifest


joedborg:
Looks good to me!

joedborg (approved):

PR #1996, opened 2021-02-09
Use hyphen instead of underscore

joedborg:
Keeps names consistent.


ktsakalozos:
To change this we will need to update our CI scripts in https://github.com/charmed-kubernetes/jenkins/blob/master/jobs/microk8s/snapstore.py#L149 and back port this rename to all branches since they go through jenkins.

chris-sanders:
@joedborg actually I think we're going the wrong direction with this. test-distros isn't a valid python module name and pytest won't pickup tests in it. The files with pytests in it should actually use the underscore or they aren't valid for test discovery.

chris-sanders (changes requested):
I think we're going the wrong direction with this. test-distros isn't a valid python module name and pytest won't pickup tests in it. The files with pytests in it should actually use the underscore or they aren't valid for test discovery.

ktsakalozos:
> To change this we will need to update our CI scripts in https://github.com/charmed-kubernetes/jenkins/blob/master/jobs/microk8s/snapstore.py#L149 and back port this rename to all branches since they go through jenkins.

Ignore me, this is a different file.

PR #2005, opened 2021-02-11
Remove invalid hooks

joedborg:
LP builds are failing post-build review because of these.  They're not needed as we use microk8s start / stop anyway.


ktsakalozos (approved):
LGTM +1

Issue #2016, opened 2021-02-16
Join node not working any more


oleo65:
On v.1.20.2 joining a node is not working any more.

Syslog shows the following stack trace.

```
Feb 16 17:12:59 pi-k8s-1 microk8s.daemon-cluster-agent[567882]: error: error running snapctl: unknown service: "microk8s.daemon-apiservice"
Feb 16 17:12:59 pi-k8s-1 microk8s.daemon-cluster-agent[554105]: [2021-02-16 17:12:59,916] ERROR in app: Exception on /cluster/api/v1.0/join [POST]
Feb 16 17:12:59 pi-k8s-1 microk8s.daemon-cluster-agent[554105]: Traceback (most recent call last):
Feb 16 17:12:59 pi-k8s-1 microk8s.daemon-cluster-agent[554105]:   File "/snap/microk8s/2048/lib/python3.5/site-packages/flask/app.py", line 2447, in wsgi_app
Feb 16 17:12:59 pi-k8s-1 microk8s.daemon-cluster-agent[554105]:     response = self.full_dispatch_request()
Feb 16 17:12:59 pi-k8s-1 microk8s.daemon-cluster-agent[554105]:   File "/snap/microk8s/2048/lib/python3.5/site-packages/flask/app.py", line 1952, in full_dispatch_request
Feb 16 17:12:59 pi-k8s-1 microk8s.daemon-cluster-agent[554105]:     rv = self.handle_user_exception(e)
Feb 16 17:12:59 pi-k8s-1 microk8s.daemon-cluster-agent[554105]:   File "/snap/microk8s/2048/lib/python3.5/site-packages/flask/app.py", line 1821, in handle_user_exception
Feb 16 17:12:59 pi-k8s-1 microk8s.daemon-cluster-agent[554105]:     reraise(exc_type, exc_value, tb)
Feb 16 17:12:59 pi-k8s-1 microk8s.daemon-cluster-agent[554105]:   File "/snap/microk8s/2048/lib/python3.5/site-packages/flask/_compat.py", line 39, in reraise
Feb 16 17:12:59 pi-k8s-1 microk8s.daemon-cluster-agent[554105]:     raise value
Feb 16 17:12:59 pi-k8s-1 microk8s.daemon-cluster-agent[554105]:   File "/snap/microk8s/2048/lib/python3.5/site-packages/flask/app.py", line 1950, in full_dispatch_request
[...]

PR #2018, opened 2021-02-16
Fix: in non-ha cluster etcd with kubelite

ktsakalozos:
Partially addresses issue https://github.com/ubuntu/microk8s/issues/2016


joedborg (commented):
LGTM

PR #2029, opened 2021-02-18
Build libuv v1.18.0 in the snap

ktsakalozos:
Raft now requires libuv v1.18.0 to build. As this is not available in core we need to build it.


joedborg (approved):
LGTM

PR #1961, opened 2021-02-01
#fixes 719 - adds support to build on armhf

angelnu:
This is a continuation of the work for #719 .

After splitting the remote build support this patch adds just the capability to build locally on armrf or remotely with `snapcraft remote-build `

The mayor change beyond fixing the check on x86 only plugins was to allow building etcd from sources when no binary is available. This way the version matches on all platforms but the official binary is used when available.

The built result is functional and I am able to add nodes successfully. If there are additional tests needed before a platform is added please let me know.

*Also verify you have:*
* [x] Read the [contributions](https://github.com/ubuntu/microk8s/CONTRIBUTING.md) page.
* [x] Submitted the [CLA form](https://ubuntu.com/legal/contributors/agreement), if you are a first time contributor.

PR #2048, opened 2021-02-24
Build microk8s for armhf and other architectures

ktsakalozos:
This is the work of @angelnu from PR https://github.com/ubuntu/microk8s/pull/1961, Quoting from there:

```
This is a continuation of the work for https://github.com/ubuntu/microk8s/issues/719 .

After splitting the remote build support this patch adds just the capability to build locally
on armrf or remotely with snapcraft remote-build

The mayor change beyond fixing the check on x86 only plugins was to allow building etcd
from sources when no binary is available. This way the version matches on all platforms
but the official binary is used when available.

The built result is functional and I am able to add nodes successfully.
If there are additional tests needed before a platform is added please let me know.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ tests/__pycache__
 /installer/build/
 .tox_env/
 __pycache__/
+microk8s_*.txt #Remote build log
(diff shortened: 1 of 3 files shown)


ktsakalozos:
Thank you @angelnu for this work and for testing the waters of armhf. I need to be clear here, we merged this work so we can build the snap but we will not be supporting this architecture. At least not for now.

angelnu:
Thanks @ktsakalozos - as a hobbies here I do not expect having official support but having a build in the snap repository makes my life much easier!

Hopefully this helps others as well as make microk8s more attractive in the k8s-at-home community where other options that support armhf (such as k3s) are more popular.

Issue #2049, opened 2021-02-25
Failing to build ARM64 snap


tbertenshaw:
Trying to build an arm64 snap, from master.

I've tried building locally on a pi4 (with 8Gb running on Ubuntu 20.10) using the same command from [the github action](https://github.com/tbertenshaw/microk8s/blob/697d2f721bef8ad215ba7f992024e8213339f49c/.github/workflows/build-snap.yml#L28)

 `sg lxd -c 'snapcraft --use-lxd'`

but it errors out after some time with.

```
Building cluster-agent
+ snapcraftctl build
Looking in links: /root/parts/cluster-agent/python-packages
Processing /root/parts/cluster-agent/python-packages/Flask-1.1.2-py2.py3-none-any.whl
Processing /root/parts/cluster-agent/python-packages/PyYAML-5.4.1.tar.gz
  Installing build dependencies ... error

  ERROR: Command errored out with exit status 1:

   command: /root/parts/cluster-agent/install/usr/bin/python3 /root/parts/cluster-agent/install/lib/python3.6/site-packa

ges/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-r9wqoe58/overlay --no-warn-script-location --no-binary :none: --only-binary :none: --no-index --find-links /root/parts/cluster-agent/python-packages -- setuptools wheel Cython
       cwd: None
  Complete output (3 lines):
  Looking in links: /root/parts/cluster-agent/python-packages
  ERROR: Could not find a version that satisfies the requirement setuptools
  ERROR: No matching distribution found for setuptools
  ----------------------------------------
WARNING: Discarding file:///root/parts/cluster-agent/python-packages/PyYAML-5.4.1.tar.gz. Command errored out with exit status 1: /root/parts/cluster-agent/install/usr/bin/python3 /root/parts/cluster-agent/install/lib/python3.6/site-packages/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-r9wqoe58/overlay --no-warn-script-location --no-binary :none: --only-binary :none: --no-index --find-links /root/parts/cluster-agent/python-packages -- setuptools wheel Cython Check the logs for full command output.
ERROR: Could not find a version that satisfies the requirement PyYAML
ERROR: No matching distribution found for PyYAML
[...]

PR #2050, opened 2021-02-25
Pin versions of pip packages

ktsakalozos:
Pip packages broke the build of ARM64. Pin the pipy packages so we do not have this problem in the future.

Fixes: https://github.com/ubuntu/microk8s/issues/2049


joedborg (approved):
LGTM!

PR #2073, opened 2021-03-04
inspect script should check for kubelite

ktsakalozos:
The inspect script checks that the apiservice is running. Now we should be checking for kubelite.


joedborg (approved):
LGTM

PR #2084, opened 2021-03-09
Simplify multus test

ktsakalozos:
The old multus test was a bit flaky and it would not work on every host. Changing this to just making sure the multus daemon set pod is running.


joedborg (approved):
Looks good to me

PR #2087, opened 2021-03-10
Go 1.16. And updated k8s patches.

ktsakalozos:
K8s 1.21.0-beta.1 enforced golang 1.16 and also caused some patch conflicts.


joedborg (approved):
LGTM

PR #2095, opened 2021-03-15
Reverting daemon type back to simple for containerd

joedborg:
Reverting until we can understand why containerd does not come back after cold restart.


ktsakalozos:
Since we revert the notify work. Shouldn't we also revert the killing of containerd shims in https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/microk8s-stop.wrapper#L45 ?

ktsakalozos:
Closing in favor of https://github.com/ubuntu/microk8s/pull/2122

ktsakalozos (approved):
LGTM +1

PR #2107, opened 2021-03-17
added check for single node and force flag

### LINKED ISSUE #2009: Disabling ha-cluster addon deletes all data
Starting point: a default, single-node microk8s installation (`snap install microk8s --classic --channel=1.19`) with several deployments installed through GitLab integration, including database pods with persistent volumes (`storage` addon).

Running `microk8s disable ha-cluster` deletes all deployments, all services, all pods, and all persistent volumes _without any warning_. Everything is gone and I am left with an empty node/cluster. This was completely unexpected and you can imagine my horror. Re-enabling the addon `microk8s enable ha-cluster` does **not** bring back the data and services, the cluster remains empty :cry: 

There should be a safety question presented to the users to confirm that they are about to destroy their data on the single-node cluster irreversibly. Disabling high-availibility does not imply deleting data, this should be communicated to the users _before_ executing the action. I assume the same happens when disabling the `storage` addon, but there it is at least somewhat obvious that this could cause data-loss.

tripolkaandrey:
Fixes #2009

Added check that we are on a single node serving workloads and we do not have a force flag according to @ktsakalozos advice

### DIFF
--- a/microk8s-resources/actions/disable.ha-cluster.sh
+++ b/microk8s-resources/actions/disable.ha-cluster.sh
@@ -8,7 +8,10 @@ export IN_SNAP_LD_LIBRARY_PATH="$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/$ARCH-linux-gn
 
 source $SNAP/actions/common/utils.sh
 
-exit_if_no_permissions
+NODES_NUM=$($KUBECTL get no -o name | wc -l)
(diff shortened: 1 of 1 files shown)


tripolkaandrey:
@ktsakalozos please take a look

ktsakalozos:
Thank you @tripolkaandrey closing this PR in favor of https://github.com/ubuntu/microk8s/pull/2143

PR #2142, opened 2021-04-02
Do not ship openfaas on arm64

ktsakalozos:
The manifests produced by the openfaas addon are amd64 specific. They have:
```
      nodeSelector:
        beta.kubernetes.io/arch: amd64
```

### DIFF
--- a/microk8s-resources/wrappers/addon-lists.yaml
+++ b/microk8s-resources/wrappers/addon-lists.yaml
@@ -204,7 +204,6 @@ microk8s-addons:
       version: "6.2.2"
       check_status: "pod/gateway"
       supported_architectures:
-        - arm64
         - amd64
 
     - name: "openebs"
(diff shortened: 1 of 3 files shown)


ktsakalozos:
@LucasRoesler maybe we can add ARM64 support at a later stage.

joedborg (approved):
LGTM

kwmonroe (approved):

LucasRoesler:
> @LucasRoesler maybe we can add ARM64 support at a later stage.

You are right. I made a mistake because we had moved all of the images to multi-arch builds and there was discussion that the helm chart should work. OpenFaaS can work on ARM, but there are just a few tweaks needed to the install flow. I will followup with the needed tweaks.

LucasRoesler:
@ktsakalozos when is the next release for microk8s? We just removed the node selector this weekend and I am waiting for a new helm chart release. Once that is released, it will work on ARM out of the box and we could revert this PR.

ktsakalozos:
> when is the next release for microk8s?

Next release v1.21 is in a couple of days (8th of April). The v1.22 should be in about 4 months.

alexellis:
We fixed this yesterday, so for the next cycle of MicroK8s, this can be enabled again on ARM64.

PR #2143, opened 2021-04-02
Require force for disable HA on single node cluster

### LINKED ISSUE #2009: Disabling ha-cluster addon deletes all data

ktsakalozos:
When we have a single node and we call disable ha-cluster we require to have the --force flag.

Fixes #2009

### DIFF
--- a/microk8s-resources/actions/disable.ha-cluster.sh
+++ b/microk8s-resources/actions/disable.ha-cluster.sh
@@ -10,6 +10,33 @@ source $SNAP/actions/common/utils.sh
 
 exit_if_no_permissions
 
+FORCE=false
+PARSED=$(getopt --options=f --longoptions=force --name "$@" -- "$@")
+eval set -- "$PARSED"
+while true; do
+    case "$1" in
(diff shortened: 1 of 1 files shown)


joedborg (approved):
LGTM

PR #2148, opened 2021-04-05
Fix yaml output of microk8s.status if not running

### LINKED ISSUE #2147: Indentation error in yaml output of microk8s.status
The output of `microk8s.status` is

```
microk8s:
  running: False
high-availability:
  enabled: False
   message: microk8s is not running. Use microk8s inspect for a deeper inspection.
```

which confuses some parsers (e.g. the built-in Python 3.8) due to the extraneous space before `message`.

krichter722:
Fix #2147

### Thank you for making MicroK8s better

*Also verify you have:*
* [x] Read the [contributions](https://github.com/ubuntu/microk8s/blob/master/CONTRIBUTING.md) page.

### DIFF
--- a/scripts/wrappers/status.py
+++ b/scripts/wrappers/status.py
@@ -98,7 +98,7 @@ def print_short_yaml(isReady, enabled_addons, disabled_addons):
             print("  {}: disabled".format(disabled["name"]))
     else:
         print(
-            "{:>2} {} {}".format(
+            "{:>2}{} {}".format(
                 "",
                 "message:",
                 "microk8s is not running. Use microk8s inspect for a deeper inspection.",
@@ -135,7 +135,7 @@ def print_yaml(isReady, enabled_addons, disabled_addons):
             print("{:>4}status: disabled".format(""))
     else:
         print(
-            "{:>2} {} {}".format(
+            "{:>2}{} {}".format(
                 "",
                 "message:",
                 "microk8s is not running. Use microk8s inspect for a deeper inspection.",


scripts/wrappers/status.py
```diff
@@ -2,7 +2,7 @@
 import os
 import argparse
 
-from common.utils import (
+from .common.utils import (
```
krichter722, line 5: Seems to be necessary for testing, I'm not 100% sure what it does, but it seems to work and break nothing.

tests/test-status.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+import sh
+from utils import (
+    microk8s_reset,
+)
+
+
+class TestStatus(object):
+    @pytest.fixture(autouse=True)
+    def setup_cluster(self):
```
krichter722, line 10: I don't even know if that would hit the bug for which this PR has been opened, however you want to have an integration test for `microk8s.status` as well

scripts/wrappers/test_status.py
```diff
@@ -0,0 +1,92 @@
+from .status import (
```
krichter722, line 1: You may want to find a better location for this. I placed this here quick and dirty because Python imports are hell. Maybe you should take the time to figure out a unit test concept and including locations for unit tests. More coverage will definitely help to save time over time.

tests/test-status.py
ktsakalozos: We already have an integration test for `microk8s status` [1].

You must have already seen that we only do integration tests and not unit tests, this is probably because we do not really have complex algorithms implemented.

Please do not continue with the refactoring you are proposing in this PR. Instead keep the patch to the `microk8s stauts` minimal (should be two lines) and extend the `test_help_text` to cover the bug you are hitting. In the test you would need to kill `kubelite` ad call `microk8s.status` before systemd restarts it.

[1] https://github.com/ubuntu/microk8s/blob/master/tests/test-addons.py#L57

krichter722: The suggestion of unit tests was outside the scope of this PR. However I didn't manage to create the integration test, also because it's very frustrating and insanely inefficient to develop without fast running unit tests. You should intensively rethink your stand towards unit tests and the need for it in this project. Maybe you add the integration test yourself - meanwhile I don't know if it makes a lot of sense.

ktsakalozos: You are right, we should invest time in unit tests. Going forward we would like to rewrite some bash scripts in python. That would be a good time to pair our code with more tests.

ktsakalozos (approved):
LGTM +1 thank you @krichter722

PR #2149, opened 2021-04-05
Prepare branch for Nvidia operator with bundled driver switch

joedborg:
Will remain in draft until the update operator is open to the public (https://github.com/NVIDIA/gpu-operator/issues/126)


joedborg:
Proposed UX:

`sudo microk8s enable gpu` for auto detect (pull from the old enable gpu script).
`sudo microk8s enable gpu:force-system-driver` to force use of the system's driver.
`sudo microk8s enable gpu:force-operator-driver` to force the use of the operator's driver.

joedborg:
cc @chris-sanders, are you happy with that UX?

microk8s-resources/actions/enable.gpu.sh
```diff
@@ -9,6 +9,27 @@ readonly SOCKET="$SNAP_COMMON/run/containerd.sock"
 
 echo "Enabling NVIDIA GPU"
 
+read -ra ARGUMENTS <<< "$1"
+
+if [[ "$ARGUMENTS" == "force-system-driver" ]]
```
ktsakalozos, line 14: Do you think we should have a check to detect if we have an nvidia GPU?

joedborg: Do you think the `force` flags should ignore that?

microk8s-resources/actions/enable.gpu.sh
```diff
@@ -20,8 +41,10 @@ echo "Installing NVIDIA Operator"
 "$SNAP/microk8s-helm3.wrapper" repo add nvidia https://nvidia.github.io/gpu-operator
 "$SNAP/microk8s-helm3.wrapper" repo update
 "$SNAP/microk8s-helm3.wrapper" install gpu-operator nvidia/gpu-operator \
+  --version=v1.7.0 \
```
shivamerla, line 44: It will be good if the operator/toolkit versions are controlled through arguments too, so we don't need to change the script with each release of operator.

joedborg: Hey @shivamerla, we'll look at adding this as an argument in the future.  Obviously, we need to hard code a default for a testable / supportable release build.

Issue #2144, opened 2021-04-03
Adding existing node to the cluster causes MicroK8s to stop working


balchua:
When adding a node which is already a member, causes the node to stop working.

To reproduce:

* Start a MicroK8s node, named node1.
* Issue the `microk8s add-node`.
* Join `node1` to itself. ex. `microk8s join 178.128.125.239:25000/qwertyuiopasdfghjklzxcvbnm123456/ce24ebf71164`

Wait for sometime and the join command will succeed.  But internally the cluster isn't functioning.

Checking the `kubelite` journal logs, you will see this.

```shell
Apr 03 01:25:07 mk8s-0 microk8s.daemon-kubelite[8595]: Starting kubelite
Apr 03 01:25:07 mk8s-0 microk8s.daemon-kubelite[8595]: I0403 01:25:07.216014    8595 daemon.go:73] Waiting for the API server
Apr 03 01:25:07 mk8s-0 microk8s.daemon-kubelite[8595]: I0403 01:25:07.219868    8595 daemon.go:65] Starting API Server
Apr 03 01:25:07 mk8s-0 microk8s.daemon-kubelite[8595]: Flag --insecure-port has been deprecated, This flag has no effect now and will be removed in v1.24.
Apr 03 01:25:07 mk8s-0 microk8s.daemon-kubelite[8595]: I0403 01:25:07.221800    8595 server.go:633] external host was not specified, using 178.128.125.239
Apr 03 01:25:07 mk8s-0 microk8s.daemon-kubelite[8595]: W0403 01:25:07.222051    8595 authentication.go:519] AnonymousAuth is not allowed with the AlwaysAllow authorizer. Resetting AnonymousAuth to false. You should use a different authorizer
Apr 03 01:26:07 mk8s-0 microk8s.daemon-kubelite[8595]: Error: context deadline exceeded
Apr 03 01:26:07 mk8s-0 microk8s.daemon-kubelite[8595]: F0403 01:26:07.260092    8595 daemon.go:67] API Server exited context deadline exceeded
Apr 03 01:26:07 mk8s-0 microk8s.daemon-kubelite[8595]: goroutine 115 [running]:
Apr 03 01:26:07 mk8s-0 microk8s.daemon-kubelite[8595]: k8s.io/kubernetes/vendor/k8s.io/klog/v2.stacks(0xc00013a001, 0xc0006de380, 0x58, 0x1b9)
[...]

PR #2151, opened 2021-04-06
Improved handling of joining cases

ktsakalozos:
Fixes included in this PR:
- Enable node fingerprint verification when provided in the join connection string
- Fail with proper error message when:
    1. We join a node to itself
    2. We join a cluster and dqlite already knows about that node
- Fixes in parsing the error messages
- Allow for forcefully remove a node from dqlite using its IP.

Fixes some of the cases reported in https://github.com/ubuntu/microk8s/issues/2144

### DIFF
--- a/scripts/cluster/agent.py
+++ b/scripts/cluster/agent.py
@@ -450,7 +450,7 @@ def configure():
     return resp
 
 
-def get_dqlite_voters():
+def get_dqlite_nodes():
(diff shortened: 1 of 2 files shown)


balchua (commented):
Lgtm thanks!

joedborg (approved):
LGTM

PR #2155, opened 2021-04-08
Fix yaml indentation

ktsakalozos:
Fixes: https://github.com/ubuntu/microk8s/issues/2147


joedborg (approved):
LGTM

krichter722:
Those seem to be the two spots. I created unit tests in https://github.com/ubuntu/microk8s/pull/2148. If the lgty you can incorporate them here merge them there. Since you already started this quick fix, I assume you're going to do the backporting as well?

Issue #2071, opened 2021-03-03
edge version gives error: Failed to start microk8s services


robertbovens:
Please run `microk8s inspect` and attach the generated tarball to this issue.

We appreciate your feedback. Thank you for using microk8s.
[inspection-report-20210303_194517.tar.gz](https://github.com/ubuntu/microk8s/files/6078742/inspection-report-20210303_194517.tar.gz)

Installed a fresh latest/edge version, enablel helm3, rbac, dns, metrics-server and storage. After enabling storage microk8s can not start and says it has problem enabling containerd.

I installed it because i want a newer version of helm3 and this version gives helm3 version 3.5, but it seems it still has problems

After i run the inspect command, dns, rbac and metrics-server wasnt installed but i was able to start microk8s. Now i enabled it again and now it runs without errors

robertbovens:
After reboot i get the following message:

root@test:/home/robert# microk8s.status
microk8s is not running. Use microk8s inspect for a deeper inspection.
root@test:/home/robert# microk8s.start
error: cannot perform the following tasks:
- Run service command "start" for services ["daemon-apiserver" "daemon-apiserver-kicker" "daemon-cluster-agent" "daemon-containerd" "daemon-control-plane-kicker" "daemon-controller-manager" "daemon-etcd" "daemon-flanneld" "daemon-kubelet" "daemon-kubelite" "daemon-proxy" "daemon-scheduler"] of snap "microk8s" ([start snap.microk8s.daemon-containerd.service] failed with exit status 1: Job for snap.microk8s.daemon-containerd.service failed because a timeout was exceeded.
See "systemctl status snap.microk8s.daemon-containerd.service" and "journalctl -xe" for details.
)
Failed to start microk8s services. Check snapd logs with "journalctl -u snapd.service"

Running microk8s.stop and then start gives the same error.

robertbovens:
root@test:/home/robert# systemctl status snap.microk8s.daemon-containerd.service
● snap.microk8s.daemon-containerd.service - Service for snap application microk8s.daemon-containerd
     Loaded: loaded (/etc/systemd/system/snap.microk8s.daemon-containerd.service; disabled; vendor preset: enabled)
     Active: failed (Result: timeout) since Wed 2021-03-03 20:08:46 UTC; 1min 23s ago
   Main PID: 7643 (code=killed, signal=KILL)

[...]

PR #2159, opened 2021-04-10
Clean leftover containers after a cold boot

ktsakalozos:
After a cold boot containerd will go over all leftover container entries and will try to revive them. This causes kubelet to not be able to start properly. In this PR we mark when containerd started so we know if this is the first time containerd started after a system boot. If it is a first start we manually remove all container entries since we know they do not run anymore.

Fixes: https://github.com/ubuntu/microk8s/issues/2071


joedborg (approved):
LGTM (I thought we already merged this 😂)

ktsakalozos:
Yes we missed merging that fix. Should be fixed on all 1.21 channels soon but I think we should also back port this to the 1.20, 1.19 and 1.18.

balchua:
Thanks 👍

PR #2177, opened 2021-04-15
Adding KVM check and formatting with Black

joedborg:
After testing this with a host with and without KVM enabled, we should check to make sure.  Otherwise, Kata seems to enable correctly but, when you try and schedule anything using the Kata RuntimeClass, it will fail.

Also passed through Black formatter in order to pass tests.


ktsakalozos (approved):
LGTM +1

PR #2196, opened 2021-04-22
Update dqlite patch for 1.19.10

ktsakalozos:
We need to update this patch because it was not applying cleanly on 1.19.10


joedborg (approved):
Looks good to me

PR #2217, opened 2021-04-29
Do not re-issue certificates on clusters

ktsakalozos:
When in a cluster you do not want to re-issue certificates due to a network change. We place a lock file indicating if the certs should be reissued or not. Some work has gone into the tests to make sure this lock exists.


joedborg (approved):
LGTM

PR #2286, opened 2021-05-25
Get containerd confined

ktsakalozos:

### DIFF
--- a/microk8s-resources/wrappers/run-containerd-with-args
+++ b/microk8s-resources/wrappers/run-containerd-with-args
@@ -2,11 +2,6 @@
 
 set -ex
 
-# Re-exec outside of apparmor confinement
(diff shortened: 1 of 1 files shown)


microk8s-resources/wrappers/run-containerd-with-args
```diff
@@ -2,11 +2,6 @@
 
 set -ex
 
-# Re-exec outside of apparmor confinement
-if [ -d /sys/kernel/security/apparmor ] && [ "$(cat /proc/self/attr/current)" != "unconfined" ]; then
-    exec aa-exec -p unconfined -- "$0" "$@"
-fi
-
```
mardy, line 9: Shouldn't you remove also the line
```bash
apparmor_parser -r $SNAP/containerd-profile
```
below?

joedborg: We were asked to apply our own profile by the security team.  If this has changed, we can remove it.

joedborg (approved):

PR #2287, opened 2021-05-25
snap: remove patches section from snapcraft file

mardy:
All patches were removed in commit
478e80939311416292e7c10d12bd0a5a9fed6a8a, and snapcraft now complains
that the path is not found.


ktsakalozos (approved):
LGTM +1

PR #2292, opened 2021-05-27
microk8s-resources: simplify the first-boot detection

mardy:
The previous code required higher privileges (access to /proc/1/ is
generally not granted to confined applications) and was overcomplicated:
if we only need to know if this is the first execution after boot, then
we can create a sentinel file under /tmp and rely on the fact that /tmp
is cleared after every boot.


ktsakalozos (approved):
LGTM +1

PR #2294, opened 2021-05-27
snap: grant kubectl access to networking

mardy:
The kubectl program opens a socket to communicate with localhost.
For the time being, let's use the docker-support interface to provide
this access. It may be that this interface is overkill and that kubectl
requires much less permissions, but that will be tracked separately.


ktsakalozos (approved):
LGTM +1

mardy:
I created https://github.com/ubuntu/microk8s/issues/2295 to track the evolution of this.

Issue #2164, opened 2021-04-11
Cluster join no longer works after enabling Cilium


ressu:
During the setup process, Cilium addon disables the cni.yaml file for microk8s. This causes cluster join to fail due to CNI autodetection. This code path has plenty of comments about Calico specific operations, so it seems that Cilium needs some work on this front.

```
Apr 11 14:04:43 wanted-wolf microk8s.daemon-cluster-agent[1598267]: Waiting for access to cluster. [2021-04-11 14:04:43,340] ERROR in app: Exception on /cluster/api/v2.0/join [POST]
Apr 11 14:04:43 wanted-wolf microk8s.daemon-cluster-agent[1598267]: Traceback (most recent call last):
Apr 11 14:04:43 wanted-wolf microk8s.daemon-cluster-agent[1598267]:   File "/snap/microk8s/2094/lib/python3.5/site-packages/flask/app.py", line 2447, in wsgi_app
Apr 11 14:04:43 wanted-wolf microk8s.daemon-cluster-agent[1598267]:     response = self.full_dispatch_request()
Apr 11 14:04:43 wanted-wolf microk8s.daemon-cluster-agent[1598267]:   File "/snap/microk8s/2094/lib/python3.5/site-packages/flask/app.py", line 1952, in full_dispatch_request
Apr 11 14:04:43 wanted-wolf microk8s.daemon-cluster-agent[1598267]:     rv = self.handle_user_exception(e)
Apr 11 14:04:43 wanted-wolf microk8s.daemon-cluster-agent[1598267]:   File "/snap/microk8s/2094/lib/python3.5/site-packages/flask/app.py", line 1821, in handle_user_exception
Apr 11 14:04:43 wanted-wolf microk8s.daemon-cluster-agent[1598267]:     reraise(exc_type, exc_value, tb)
Apr 11 14:04:43 wanted-wolf microk8s.daemon-cluster-agent[1598267]:   File "/snap/microk8s/2094/lib/python3.5/site-packages/flask/_compat.py", line 39, in reraise
Apr 11 14:04:43 wanted-wolf microk8s.daemon-cluster-agent[1598267]:     raise value
Apr 11 14:04:43 wanted-wolf microk8s.daemon-cluster-agent[1598267]:   File "/snap/microk8s/2094/lib/python3.5/site-packages/flask/app.py", line 1950, in full_dispatch_request
[...]

PR #2299, opened 2021-05-29
First issue clusterjoin

Jorgeewa:
I am a first time contributor. This ticket fixes this issue: https://github.com/ubuntu/microk8s/issues/2164

I have had some issues myself.
1) When cilium addon is enabled and you try to join a cluster after the failure it starts giving an error: `connection to server 127.0.0.1:16443 was refused`
2) I have permission issues with my ubuntu vm in my mac os when I was testing but once I get feedback on the ticket, I will be sure to find a way to do this test.

In the ticket, I create a function to identify if cilium is enabled, I warn the user and let it fail gracefully. I had a problem selecting exactly where to make this check. First I made it inside the offending function but like pointed out above somewhere along the transaction, it becomes impossible to query kubectl because of the connectivity issues.

### DIFF
--- a/microk8s-resources/wrappers/microk8s-add-node.wrapper
+++ b/microk8s-resources/wrappers/microk8s-add-node.wrapper
@@ -17,6 +17,11 @@ then
   exit 1
 fi
 
+if [ -e  "$SNAP_DATA/args/cni-network/cni.yaml.disabled" ] && [ -e "$SNAP_DATA/args/cni-network/05-cilium-cni.conf" ] ; then 
+  echo "Cilium is enabled and is not suitable for multi node clusters."
+  exit 1
+fi
+
 if echo "$*" | grep -q -- 'help'; then
   # Call add_token.py help
   LD_LIBRARY_PATH=$IN_SNAP_LD_LIBRARY_PATH ${SNAP}/usr/bin/python3 ${SNAP}/scripts/cluster/add_token.py --help
--- a/microk8s-resources/wrappers/microk8s-join.wrapper
+++ b/microk8s-resources/wrappers/microk8s-join.wrapper
@@ -9,6 +9,11 @@ export PYTHONNOUSERSITE=false
 
 source $SNAP/actions/common/utils.sh
 
+if [ -e  "$SNAP_DATA/args/cni-network/cni.yaml.disabled" ] && [ -e "$SNAP_DATA/args/cni-network/05-cilium-cni.conf" ] ; then 
+  echo "Cilium is enabled and is not suitable for multi node clusters."
+  exit 1
+fi
+
 exit_if_no_permissions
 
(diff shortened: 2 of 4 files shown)


ktsakalozos:
Thank you for this PR @Jorgeewa. Let me first start with a description on what the code does when we join two nodes.

1. The user calls `microk8s add-node` on a node already in the cluster. 
The `microk8s-add-node.wrapper` [1] is called in order to produce a token and print out the command the user has to issue on the joining node.

2. The user calls `microk8s join <connection string>` on the joining node.
The `join.py` [2] script in called. `join.py` script detects if the joining node is a dqlite or etcd based node and calls the respective join endpoint of the `agent.py` [3] to fetch all the information needed to join the cluster.

Our goal is to inform the user that either of the nodes is running `cilium` and therefore clustering is not supported (yet). We will need to place check/guards in both of the above steps. We should not outpu a joining connection string in the case where the cluster runs `cilium` on the first step. We should fail gracefully the joining process in step two if the node is running cilium.

Let's see how to detect that cilium is enabled. The `cilium` addon [4] renames the `$SNAP_DATA/args/cni-network/cni.yaml` to `$SNAP_DATA/args/cni-network/cni.yaml.disabled` and also creates `$SNAP_DATA/args/cni-network/05-cilium-cni.conf`. To detect `cilium` we can do something similar to this:
```
# in bash
if ! [ -e  "$SNAP_DATA/args/cni-network/cni.yaml" ] && [ -e "$SNAP_DATA/args/cni-network/05-cilium-cni.conf" ] ; then 
  echo "Cilium is enabled"
fi

# in python
snapdata_path = os.environ.get("SNAP_DATA")
cni_yaml = "{}/args/cni-network/cni.yaml".format(snapdata_path)
cilium_conf = "{}args/cni-network/05-cilium-cni.conf".format(snapdata_path)
if not os.path.exists(cni_yaml) and os.path.exists(cilium_conf):
    print("Cilium is enabled")
```

We need a check and a proper error message in [5] to stop users from trying to produce connection strings for a cilium cluster. We need a check and a proper error message in [6] so we do not allow connecting to a cluster from a node with cilium enabled.


[1] https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/microk8s-add-node.wrapper
[2] https://github.com/ubuntu/microk8s/blob/master/scripts/cluster/join.py
[3] https://github.com/ubuntu/microk8s/blob/master/scripts/cluster/agent.py
[4] https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/enable.cilium.sh
[5] https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/microk8s-add-node.wrapper#L35
[6] https://github.com/ubuntu/microk8s/blob/master/scripts/cluster/join.py#L1057

ktsakalozos:
@Jorgeewa am closing this PR in favor of https://github.com/ubuntu/microk8s/pull/2299

I needed to do a few minor changes but it seems I do not have enough permissions to do so, have a look at https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

PR #2307, opened 2021-06-01
Update Kubeflow Revisions

DomFleischmann:
These new bundles contain several important bug fixes.

More information [here](https://github.com/canonical/bundle-kubeflow/pull/362)

### DIFF
--- a/microk8s-resources/actions/enable.kubeflow.sh
+++ b/microk8s-resources/actions/enable.kubeflow.sh
@@ -431,7 +431,7 @@ def print_info(hostname, password):
 @click.command()
 @click.option(
     "--bundle",
-    default="cs:kubeflow-267",
+    default="cs:kubeflow-269",
     help="The Kubeflow bundle to deploy. Can be one of full, lite, edge, or a charm store URL.",
 )
 @click.option(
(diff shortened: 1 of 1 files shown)


microk8s-resources/actions/enable.kubeflow.sh
```diff
@@ -475,9 +475,9 @@ def kubeflow(bundle, channel, debug, hostname, ignore_min_mem, no_proxy, passwor
     # user to specify a full charm store URL if they'd like, such as
     # `cs:kubeflow-lite-123`.
     if bundle == "full":
-        bundle = "cs:kubeflow-267"
+        bundle = "cs:kubeflow-269"
     elif bundle == "lite":
-        bundle = "cs:kubeflow-lite-51"
+        bundle = "cs:kubeflow-lite-53"
     elif bundle == "edge":
         bundle = "cs:kubeflow-edge-44"
```
knkski, line 482: I see `cs:kubeflow-edge` has a revision 45 that's on edge, and not yet pushed to stable. Should we include that revision in this update?

DomFleischmann: I'm not entirely sure where that revision is coming from, as the most recent changes weren't for charms present in the edge bundle.

PR #2316, opened 2021-06-03
snap: add kernel-module-control plug to the apiserver

mardy:
This gets rid of these two AppArmor errors:

    AVC apparmor="DENIED" operation="capable" profile="snap.microk8s.daemon-apiserver-kicker" pid=46846 comm="ip" capability=16  capname="sys_module"
    AVC apparmor="DENIED" operation="capable" profile="snap.microk8s.daemon-apiserver-kicker" pid=36170 comm="ifconfig" capability=16  capname="sys_module"

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -234,6 +234,7 @@ apps:
     command: apiservice-kicker
     daemon: simple
     plugs:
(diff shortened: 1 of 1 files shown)


ktsakalozos:
LGTM +1

PR #2319, opened 2021-06-04
coredns: allow privilege escalation

mardy:
Setting this flag to "true" causes runc not to set the NO_NEW_PRIVS
flag on the coredns. The reason why we don't want the NO_NEW_PRIVS flag
set on our process is that this flag does not play well with AppArmor,
and having it set prevents a process from switching to a different
AppArmor profile.


ktsakalozos:
I am going to merge this although I believe the issue is also with the kernel version.

PR #2345, opened 2021-06-17
Remove useless iptables rule

jungeonkim:
By default, traffic is handled by the iptables rule generated by calico.
And this rule blocks working NetworkPolicy when FELIX_CHAININSERTMODE is Append.

### DIFF
--- a/microk8s-resources/wrappers/run-kubelite-with-args
+++ b/microk8s-resources/wrappers/run-kubelite-with-args
@@ -55,25 +55,6 @@ then
   fi
 fi
 
-## Kubelet configuration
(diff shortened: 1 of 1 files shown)


ktsakalozos:
Hi @jungeonkim we put this rule there so packets can be forwarded properly. Can you give me a step by step setup or a script showing when this rule becomes a problem?

jungeonkim:
@ktsakalozos As I said in the original article, traffic is forwarded by the iptables rule that calico is created by default.
By default, packets are accepted or dropped by calico's rule before reaching this rule. So it's not a problem.
However, if  change ```FELIX_CHAININSERTMODE`` to ```Insert```` in order to apply the user-defined rule first, All packets are allowed by this rule before calico's rule is applied. So networkpolicy doesn't work properly.

PR #2346, opened 2021-06-18
Invert aa nnp order

mardy:
This is a partial revert of a previous commit, and bring in a change to runc.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -743,6 +743,9 @@ parts:
       (
         cd $GOPATH/src/github.com/opencontainers/runc
         git checkout ${RUNC_COMMIT}
+        git config user.email "microk8s-builder-bot@ubuntu.com"
+        git config user.name "MicroK8s builder bot"
+        git am $SNAPCRAFT_PART_SRC/patches/runc/*
         make BUILDTAGS='seccomp apparmor'
       )
       cp $GOPATH/src/github.com/opencontainers/runc/runc $SNAPCRAFT_PART_INSTALL/bin/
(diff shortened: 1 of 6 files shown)


ktsakalozos:
@magne thank you for this PR.

The build is failing for me, I had to add the git user:
```
diff --git a/snap/snapcraft.yaml b/snap/snapcraft.yaml
index ff2dd9b..26e5894 100644
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -743,6 +743,8 @@ parts:
       (
         cd $GOPATH/src/github.com/opencontainers/runc
         git checkout ${RUNC_COMMIT}
+        git config user.email "microk8s-builder-bot@ubuntu.com"
+        git config user.name "MicroK8s builder bot"
         git am $SNAPCRAFT_PART_SRC/patches/runc/*
         make BUILDTAGS='seccomp apparmor'
       )
```

Other than that it is ok. Could we also have someone from security to review the runc patch? I am not the right person to say if it ok or not.

snap/snapcraft.yaml
```diff
@@ -743,6 +743,9 @@ parts:
       (
         cd $GOPATH/src/github.com/opencontainers/runc
         git checkout ${RUNC_COMMIT}
+        git config user.email "microk8s-builder-bot@ubuntu.com"
+        git config user.name "MicroK8s builder bot"
```
anonymouse64, line 747: I don't recall this being necessary, can you not just do `git apply ...` ? I suppose that gives you a dirty tree, does the makefile for runc not like building dirty trees?

anonymouse64 (approved):
the runc patch trivially looks good to me in that it does what we want to have runc do, I can't speak to whether it is fully correct or not for all runc use cases

Issue #2290, opened 2021-05-26
Additional setup when enabling gpu (avoid the crash loop)


treo:
I've spent an entire day trying to debug why enabling the GPU on 1.21 on Ubuntu 20.04 resulted in a crash loop. 

I tried different versions of the gpu operator by modifying the enable.gpu.sh script, and I looked at a lot of logs. 

In the end this issue had the solution: https://github.com/NVIDIA/gpu-operator/issues/173#issuecomment-804836922

For completeness sake:
```
# Update /etc/systemd/system/multi-user.target.wants/snap.microk8s.daemon-containerd.service
- Restart=on-failure
- Type=simple
+ Restart=always
+ Type=notify
+ Delegate=yes
+ KillMode=process

sudo systemctl daemon-reload
sudo systemctl restart snap.microk8s.daemon-containerd
```

As I expect that more people may run into this particular problem, it would be nice to document this in the troubleshooting section.

ktsakalozos:
Thank you for reporting this. There is some work we need to do before we support this form of `Type=notify` `KillMode=process` daemon.

The problem we face is that when the snap refreshes any pods mounting parts of the SNAP_DATA directory (`/var/snap/microk8s/current/...`) need to be recreated. The most important of the pods that need a restart is the calico ones. So, if in the future you see your deployment failing after a refresh make sure you try killing the calico pods.

treo:
Just to make sure I understood this correctly:

With the necessary changes (`Type=notify` `KillMode=process`), I should stop Microk8s on the node, do the refresh, and only start it again afterwards, because some pods (esp. the calico ones) need to remount their mounts in the `/var/snap/microk8s/current` directory.

ktsakalozos:
Let me first explain what happens during a refresh. If you list the contents of `/var/snap/microk8s/` you will see one or more directories named after MicroK8s revisions and a symlink (current) pointing to the current revision. Here is an example:
```
ls -lh /var/snap/microk8s/
total 11K
drwxr-xr-x 10 root root 10 May 28 08:01 2213
drwxr-xr-x  4 root root  4 May 28 08:01 common
lrwxrwxrwx  1 root root  4 May 28 08:01 current -> 2213
```
[...]

PR #2347, opened 2021-06-18
Reload cni when refreshing

ktsakalozos:
Set the containerd daemon type to notification. This does not stop the containers when containerd restarts. During a refresh we have to make sure the CNI pods get restarted because they mount locations of $SNAP_DATA.

Fixes: https://github.com/ubuntu/microk8s/issues/2290


joedborg (approved):
LGTM

PR #2383, opened 2021-06-28
This probably shouldn't be running on the actual master branch

joedborg:
This is running on the push to master and failing due to:
```
Run canonical/has-signed-canonical-cla@1.1.4
Installing python3-launchpadlib

Error: Cannot read property 'commits_url' of undefined
```

I don't think it's meant to be run on push, only PR.


ktsakalozos (approved):
LGTM +1

PR #2416, opened 2021-07-09
snap: declare the microk8s user and group

mardy:
**NOTE**: this depends on https://github.com/snapcore/snapcraft/pull/3545

With this declaration, snapd will automatically create a user and a
group called "snap_microk8s".

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -12,6 +12,9 @@ grade: stable
 confinement: strict
 base: core18
 
+system-usernames:
+  snap_microk8s: shared
+
 plugs:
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -12,6 +12,9 @@ grade: stable
 confinement: strict
 base: core18
 
+system-usernames:
+  snap_microk8s: shared
```
anonymouse64, line 16: you can make this work without the snapcraft change using passthrough

```suggestion
passthrough:
  system-usernames:
    snap_microk8s: shared
```

PR #2417, opened 2021-07-09
snap: declare the microk8s user and group

mardy:
With this declaration, snapd will automatically create a user and a
group called "snap_microk8s". We use the passthrough feature to force
snapcraft to accept this value even though it doesn't understand it.

Note: https://github.com/ubuntu/microk8s/pull/2416 would be the best solution, but it's currently blocked by https://github.com/snapcore/snapcraft/pull/3545. While we wait for that change to land, here's a more immediate solution.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -12,6 +12,10 @@ grade: stable
 confinement: strict
 base: core18
 
+system-usernames:
+  passthrough:
+    snap_microk8s: shared
+
 plugs:
   home-read-all:
     interface: home


snap/snapcraft.yaml
```diff
@@ -12,6 +12,10 @@ grade: stable
 confinement: strict
 base: core18
 
+system-usernames:
+  passthrough:
```
anonymouse64, line 16: ah I just proposed the inverse usage of passthrough on your other branch, but both should work I think

PR #2426, opened 2021-07-12
Add config and tips for use with UFW

### LINKED ISSUE #2418: apiserver blocked when running microk8s with ufw (or any firewall?) enabled
I'm trying to use microk8s with ufw enabled but hitting networking problems (pods cannot read the apiserver).  I believe #1712 saw the same issues I am encountering, but the resolution might not have gone far enough.  It appears we need additional guidance about ufw rules to allow for the network interfaces calico generates for every pod (or maybe there's a configuration to let calico manage this for us?)

## Reproduction

* `sudo ufw enable`
* `sudo snap install microk8s --classic --channel 1.21/stable`
* `microk8s start`
* monitor microk8s come up using `watch microk8s kubectl get all --all-namespaces`, paying attention to `pod/calico-kube-controllers` which will not fully start.  check `microk8s kubectl logs calico-kube-controllers-YOURID` to see that it struggles with networking

For an even more obvious failure, do `microk8s enable storage` and monitor the associated pod

## Description

As described in #1712, the calico-kube-controllers pod is unable to reach the apiserver and goes into a crash loop because ufw is blocking the connection.  This can be fixed by adding the `calico-kube-controllers-*` pod's network interface (randomly named `caliXXXXXXXXXXX`) by `sudo ufw allow in on caliXXXXXXXXXXX ; sudo ufw allow out on caliXXXXXXXXXXX`.  

This is fine for a one-off case, but calico creates a new interface device for each pod, so the user would need to repeat the process for every pod they generate if that pod needs API server access (and maybe other things?).  

## Possible solutions

* Using a wildcard in the ufw allow rule will catch all cali-generated interfaces (`sudo ufw allow in on cali+ ; sudo ufw allow out on cali+`).  This works, but is maybe too permissive (can someone take advantage by making similarly-named interfaces?)
* (my iptables knowledge is **very** limited, but...) cali auto-generates iptables rules that (I think) mark traffic on these auto-generated interfaces (ex: `-A cali-tw-cali7fba7a35b74 -m comment --comment "cali:dO9Vjxk60RO-1o0o" -j MARK --set-xmark 0x0/0x10000`).  Can we add an allow rule that allows only this marked traffic, rather than allowing all traffic in/out of `cali+`?

## Other notes:

`microk8s inspect` advises users do `sudo ufw allow in on vxlan.calico && sudo ufw allow out on vxlan.calico` if ufw is on.  afaict this doesn't do anything (at least not for my configuration).  If I allow on `vxlan.calico` the problem persists, and if I instead enable the `caliX` interfaces (or `cali+`) in ufw everything appears to work normally.  Maybe there's a misconfiguration somewhere and the intent was things go through `vxlan.calico` but they're running elsewhere?  Or maybe this is just stale advice?  Either way, if we find a better config we should update `microk8s inspect` to give that advice.

ca-scribner:
Adds:
* UFW rules for network devices when UFW is detected to make networking work out of the box on systems using UFW
* Warning about missing UFW rules in `microk8s inspect` when rules are missing 

Closes #2418

### DIFF
--- a/microk8s-resources/wrappers/run-kubelite-with-args
+++ b/microk8s-resources/wrappers/run-kubelite-with-args
@@ -74,6 +74,16 @@ then
   fi
 fi
 
+#UFW configuration
+ufw=$(ufw status)
+if echo $ufw | grep -q "Status: active"
+then
+  # These succeed regardless of whether the rule exists already or not
+  echo "Found enabled UFW: adding rules to allow in/out traffic on 'cali+' devices"
+  ufw allow in on cali+
+  ufw allow out on cali+
+fi
+
 # wait for containerd socket
 if grep -e "--address " $SNAP_DATA/args/containerd &> /dev/null
 then
--- a/microk8s-resources/wrappers/run-kubelite-with-args
+++ b/microk8s-resources/wrappers/run-kubelite-with-args
@@ -76,10 +76,13 @@ fi
 
 #UFW configuration
 ufw=$(ufw status)
-if echo $ufw | grep -q "Status: active"
+if echo $ufw | grep -q "Status: active" &&
+   ! [ -e ${SNAP_DATA}/var/lock/skip.ufw ]
 then
   # These succeed regardless of whether the rule exists already or not
-  echo "Found enabled UFW: adding rules to allow in/out traffic on 'cali+' devices"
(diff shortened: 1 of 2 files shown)


ktsakalozos (approved):
LGTM +1 thank you for this work @ca-scribner

PR #2427, opened 2021-07-13
Use libsqlite-dev from the sqlite build

ktsakalozos:
Fixes the following error:
```
dqlite: symbol lookup error: /snap/microk8s/2291/bin/dqlite: undefined symbol: sqlite3_system_errno 
```


joedborg (approved):
LGTM

PR #2429, opened 2021-07-13
Use libsqlite-dev from the sqlite build

ktsakalozos:
We want to build the dqlite-client using the libs from the sqlite we build.


joedborg (approved):
LGTM

PR #2444, opened 2021-07-16
Worker only node with dqlite

ktsakalozos:
`microk8s join` command has an additional `--worker` option to add the node as a worker node not running the control plane.

Kubelet and kube-proxy contact a Traefik loadbalancer running locally on port 16443 (same as the API server) and Traefik forwards the requests towards a set of configured API servers.

Kubelet and kube-proxy are authenticated based on x509 certificates so it does not matter which master node they contact.

Nodes running the control plane are labeled with `node.kubernetes.io/microk8s-controlplane=microk8s-controlplane`, worker nodes are labeled with `node.kubernetes.io/microk8s-worker=microk8s-worker`.

Nodes labeled with `node.kubernetes.io/microk8s-controlplane=microk8s-controlplane` at the time of the joining operation will be added to the Traefik loadbalancer.

The leave, join and reset operations were refactored to use click.

This PR is available in the channel branch `latest/edge/worker`.


microk8s-resources/client-x509.config.template
```diff
@@ -0,0 +1,19 @@
...
+    user: NAME
+  name: microk8s
+current-context: microk8s
+kind: Config
+preferences: {}
+users:
+- name: NAME
+  user:
+    client-certificate: PATHTOCERT
+    client-key: PATHTOKEYCERT
```
neoaggelos, line 19: final newline?

scripts/cluster/join.py
```diff
@@ -120,6 +140,7 @@ def get_connection_info(
     :param cluster_type: the type of cluster we want to join, etcd or dqlite
     :param verify_peer: flag indicating if we should verify peers certificate
     :param fingerprint: the certificate fingerprint we expect from the peer
+    :param worker: this is a worker only node
```
joedborg, line 143: In the future and in lieu of type hinting, it would be really nice to put the expected type here.  This is expected with the Sphinx style docstrings we've settled on (https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html).

joedborg:
@ktsakalozos all looks good to me.  I have seen a couple of things that I'm not sure will work under strict.  Has this been borne in mind at all?

ktsakalozos:
> @ktsakalozos all looks good to me. I have seen a couple of things that I'm not sure will work under strict. Has this been borne in mind at all?

@joedborg could you point me to parts of code that would become an issue in strict confinement? I do not think we will have problems porting this work to strict but I might have missed something.

scripts/cluster/join.py
ktsakalozos: +1, I am looking forward to the PR that adds strong typing.

scripts/cluster/join.py
```diff
@@ -302,6 +408,66 @@ def update_kubeproxy(token, ca, master_ip, api_port, hostname_override):
...
+
+    :param token: the token to be in the kubeconfig
+    :param ca: the ca
+    :param master_ip: the master node IP
+    :param master_port: the master node port where the cluster agent listens
+    :param hostname_override: the hostname override in case the hostname is not resolvable
+    """
+    proxy_token = "{}-proxy".format(token)
+    traefik_port = get_traefik_port()
+    cert = get_client_cert(master_ip, master_port, proxy_token, "kubeproxy")
```
neoaggelos, line 423: This is needed for RBAC to work.

```suggestion
    cert = get_client_cert(master_ip, master_port, proxy_token, "system:kube-proxy")
```

scripts/cluster/join.py
```diff
@@ -302,6 +408,66 @@ def update_kubeproxy(token, ca, master_ip, api_port, hostname_override):
...
+    :param token: the token to be in the kubeconfig
+    :param ca: the ca
+    :param master_ip: the master node IP
+    :param master_port: the master node port where the cluster agent listens
+    :param api_port: the API server port
+    :param hostname_override: the hostname override in case the hostname is not resolvable
+    """
+    traefik_port = get_traefik_port()
+    kubelet_token = "{}-kubelet".format(token)
+    cert = get_client_cert(master_ip, master_port, kubelet_token, "kubelet", "system:nodes")
```
neoaggelos, line 452: Required for RBAC

```suggestion
    cert = get_client_cert(master_ip, master_port, kubelet_token, "system:node:{}".format(get_my_hostname() "system:nodes")
```

neoaggelos (changes requested):

scripts/cluster/join.py
neoaggelos: LGTM, but up for doing something for the filename, just in case that `:` comes back to haunt us in a few years time.

Up to you to update as you see fit.

scripts/cluster/join.py
neoaggelos: Resolving, see other comment

neoaggelos (approved):
See my minor about the filenames, other than that LGTM, nice work

neoaggelos (approved):
Make sure to fix the minor linter issues

ktsakalozos:
Thank you for the reviews. I am going to merge this so we get it into the 1.23 rc. Bug fixes from now on.

PR #2467, opened 2021-07-23
Update addon versions reported by the status command

ktsakalozos:
We seem to have updated the addons but forgot to update their version numbers.


balchua (approved):
Oh yeah. Thanks 👍

PR #2470, opened 2021-07-25
Migrate Cilium to install/uninstall via new CLI

joestringer:
The Cilium community has developed a new CLI for installing, operating
and debugging Kubernetes clusters with Cilium installed[0]. We can reuse
that CLI here for a much simpler, less error-prone upgrade process.

For the most part, the new CLI almost removes the need for these
microk8s-specific scripts. However, the CLI cannot know how to uninstall
existing CNIs in a microk8s-specific environment, so we still need some
logic here to clean that up in order for Cilium to successfully deploy
and avoid conflicts with existing components.

A few notes about implementation choices:
* Remove previous CNIs and CNI config before initiating Cilium install
  to ensure that Cilium manages pods after the cilium install command
  has been run.
* disable.cilium.sh is handling disabling the old version during
  microk8s upgrade.
* Use 'rm -rf -- $directories" to reduce likelihood of removing bad
  directories, rather than 'rm -rf $directories'

[0] https://github.com/cilium/cilium-cli/

### DIFF
--- a/microk8s-resources/actions/disable.cilium.sh
+++ b/microk8s-resources/actions/disable.cilium.sh
@@ -6,10 +6,15 @@ source $SNAP/actions/common/utils.sh
 
 echo "Disabling Cilium"
 
-"$SNAP/kubectl" "--kubeconfig=$SNAP_DATA/credentials/client.config" delete -f "$SNAP_DATA/actions/cilium.yaml"
+# Upgrade from v1.21 or earlier.
+if [ -e "$SNAP_DATA/actions/cilium.yaml" ]; then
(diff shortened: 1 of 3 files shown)


joestringer:
This PR proposes to modify code that has now [moved over to the microk8s-addons project](https://github.com/canonical/microk8s-addons/tree/main/addons/cilium). I still think that it makes sense for microk8s to adopt these changes, however I haven't heard anything from microk8s maintainers about this and I no longer use microk8s for my own Cilium development, so I do not intend to further pursue this improvement to the microk8s-cilium integration.

Feel free to reach out if you're interested in pushing for inclusion of this code into microk8s, maybe I can help out with that process.

Issue #2496, opened 2021-08-05
Kubeflow addon not yet available on Kubernetes v1.22


ktsakalozos:
At the day of the Kubernetes v1.22 release the kubeflow addon is failing. The `seldon-controller-manager` is crashlooping with:
```
ubuntu@ip-172-31-46-67:~$ microk8s.kubectl logs -n kubeflow          pod/seldon-controller-manager-d797647d9-9pnbm
{"level":"info","ts":1628141400.383196,"logger":"setup","msg":"Intializing operator"}
{"level":"info","ts":1628141400.3900747,"logger":"setup","msg":"CRD not found - trying to create"}
{"level":"error","ts":1628141400.4404993,"logger":"setup","msg":"unable to initialise operator","error":"the server could not find the requested resource","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\nmain.main\n\t/workspace/main.go:149\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"}
```

PR #2497, opened 2021-08-05
Disable Kubeflow addon for the v1.22 release

ktsakalozos:
Kubeflow is not available on K8s v1.22 yet so we need to fail gracefully and direct users to a github issue where they can see the progress we are making.

Due to https://github.com/ubuntu/microk8s/issues/2496


DomFleischmann (approved):
LGTM

PR #2503, opened 2021-08-06
Do not run cilium test on lxc

ktsakalozos:
Cilium is complaining when enabled on lxc:
```
kube-system   pod/cilium-bzzqp                               0/1     Init:CreateContainerError
```
and 
```
(combined from similar events): Error: failed to generate container "95c9998218f6b6bc29c56b4d08a4b64de4b705f786d86e7bf8ebd599ca155328" spec: failed to generate spec: path "/run/cilium/cgroupv2" is mounted on "/run/cilium/cgroupv2" but it is not a shared or slave mount
```

This PR runs the cilium test on GH actions and not on lxc containers.

### DIFF
--- a/tests/test-addons.py
+++ b/tests/test-addons.py
@@ -153,31 +153,6 @@ class TestAddons(object):
         """
 
     @pytest.mark.skipif(
-        platform.machine() != "x86_64", reason="Cilium tests are only relevant in x86 architectures"
-    )
-    @pytest.mark.skipif(
-        os.environ.get("UNDER_TIME_PRESSURE") == "True",
-        reason="Skipping cilium tests as we are under time pressure",
-    )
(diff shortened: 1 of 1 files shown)


tests/test-addons.py
```diff
@@ -251,14 +251,17 @@ def test_prometheus(self):
...
-        reason="Skipping cilium tests as we are under time pressure",
-    )
     def test_cilium(self):
         """
         Sets up and validates Cilium.
         """
+        try:
+            check_call("sudo grep -E lxc /proc/1/environ".split())
+            print("Cilium test skipped on lxc containers")
+            return True
```
johnsca, line 261: Why not use `pytest.skip(msg)` here?

johnsca (approved):
LGTM

Issue #2488, opened 2021-08-03
Service snap.microk8s.daemon-kubelite is not running


golgotow:
Hello,

3 node microk8s cluster on raspberrypi 4 8G 

after OS upgrade one by one with drain nodes before reboot.

microk8s stop to work, 2 first node upgrade did not made problem, after 3 node was update this issue appear and screw whole cluster.

Can you advice if this is bug or some know problem ?

root@rpi8g-2:/tmp/inspection-report# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.2 LTS
Release:	20.04
Codename:	focal

Inspecting Certificates
Inspecting services
  Service snap.microk8s.daemon-cluster-agent is running
  Service snap.microk8s.daemon-containerd is running
  Service snap.microk8s.daemon-apiserver-kicker is running
 FAIL:  Service snap.microk8s.daemon-kubelite is not running
For more details look at: sudo journalctl -u snap.microk8s.daemon-kubelite
  Copy service arguments to the final report tarball
Inspecting AppArmor configuration
Gathering system information
  Copy processes list to the final report tarball
  Copy snap list to the final report tarball
  Copy VM name (or none) to the final report tarball
  Copy disk usage information to the final report tarball
  Copy memory usage information to the final report tarball
  Copy server uptime to the final report tarball
  Copy current linux distribution to the final report tarball
  Copy openSSL information to the final report tarball
  Copy network configuration to the final report tarball
Inspecting kubernetes cluster
  Inspect kubernetes cluster
Inspecting juju
  Inspect Juju
Inspecting kubeflow
  Inspect Kubeflow

Building the report tarball
  Report tarball is at /var/snap/microk8s/2347/inspection-report-20210803_090500.tar.gz

Best regards
[inspection-report-20210803_090500.tar.gz](https://github.com/ubuntu/microk8s/files/6921990/inspection-report-20210803_090500.tar.gz)

golgotow:
] API Server exited create node: failed to set bind address "@snap.microk8s.dqlite-8278944527939517719": 1

balchua:
I have seen this issue before only on nodes with glusterfs running on it.  If u stop gluster and then run MicroK8s does it work?

golgotow:
yes if I remove glusterfs this start to works... do you know any workaround to made glusterfs work with microk8s? or why this happen, as this is interesting stuff

balchua:
[...]

PR #2519, opened 2021-08-11
Fail gracefully in setting ufw rules

ktsakalozos:
Handles the case where UFW is failing for some reason. See https://github.com/ubuntu/microk8s/issues/2488#issuecomment-896827686


joedborg (approved):
LGTM

PR #2522, opened 2021-08-13
snap: bind-mount the whole /etc/service/enabled/ directory

mardy:
There are other lock files that are used, such as

    /etc/service/enabled/monitor-addresses/supervise/lock
    /etc/service/enabled/allocate-tunnel-addrs/supervise/lock

It's probably a good idea to bind-mount the whole parent directory.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -443,8 +443,8 @@ passthrough:
       bind: $SNAP_DATA/kubelet
     /var/lib/kube-proxy:
       bind: $SNAP_DATA/kube-proxy
-    /etc/service/enabled/felix/supervise/lock:
-      bind-file: $SNAP_COMMON/etc/service/enabled/felix/supervise/lock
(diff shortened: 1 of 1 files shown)


ktsakalozos:
This seems to be needed since we get the error:
```
[50646.708860] audit: type=1400 audit(1630504511.293:154099): apparmor="DENIED" operation="file_lock" profile="snap.microk8s.daemon-containerd" name="/etc/service/enabled/monitor-addresses/supervise/lock" pid=1271136 comm="runsv" requested_mask="k" denied_mask="k" fsuid=0 ouid=0
[50646.708865] audit: type=1400 audit(1630504511.293:154100): apparmor="DENIED" operation="file_lock" profile="snap.microk8s.daemon-containerd" name="/etc/service/enabled/allocate-tunnel-addrs/supervise/lock" pid=1271137 comm="runsv" requested_mask="k" denied_mask="k" fsuid=0 ouid=0
```

PR #2537, opened 2021-08-25
snap/hooks/configure: ensure that $SNAP_DATA/certs/ exists

mardy:
We do want to have a $SNAP_DATA/certs/ directory in any case because of
`csr.conf.template`. Since the order of the operations is not important,
change it slightly to ensure that the directory is created before
patching the other certificates.

Fixes: https://bugs.launchpad.net/snapd/+bug/1940985


ktsakalozos (approved):
LGTM +1 thank you @mardy

PR #2568, opened 2021-09-10
Copy users to the new snap_microk8s group

ktsakalozos:
This is to prepare the transition to the strict snap where we have the new group snap_microk8s


evilnick:
Just for my own benefit, this isn't actually migrating per se is it? I mean, the user will exist in both groups?

ktsakalozos:
@evilnick you are right. I updated the title of the PR.

joedborg (approved):
LGTM

PR #2588, opened 2021-09-23
Build k8s binaries with go v1.17

ktsakalozos:
Update the k8s build process to use go 1.17. This is needed for v1.23.


joedborg (approved):
Looks good to me

PR #2590, opened 2021-09-23
Upload AppArmor denials post tests

joedborg:
Upload any AppArmor denials post strict tests.


ktsakalozos:
Lets have them in for now, although note that the full dmesg output is in the inspection tarball

PR #2624, opened 2021-10-04
Bump containerd and runc versions

joedborg:
Containerd is getting behind on patch released and runc is now out of RC.


balchua:
The containerd 1.5 has a new and i think better way to manage cri configurations.

https://github.com/containerd/containerd/blob/main/docs/cri/config.md
We can point to a location where the registries config are located.

PR #2649, opened 2021-10-12
Fix Kubeflow cleanup for Juju 2.8+

knkski:
Fixes issue where admission-webhook would leave behind a webhook that would attempt to talk to the non-existent admission-webhook service, blocking all attempts to create a new Pod.


knkski:
@ktsakalozos I'd like to get this into master, since 1.22+ will still need this fix. After it lands there, I was planning on backporting PRs, yeah.

PR #2692, opened 2021-10-29
snap: Remove kernel-module-control interface

mardy:
I added this interface just to suppress some AppArmor denials, which were anyway harmless. We are about to land a change in snapd which will suppress these denials anyway (https://github.com/snapcore/snapd/pull/10933) and will not require further changes in microk8s.

It's up to you whether you want to merge this right now or wait for a newer snapd. As I said, these denials are cosmetic, so I'd vote for merging this now and concede that we'll see a couple of denials until a newer snapd fixes them.


joedborg:
Thanks @mardy.  Moved to https://github.com/ubuntu/microk8s/pull/2861 so that I could update the PR.

PR #2174, opened 2021-04-14
Kata containers addon

ktsakalozos:
[Kata Containers](https://katacontainers.io/), is used to build a secure container runtime with lightweight virtual machines that feel and perform like containers, but provide stronger workload isolation using hardware virtualization technology as a second layer of defense. You can enable kata support with: 

``` microk8s enable kata``` 

The addon adds the `kata` `runtimeClassName` that allows you to specify what workloads should be started in Kata containers. For instance, the following manifest starts nginx in a Kata container:

```
apiVersion: v1
kind: Pod
metadata:
  labels:
    app: kata
  name: nginx-kata
spec:
  runtimeClassName: kata
  containers:
    - name: nginx
      image: nginx
```

By default the addon will install the Kata runtime via the [kata-containers snap](https://snapcraft.io/kata-containers). Alternatively, you can set the path of the where kata runtime is installed using the `--runtime-path` argument. The path you provide should include the `kata-runtime` binary:

```
microk8s enable kata --runtimepath=/path/to/runtime
```

To disable Kata support: 

```
microk8s disable kata
``` 

Note that disabling Kata will not remove the runtime from your system.

PR #2696, opened 2021-11-01
 Points Kubeflow's check_status value to kubeflow.enabled lock file 

### LINKED ISSUE #2695: [kubeflow add-on] microk8s status does not show kubeflow as enabled
The `kubeflow` add-on is not listed under the "enabled" section when running `microk8s status`. All Kubernetes resources are created, and Kubeflow seems to be working properly, it is just the output of the command that seems off.

#### Steps to reproduce
```bash
$ sudo snap install microk8s --classic --channel=1.21/stable
$ microk8s status --wait-ready
$ microk8s enable kubeflow
$ microk8s status
microk8s is running
high-availability: no
  datastore master nodes: 127.0.0.1:19001
  datastore standby nodes: none
addons:
  enabled:
    dns                  # CoreDNS
    ha-cluster           # Configure high availability on the current node
    ingress              # Ingress controller for external access
    istio                # Core Istio service mesh services
    metallb              # Loadbalancer for your Kubernetes cluster
    storage              # Storage class; allocates storage from host directory
  disabled:
    ambassador           # Ambassador API Gateway and Ingress
    cilium               # SDN, fast with full network policy
    dashboard            # The Kubernetes dashboard
    fluentd              # Elasticsearch-Fluentd-Kibana logging and monitoring
    gpu                  # Automatic enablement of Nvidia CUDA
    helm                 # Helm 2 - the package manager for Kubernetes
    helm3                # Helm 3 - Kubernetes package manager
    host-access          # Allow Pods connecting to Host services smoothly
    jaeger               # Kubernetes Jaeger operator with its simple config
    keda                 # Kubernetes-based Event Driven Autoscaling
    knative              # The Knative framework on Kubernetes.
    kubeflow             # Kubeflow for easy ML deployments
```
#### Possible cause
The add-on status is checked by [`status.py`](https://github.com/ubuntu/microk8s/blob/master/scripts/wrappers/status.py#L153) which relies on the value of `check_status` in [`addons-lists.yaml`](https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/addon-lists.yaml#L150). For kubeflow, it points to `pod/ambassador-operator-0`, an application that is not listed in [`bundle.yaml`](https://github.com/canonical/bundle-kubeflow/blob/master/bundle.yaml) and thus not deployed.

#### Possible solution
Point `check_status` to something else included in both [`bundle.yaml`](https://github.com/canonical/bundle-kubeflow/blob/master/bundle.yaml) and [`bundle-lite.yaml`](https://github.com/canonical/bundle-kubeflow/blob/master/bundle-lite.yaml). I propose `istio-ingressgateway-operator-0`

#### My setup
microk8s v1.21.5
Ubuntu 20.04 LTS
#### microk8s inspect
[inspection-report-20211031_210105.tar.gz](https://github.com/ubuntu/microk8s/files/7450527/inspection-report-20211031_210105.tar.gz)

DnPlas:
Creates lock file after all actions in enable.kubeflow.sh
and points check_status in addon-lists.yaml to that file.
This ensures Kubeflow addon is listed as enabled when
running microk8s status

Fixes #2695

### DIFF
--- a/microk8s-resources/actions/disable.kubeflow.sh
+++ b/microk8s-resources/actions/disable.kubeflow.sh
@@ -1,10 +1,20 @@
 #!/usr/bin/env python3
 
 import os
+import sys
 import subprocess
 
 import click
 
+def mark_kubeflow_disabled():
+    """Mark Kubeflow addon as disabled by removing kubeflow.enabled lock file."""
+    try:
+        snapdata_path = os.environ.get("SNAP_DATA")
+        lock_fname = "{}/var/lock/kubeflow.enabled".format(snapdata_path)
+        subprocess.call(['sudo', 'rm', lock_fname])
+    except (subprocess.CalledProcessError):
+        print("Failed to mark Kubeflow addon as disabled." )
+        sys.exit(4)
 
 @click.command()
 def kubeflow():
@@ -65,6 +75,8 @@ def kubeflow():
             )
         except subprocess.CalledProcessError:
             pass
+
+    mark_kubeflow_disabled()
 
     click.echo("Destruction complete.")
 
--- a/microk8s-resources/actions/enable.kubeflow.sh
+++ b/microk8s-resources/actions/enable.kubeflow.sh
@@ -29,6 +29,16 @@ CONNECTIVITY_CHECKS = [
 ]
 
 
+def mark_kubeflow_enabled():
+    """Mark Kubeflow addon as enabled by creating kubeflow.enabled."""
+    try:
+        snapdata_path = os.environ.get("SNAP_DATA")
+        lock_fname = "{}/var/lock/kubeflow.enabled".format(snapdata_path)
+        subprocess.call(['sudo', 'touch', lock_fname])
+    except (subprocess.CalledProcessError):
+        print("Failed to mark Kubeflow addon as enabled." )
+        sys.exit(4)
+
 def charm_exists(name: str):
     try:
         run("microk8s-juju.wrapper", "show-application", name, die=False)
@@ -519,6 +529,7 @@ def kubeflow(bundle, channel, debug, hostname, ignore_min_mem, no_proxy, passwor
 
     print_info(hostname, password)
 
+    mark_kubeflow_enabled()
 
 if __name__ == "__main__":
     kubeflow(prog_name="microk8s enable kubeflow", auto_envvar_prefix="KUBEFLOW")
--- a/microk8s-resources/wrappers/addon-lists.yaml
+++ b/microk8s-resources/wrappers/addon-lists.yaml
@@ -141,7 +141,7 @@ microk8s-addons:
     - name: "kubeflow"
       description: "Kubeflow for easy ML deployments"
       version: ""
-      check_status: "pod/ambassador-operator-0"
+      check_status: "{SNAP_DATA}/var/lock/kubeflow.enabled"
       supported_architectures:
         - amd64
 


ktsakalozos:
We may want to backport this PR to the branches where the kubeflow addon is available (1.21, 1.20,....)

DnPlas:
Hi @ktsakalozos I've rebased my dev branch and also changed the target branch to 1.21, which is the last supported version for Kubeflow. Note that I went with the approach of creating a lock file instead of just checking for a resource.

ktsakalozos:
@DnPlas I think we also need to remove the lock file when we disable the add on in https://github.com/ubuntu/microk8s/blob/1.21/microk8s-resources/actions/disable.kubeflow.sh

DnPlas:
> @DnPlas I think we also need to remove the lock file when we disable the add on in https://github.com/ubuntu/microk8s/blob/1.21/microk8s-resources/actions/disable.kubeflow.sh

Ah you are absolutely right. Let me add that.

microk8s-resources/actions/disable.kubeflow.sh
```diff
@@ -1,10 +1,20 @@
...
 import subprocess
 
 import click
 
+def mark_kubeflow_disabled():
+    """Mark Kubeflow addon as disabled by removing kubeflow.enabled lock file."""
+    try:
+        snapdata_path = os.environ.get("SNAP_DATA")
+        lock_fname = "{}/var/lock/kubeflow.enabled".format(snapdata_path)
+        subprocess.call(['sudo', 'rm', lock_fname])
```
addyess, line 14: you might consider python's `pathlib` here.  
I can't comment on the `sudo` here asn i don't know its relevance.  It seems odd to elevate to sudo in your code -- what if the user cannot `sudo`?   What if the user can but didn't know they were doing something needing sudo.  

is it ok to disable if it's already disabled?  I don't know in this context what the intent is - `rm missing_file` will raise an exception but the except case here doesn't indicate why.  

```python
from pathlib import Path

lock_fname = Path(snapdata_path) / "var" / "lock" / "kubeflow.enabled"
lock_fname.unlink(missing_ok=True)   
```

microk8s-resources/actions/enable.kubeflow.sh
```diff
@@ -29,6 +29,16 @@ CONNECTIVITY_CHECKS = [
 ]
 
 
+def mark_kubeflow_enabled():
+    """Mark Kubeflow addon as enabled by creating kubeflow.enabled."""
+    try:
+        snapdata_path = os.environ.get("SNAP_DATA")
+        lock_fname = "{}/var/lock/kubeflow.enabled".format(snapdata_path)
+        subprocess.call(['sudo', 'touch', lock_fname])
```
addyess, line 37: See comment above concerning pathlib.  There's also a ```Path.touch()``` method

addyess (commented):
Who should be the owner of this file you're creating?  Root?  the snap owner?

microk8s-resources/actions/disable.kubeflow.sh
DnPlas: > I can't comment on the sudo here as i don't know its relevance. It seems odd to elevate to sudo in your code -- what if the user cannot sudo? What if the user can but didn't know they were doing something needing sudo.

I followed what I thought was the standard, for kata-containers (see [PR#2174](https://github.com/ubuntu/microk8s/pull/2174)) it is done that way. @ktsakalozos may give us his input here.

> is it ok to disable if it's already disabled? I don't know in this context what the intent is - rm $SNAP_DATA/var/lock/kubeflow.enabled

What do you mean by "is it ok to disable if it's already disabled?"?

addyess: can someone run this back-to-back?  What would be expected to occur? wouldn't the `rm $lock_fname` fail throwing a subprocess error?

```bash
microk8s disable kubeflow
microk8s disable kubeflow
```

knkski: Personally, I'd expect a message like this:

https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/actions/enable.kubeflow.sh#L203-L212

i.e.

> Kubeflow has already been disabled.

DnPlas: I agree, running `microk8s disable <addon>` when the addon is already disabled is something we have to take care, though that is not exclusive to the lock file proposed in this PR, it applies to pretty much everything we run in the disable action. 
I see that microk8s handles this problem for other addons in [this](https://github.com/ubuntu/microk8s/blob/master/scripts/wrappers/common/utils.py#L275) check. I also see that kubeflow is excluded, see [L271](https://github.com/ubuntu/microk8s/blob/master/scripts/wrappers/common/utils.py#L271) in the same file. We could either do the check manually inside `disable.kubeflow.sh` or re-use what microk8s already does for other addons. I am open to both solutions, though I'd prefer the latter as it is more scalable.

addyess (approved):
i approve of this if there's future work to be performed.

Then, again, i cannot merge for you in this organization

knkski (approved):
LGTM, though @ktsakalozos should give final approval

Issue #2639, opened 2021-10-09
Windows 10 user's path variable gets "corrupted" after installation of Multipass/Microk8s


danifgxcom:
Hi Joe,

Maybe another issue with the installer in Windows 10. The installer's version is 2.2.0 and Multipass 1.7.2+.
I have tried a few times to install / uninstall MicroK8s and Multipass to identity the issue that I mentioned in the guidelines.

What I have seen from the very beginning is that the PATH variable for the user in Windows 10 gets somehow corrupted for the MicroK8s path. I attach a picture.
![image](https://user-images.githubusercontent.com/16970877/136657607-aba0f2f3-0a7c-41a3-bde3-f86f9b611bea.png)

It is a small thing, but just to let you know as people will be unable to use it as soon as they open a PS or CMD terminals.

Hope it helps.

Thanks.

P.S. I do not attach the 'microk8s inspect' tarball file as it is something related with the installer, and not with the MicroK8s itself.

Best regards,
   Dani.

PR #2704, opened 2021-11-01
Change method of setting PATH

joedborg:
Fixing #2639


joedborg:
Could reproduce before the update.  Have tested this locally and it seems to fix the issue reported.

ktsakalozos (approved):
LGTM +1

PR #2731, opened 2021-11-16
Disable profiling and reduce event TTL for Kubernetes services

neoaggelos:
## Summary

Update default API server arguments

- Disable profiling (--profiling=false)
- Reduce event TTL to 5 minutes (--event-ttl=5m)


ktsakalozos (approved):
LGTM +1 thank you @neoaggelos

Issue #2657, opened 2021-10-14
script/inspect.sh use of ifconfig/netstat commands


odufourc:
Hello everyone,

When playing with microk8s with its command inspect, I came accross some error on ifconfig and netstat generated files :
```
ubuntu@odc-k8s:~/logs/inspection-report/network$ cat ifconfig 
/snap/microk8s/2551/inspect.sh: line 56: ifconfig: command not found
ubuntu@odc-k8s:~/logs/inspection-report/network$ cat netstat 
/snap/microk8s/2551/inspect.sh: line 55: netstat: command not found
```
[inspection-report-20211008_133310.tar.gz](https://github.com/ubuntu/microk8s/files/7345235/inspection-report-20211008_133310.tar.gz)

These errors are happening because the package net-tools wasn't installed by default on my Ubuntu image.
But there are 2 things to consider :
* these 2 commands have been deprecated for a while since they aren't maintained anymore and have been replaced by iproute2 tools, which their equivalent would be : 
  * ifconfig --> ip addr
  * netstat -lpn --> ss -lpn
* on some other Linux distributions, they aren't installed anymore by default, but iproute2 is

I did a little bit of playing with various Linux distributions (Arch, Debian, Ubuntu, Fedora, Centos, Alpine, ...) :
- ip addr works everywhere
- ss doesn't work only in Alpine by default (iproute2 package must be installed) and busybox

I can work on a short pull-request, but I wanted your opinions first.
1) Shall I just replace ifconfig and netstat with their counterparts ?
2) Shall I test replace ifconfig but test if ss is available or use netstat otherwise ?

ktsakalozos:
Hi @odufourc, thank you for volunteering for this work, it has been in our backlog for a long time. I would prefer the first option to just replace the two utilities with their counterparts, I think it is cleaner. You could try to use those utilities from inside inside the snap (`/snap/microk8s/current`). If they are missing we could add them in the list [1].

[1] https://github.com/ubuntu/microk8s/blob/master/snap/snapcraft.yaml#L441

PR #2751, opened 2021-11-25
Replace depreated call from net-tools to iproute2

odufourc:
It consist basically of the replacement of the long deprecated ifconfig/netstat to ip addr/route/ss tools from iproute2.

The first commit fixes the issue #2657 .
While at it, it also takes care, in a second commit, of other calls of ifconfig and netstat in other scripts such as :  
* enable/disable host-access
* common/utils functions
* inside the wrapper "run-kubelite-with-args" which simply enable the host-access feature

With this, theoretically, it should be possible to remove net-tools from the snap build.

### DIFF
--- a/microk8s-resources/actions/enable.host-access.sh
+++ b/microk8s-resources/actions/enable.host-access.sh
@@ -30,7 +30,7 @@ fi
 echo "$IP_ADDRESS" > "${SNAP_DATA}/var/lock/host-access-enabled"
 echo "Setting ${IP_ADDRESS} as host-access"
 
-run_with_sudo "$SNAP/sbin/ifconfig" lo:microk8s "$IP_ADDRESS" up
+run_with_sudo "$SNAP/sbin/ip" addr add "$IP_ADDRESS"/32 dev lo label lo:microk8s
 
 echo "Host-access is enabled"
 
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -347,7 +347,7 @@ wait_for_service_shutdown() {
 
 get_default_ip() {
     # Get the IP of the default interface
-    local DEFAULT_INTERFACE="$($SNAP/bin/netstat -rn | $SNAP/bin/grep '^0.0.0.0' | $SNAP/usr/bin/gawk '{print $NF}' | head -1)"
+    local DEFAULT_INTERFACE="$($SNAP/sbin/ip route show default | $SNAP/usr/bin/gawk '{for(i=1; i<NF; i++) if($i=="dev") print$(i+1)}')"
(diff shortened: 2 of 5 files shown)


microk8s-resources/actions/enable.host-access.sh
```diff
@@ -30,7 +30,7 @@ fi
 echo "$IP_ADDRESS" > "${SNAP_DATA}/var/lock/host-access-enabled"
 echo "Setting ${IP_ADDRESS} as host-access"
 
-run_with_sudo "$SNAP/sbin/ifconfig" lo:microk8s "$IP_ADDRESS" up
+run_with_sudo "$SNAP/sbin/ip" addr add "$IP_ADDRESS"/32 dev lo label lo:microk8s
```
neoaggelos, line 33: `ifconfig lo:microk8s 10.0.1.1 up` assigns the IP address 10.0.1.1/**8** to the interface.

This is all link-local, so the subnet size does not matter.

neoaggelos: I would prefer to preserve the existing behaviour, but this is fine as well.

neoaggelos (approved):
LGTM, thanks @odufourc! Added a minor comment.

microk8s-resources/actions/enable.host-access.sh
odufourc: About the /8, I believe it is a behavior inherited from ifconfig as an IP from between 0.X.X.X and 127.X.X.X was considered an A class subnet coming in /8 subnet sizes and without any details ifconfig would configure it like this.
If for example the user decides to specify an IP like 192.168.243.1, ifconfig would then suppose it is a class C address and then add the IP address 192.168.243.1/24 but not a /8 so the behavior is not even consistent as it is today.

On the opposite, **ip addr** would create by default a /32 even if nothing is precised after the IP address.
I only added the /32 to be consistent to the command when disabling the host-access and thus removing the IP on localhost interface and to make sure the behavior stays exactly the same even if **ip addr** changes its default subnet size in the future.

When deleting an ip without the subnet indicated, it would works but you would be greated with this message : 
> $ sudo ip addr del 10.0.1.1 dev lo label lo:microk8s
> Warning: Executing wildcard deletion to stay compatible with old scripts.
>          Explicitly specify the prefix length (10.0.1.1/32) to avoid this warning.
>          This special behaviour is likely to disappear in further releases,
>          fix your scripts!

As I don't have perfect knowledge of microk8s' networking, if required the subnet can be expended to a less specific CIDR, however it shouldn't be a /8 as this won't work with the remaining rfc1918 subnets, the behavior of Class A/B/C is something that is deprecated since a while and shouldn't need to be reproduced.

PR #2753, opened 2021-11-25
Improve handling for missing k8s-dqlite arguments file

neoaggelos:
### Summary

When the `k8s-dqlite` arguments file is missing, then simply `exit 0` from the `snap.microk8s.daemon.k8s-dqlite` service.

This is to avoid a race condition where during a refresh, the k8s-dqlite daemon restarts too many times before the configure hook is able to properly cp the required configuration files.

### Notes

- This is similar to what we do for traefik on the worker nodes.


ktsakalozos (approved):
LGTM +1

PR #2756, opened 2021-11-26
Remove the --insecure-port argument

ktsakalozos:
This argument will either way be removed on v1.24 and has currently no effect.


neoaggelos (approved):
LGTM.

Issue #2750, opened 2021-11-23
Dns Add-On not working correctly and 


RafaelMoreira1180778:
I have a non-ha cluster with 3 nodes configured:

```
$ kubectl get nodes -o wide
NAME            STATUS   ROLES    AGE    VERSION                    INTERNAL-IP   EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION    CONTAINER-RUNTIME
ip-10-0-0-103   Ready    <none>   4d1h   v1.22.3-3+64e63223b19811   10.0.0.103    <none>        Ubuntu 20.04.3 LTS   5.11.0-1021-aws   containerd://1.5.2
ip-10-0-0-115   Ready    <none>   24h    v1.22.3-3+64e63223b19811   10.0.0.115    <none>        Ubuntu 20.04.2 LTS   5.4.0-1037-aws    containerd://1.5.2
ip-10-0-0-163   Ready    <none>   24h    v1.22.3-3+64e63223b19811   10.0.0.163    <none>        Ubuntu 20.04.2 LTS   5.4.0-1037-aws    containerd://1.5.2
```

After enabling the `microk8s enable dns``` this is the output:

```
$ microk8s enable dns
Enabling DNS
Applying manifest
serviceaccount/coredns created
configmap/coredns created
Warning: spec.template.metadata.annotations[scheduler.alpha.kubernetes.io/critical-pod]: non-functional in v1.16+; use the "priorityClassName" field instead
deployment.apps/coredns created
service/kube-dns created
clusterrole.rbac.authorization.k8s.io/coredns created
clusterrolebinding.rbac.authorization.k8s.io/coredns created
Restarting kubelet
Adding argument --cluster-domain to nodes.
Applying to node ip-10-0-0-242.
Node ip-10-0-0-242 not present
Applying to node ip-10-0-0-222.
Node ip-10-0-0-222 not present
Applying to node ip-10-0-0-244.
Node ip-10-0-0-244 not present
Applying to node ip-10-0-0-119.
Node ip-10-0-0-119 not present
Applying to node ip-10-0-0-107.
Node ip-10-0-0-107 not present
Applying to node ip-10-0-0-159.
Node ip-10-0-0-159 not present
Applying to node ip-10-0-0-59.
Node ip-10-0-0-59 not present
Applying to node ip-10-0-0-219.
Node ip-10-0-0-219 not present
Applying to node ip-10-0-0-206.
Node ip-10-0-0-206 not present
[...]

PR #2759, opened 2021-11-26
Fix removal of callback tokens

ktsakalozos:
Parsing the callback tokens file was not correct and we were not removing callback tokens of nodes that had departed the cluster. This was casing the distributed ops to iterate over nodes that had already been removed from the cluster as described in  https://github.com/ubuntu/microk8s/issues/2750

### DIFF
--- a/scripts/cluster/reset.py
+++ b/scripts/cluster/reset.py
@@ -183,7 +183,7 @@ def remove_callback_token(node):
         with open(callback_tokens_file, "r+") as callback_fp:
             # Entries are of the format: 'node_hostname:agent_port token'
             # We need to get the node_hostname part
-            for _, line in enumerate(callback_fp):
+            for line in callback_fp:
                 parts = line.split(":")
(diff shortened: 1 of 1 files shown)


scripts/cluster/reset.py
```diff
@@ -181,8 +181,10 @@ def remove_callback_token(node):
     with open(tmp_file, "w") as backup_fp:
         os.chmod(tmp_file, 0o600)
         with open(callback_tokens_file, "r+") as callback_fp:
+            # Entries are of the format: 'node_hostname:agent_port token'
+            # We need to get the node_hostname part
             for _, line in enumerate(callback_fp):
```
neoaggelos, line 186: Unrelated to the PR, but since we are here anyway:

```suggestion
            for line in callback_fp:
```

neoaggelos (approved):
LGTM

neoaggelos (approved):

PR #2768, opened 2021-11-30
Build traefik from source for s390x

ktsakalozos:


neoaggelos (approved):
LGTM to the extend I was able to test it (amd64 build does not break, s390x traefik binary runs)

Issue #1721, opened 2020-11-07
microk8s inspect incorrectly claims that docker is installed


gerrywastaken:
Running `microk8s inspect` was telling me that docker was installed. I fully removed including running `apt purge` it and still got the following message:

```
WARNING:  Docker is installed. 
File "/etc/docker/daemon.json" does not exist. 
You should create it and add the following lines: 
{
    "insecure-registries" : ["localhost:32000"] 
}
and then restart docker with: sudo systemctl restart docker
```

It was only after looking at the code that I figured out what was actually happening. It turns out that that just having `/etc/docker` is all that is needed to get this warning.

https://github.com/ubuntu/microk8s/blob/5e23ac6a001bf3fee2cbb3cd4f1781c6bc945e7e/scripts/inspect.sh#L161

wayneconnolly:
On a previous server I had docker installed, then I uninstalled it. It left the directory `/etc/docker` behind. When I installed microk8s it complained of docker being installed. I deleted the directory and all was good after that. I think a better check would be `/bin/docker`.

gerrywastaken:
Yeah that was pretty much the same situation for me.

gerrywastaken:
The issue still exists. What's the bet this bot isn't able to reopen issues.

gerrywastaken:
Thank you @neoaggelos :heart:

PR #2769, opened 2021-11-30
Print warning for docker if docker daemon is installed

neoaggelos:
### Summary

Closes https://github.com/ubuntu/microk8s/issues/1721

### Changes

- Check that `dockerd` binary is installed, otherwise do not print the warning message.

### DIFF
--- a/scripts/inspect.sh
+++ b/scripts/inspect.sh
@@ -184,7 +184,7 @@ function suggest_fixes {
 
   # check for docker
   # if docker is installed
-  if [ -d "/etc/docker/" ]; then 
+  if [ -d "/etc/docker/" ] && ! [ -z "$(which dockerd)" ]; then
     # if docker/daemon.json file doesn't exist print prompt to create it and mark the registry as insecure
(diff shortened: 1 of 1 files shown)


scripts/inspect.sh
```diff
@@ -184,7 +184,7 @@ function suggest_fixes {
 
   # check for docker
   # if docker is installed
-  if [ -d "/etc/docker/" ]; then 
+  if [ -d "/etc/docker/" && ! -z "$(which dockerd)" ]; then
```
ktsakalozos, line 187: This is giving me `/snap/microk8s/x6/inspect.sh: line 187: [: missing ]`

Maybe we need to split the tests in two, see for example:
```
>  if [ -d "/etc/docker/" ] && ! [ -z "$(which ls)" ]; then echo "docker is here"; fi 
docker is here
>  if [ -d "/etc/docker/" && ! -z "$(which ls)" ]; then echo "docker is here"; fi 
bash: [: missing `]'
```

ktsakalozos (changes requested):

ktsakalozos (approved):

PR #2776, opened 2021-12-02
Initial pass at upgrading our hostPath provider

joedborg:
First pass on this.

Tested with strict, juju and the LMA stack.


tests/validators.py
```diff
@@ -69,7 +69,9 @@ def validate_storage():
     """
     Validate storage by creating a PVC.
     """
-    wait_for_pod_state("", "kube-system", "running", label="k8s-app=hostpath-provisioner")
+    wait_for_pod_state(
+        "", "kube-system", "running", label="app.kubernetes.io/name=hostpath-provisioner"
```
ktsakalozos, line 73: This is causing the test-upgrade to fail. The test upgrade does the following:

  1. installs the old snap
  2. enables some addons from the old snap including the storage
  3. runs the validations for the enabled addons
  4. refreshes to the new snap
  5. runs the validations again to make sure nothing broke

In the old storage addon that is enabled during the test-upgrade the label is `label="k8s-app=hostpath-provisioner"` so the new validator will have to check the old label as well as the new one.

microk8s-resources/actions/enable.storage.sh
```diff
@@ -7,9 +7,16 @@ source $SNAP/actions/common/utils.sh
...
 
-declare -A map
-map[\$SNAP_COMMON]="$SNAP_COMMON"
-use_manifest storage apply "$(declare -p map)"
+"$SNAP/microk8s-enable.wrapper" dns
+"$SNAP/microk8s-enable.wrapper" helm3
+
+"$SNAP/microk8s-helm3.wrapper" repo add rimusz https://charts.rimusz.net
+"$SNAP/microk8s-helm3.wrapper" repo update
+"$SNAP/microk8s-helm3.wrapper" install hostpath-provisioner rimusz/hostpath-provisioner \
```
ktsakalozos, line 15: If we decide this is he way to go we should move/fork the storage repo somewhere we own.

neoaggelos: 1. We should probably fork the repository, and consider what to do with https://github.com/juju-solutions/hostpath-provisioner

2. I don't feel comfortable with the idea of requiring helm in order to enable the hostpath provisioner. This stems from the fact that we lazily pull in the helm3 binary at runtime, which complicates things in environments where external connectivity may be restricted. I would prefer rendering the templates with `helm3 template $NAME $CHARM`, and then maintain the templates instead. The storage provider is something that will barely ever change, no need to introduce an upstream dependency here.

3. The default helm installation uses `/mnt/hostpath`, which is wrong.

4. We probably also need to preserve the provisioner and the storage class name. This provisioner uses `hostpath`, whereas downstream have started depending on `microk8s.io/hostpath` (e.g. Juju). This point probably ties in with 1, looking at the code of our current hostpath-provisioner and the one from the helm chart, it is probably in our best interest to simply update the dependencies of https://github.com/juju-solutions/hostpath-provisioner

neoaggelos (changes requested):

ktsakalozos:
I have second thoughts about this PR. Apart from the from helm which (@neoaggelos correctly points out) we do not want, it is not clear to me what features this implementation brings compared to the the hostpath implementation we already have.

ktsakalozos:
As we discussed, we will not be moving forward with this PR.

Issue #1438, opened 2020-07-28
Need docs linked from inspect to fix nf_conntrack sizing by lxc guest for daemon-proxy


kriswuollett:
Seems same issue as in the closed #119, daemon-proxy in microk8s is attempting to update my system's `/sys/module/nf_conntrack/parameters/hashsize`. Based on some research it sounds like it needs to update it since my Ubuntu 20.04 machine has a lot of CPU cores. There should be a mechanism like in microk8s.inspect to detect this type of error that causes `daemon-proxy` to fail to start, and link to some documentation on how to fix (per OS as needed).

I think there was another issue that a command line flag used to be supported that would cause `daemon-proxy` to skip trying to update these settings, but is now ignored? Is there anyway I can configure microk8s instead of updating my system settings?

[inspection-report-20200728_004132.tar.gz](https://github.com/ubuntu/microk8s/files/4985511/inspection-report-20200728_004132.tar.gz)

My current workaround is:

```shell
sudo /sbin/sysctl -w net.nf_conntrack_max=1572864
sudo /sbin/sysctl -w net.netfilter.nf_conntrack_max=1572864
echo 393216 | sudo tee -a /sys/module/nf_conntrack/parameters/hashsize
```

I assume after reboot updates to these files will cause the fix to be persisted:

```
# /etc/modules
nf_conntrack

# /etc/sysctl.conf
net.netfilter.nf_conntrack_max=1572864
net.nf_conntrack_max=1572864

# /etc/modprobe.d/ip_conntrack_hashsize.conf
options ip_conntrack hashsize=393216
```

ktsakalozos:
Thank you for reporting this @kriswuollett and offering a workaround.

Indeed updating the inspection script to detect this case would be the a first step. 

Have you tried adding `--conntrack-max-per-core=0` in `/var/snap/microk8s/current/args/kube-proxy`?

Another approach would be to see why the lxc profile in `https://github.com/ubuntu/microk8s/blob/master/tests/lxc/microk8s.profile` did not allow access to `/sys/module/nf_conntrack/parameters/hashsize`.

kriswuollett:
I reduced `hashsize` on the host and I can confirm `--conntrack-max-per-core=0` in the config file works. Thanks for pointing that out to me.

[...]

PR #2777, opened 2021-12-03
Allow kube-proxy to change nf_conntrack_max on lxc

ktsakalozos:
When kube-proxy starts it tries to set the contents of `/proc/sys/net/netfilter/nf_conntrack_max` to `131072` if it is set to another number. On lxc containers this file is read only causing kube-proxy to fail. With this PR we update the LXC profiles to allow changing this file. This will address the issue on the pre-1.23 tracks

We also try to detect that we run on lxc and we set the conntrack-max-per-core argument so kube-proxy will not bother with contrack. This behavior will apply for the 1.23+ releases.

Fixes: https://github.com/ubuntu/microk8s/issues/1438

### DIFF
--- a/microk8s-resources/wrappers/run-kubelite-with-args
+++ b/microk8s-resources/wrappers/run-kubelite-with-args
@@ -174,6 +174,14 @@ then
   fi
 fi
 
+# on lxc containers do not try to change the conntrack configuration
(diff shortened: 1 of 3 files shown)


neoaggelos (approved):
LGTM, tested both clean installs as well as refreshing from older versions

PR #2828, opened 2022-01-10
Add low-memory guard

neoaggelos:
### Summary

Prevent MicroK8s from starting control plane components when installed on a machine with low RAM (< 512MB).

Some things below need to be ironed out, but this PR is mostly ready for testing. The changes appear simple, but it is important that we get the UX right.

### Changes

The low-memory-guard is implemented using a lock file.

- During the `install` hook, we check the available RAM and create the lock file if needed.
- `microk8s start/stop/status` will refuse to run when the lock file exists, printing a relevant error message.
- `microk8s inspect` checks for the existence of the lock file, as well as the amount of available RAM.

When the guard is enabled, the node can be joined as a worker node in a cluster. The low memroy guard can be disabled with:

```bash
microk8s start --disable-low-memory-guard
```

### Pending

- [x] Reject joining a cluster as a control-plane node when the low memory guard is enabled.

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -761,3 +761,19 @@ then
     exit 1
   fi
 fi
+
+exit_if_low_memory_guard() {
+  if [ -e ${SNAP_DATA}/var/lock/low-memory-guard.lock ]
+  then
+    echo ''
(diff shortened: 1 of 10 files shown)


neoaggelos:
For testing:

```bash
sudo snap install microk8s --classic --channel latest/edge/low-memory
```

Issue #2829, opened 2022-01-10
Stop relying setting RemoveSelfLink=false feature-gate - we're locking it to true with GA graduation in 1.24


wojtek-t:
Upstream Kubernetes is going to get rid of ability to set SelfLink in release 1.24 (basically graduate graduate RemoveSelfLink feature gate to stable - see the KEP https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1164-remove-selflink, for a bit more details).

It seems that microk8s is still relying on setting it to `false` in two places:
https://github.com/ubuntu/microk8s/blob/master/snap/hooks/configure#L424
https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/default-args/kube-apiserver#L16

This is one more heads'up to stop relying on it.

PR #2833, opened 2022-01-11
Deprecate storage addon

joedborg:
Here we're starting the process of deprecating the `storage` addon and replace it with the `hostpath-storage` addon.

We're also acknowledging the limitations of hostpath storage.

### DIFF
--- a/microk8s-resources/actions/disable.storage.sh
+++ b/microk8s-resources/actions/disable.storage.sh
@@ -4,4 +4,7 @@ set -e
 
 source $SNAP/actions/common/utils.sh
 
+echo "DEPRECIATION WARNING: 'storage' is deprecated and will soon be removed. Please use 'hostpath-storage' instead."
+echo ""
+
 "$SNAP/microk8s-disable.wrapper" hostpath-storage
--- a/microk8s-resources/actions/disable.storage.sh
+++ b/microk8s-resources/actions/disable.storage.sh
@@ -7,4 +7,4 @@ source $SNAP/actions/common/utils.sh
 echo "DEPRECIATION WARNING: 'storage' is deprecated and will soon be removed. Please use 'hostpath-storage' instead."
 echo ""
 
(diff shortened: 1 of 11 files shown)


balchua (commented):
IIRC We need to modify `portainer` addon as it is enabling the `storage` addon.
Adding @balasu in the loop, in case he recalls.

joedborg:
Thanks @balchua, have updated that here https://github.com/ubuntu/microk8s/pull/2833/files#diff-3330b8013f6cdb25fbf80b56147a0f8246f1a993bb27b3d7e38bb9ee4819a67dR13-R14

balchua:
Oh the registry also uses the `storage` addon.

microk8s-resources/actions/disable.storage.sh
```diff
@@ -4,20 +4,7 @@ set -e
 
 source $SNAP/actions/common/utils.sh
 
-echo "Disabling default storage"
-read -ra ARGUMENTS <<< "$1"
```
ktsakalozos, line 8: What happened to the argument here? It is passed to the new addon?

ktsakalozos:
Should we also update the addons list https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/addon-lists.yaml ?

joedborg:
> Should we also update the addons list https://github.com/ubuntu/microk8s/blob/master/microk8s-resources/wrappers/addon-lists.yaml ?

Yes!

neoaggelos (approved):
LGTM

balchua (commented):
Should we consider this issue? I.e. `RemoveSelfLink=false` being deprecated.

https://github.com/ubuntu/microk8s/issues/2829

ktsakalozos:
> Should we consider this issue? I.e. `RemoveSelfLink=false` being deprecated.

Yes, this will be handled by a separate PR.

PR #2838, opened 2022-01-13
Reject joins when the hostname cannot be resolved

neoaggelos:
### Summary

Replaces #2835. 

When the hostname of the joining node does not resolve, outright refuse to join in the cluster. This is because DNS resolution is required for most kubectl commands, which fail otherwise.


ktsakalozos (approved):
LGTM +1 thank you @neoaggelos

PR #2813, opened 2021-12-24
fix ingress leader election when rbac is enabled

balchua:
This PR addresses a breaking change in the way ingress access the configmap for its leader election.
It was previously called `<election-id>-<ingress-class>` where `election-id` is named `ingress-controller-leader` and `ingress-class` is `public`.

Now the resourceName is just the `election-id`

*Also verify you have:*
* [x] Read the [contributions](https://github.com/ubuntu/microk8s/blob/master/CONTRIBUTING.md) page.
* [X] Submitted the [CLA form](https://ubuntu.com/legal/contributors/agreement), if you are a first time contributor.

PR #2849, opened 2022-01-18
Update the ingress RBAC for allowing leader election

ktsakalozos:
Back porting the #2813 to 1.21 .

This is needed due to the 1.0.5 update.

### DIFF
--- a/microk8s-resources/actions/ingress.yaml
+++ b/microk8s-resources/actions/ingress.yaml
@@ -95,7 +95,7 @@ rules:
   resources:
   - configmaps
   resourceNames:
-  - ingress-controller-leader-public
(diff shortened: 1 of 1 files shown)


balchua (commented):
Lgtm

balchua (approved):
Lgtm

neoaggelos (approved):
LGTM, Users already affected would need to disable/enable the ingress addon for this to work, right? Is this acceptable?

ktsakalozos:
> LGTM, Users already affected would need to disable/enable the ingress addon for this to work, right? Is this acceptable?

Unfortunately, this is the case yes. As with the ingress version update the only way to get the update is to re-apply the manifest.

PR #2850, opened 2022-01-18
Feature/jdb/strict merge no ff

joedborg:
Merge latest changes into strict branch.

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -775,3 +775,19 @@ then
     exit 1
   fi
 fi
+
+exit_if_low_memory_guard() {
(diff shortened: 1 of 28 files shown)


joedborg:
@ktsakalozos @neoaggelos I think the memory gating might be breaking this.  It seems fine if I manually do `snap start microk8s` after install and connecting interfaces so perhaps the memory gating is breaking https://github.com/ubuntu/microk8s/blob/feature/jdb/strict/microk8s-resources/actions/common/utils.sh#L671 and its references in https://github.com/ubuntu/microk8s/blob/feature/jdb/strict/snap/hooks/configure#L12 and https://github.com/ubuntu/microk8s/blob/feature/jdb/strict/snap/hooks/configure#L630

Any chance you could have a quick look?  Thanks.

PR #2851, opened 2022-01-19
Do not start traefik and k8s-dqlite until interfaces are connected

ktsakalozos:
These are two new services that did not have the disable flag.


joedborg:
Thanks!

PR #2853, opened 2022-01-19
Upgrade hostpath-provisioner

neoaggelos:
### Summary

Update hostpath-provisioner version to 1.1.0.

### Changes

- Update hostpath-provisioner image to `cdkbot/hostpath-provisioner:1.1.0`
- New multi-arch image supports s390x, so enable storage addon for s390x architectures as well.
- After this PR is merged, we can begin with removing the `RemoveSelfLink=false` feature flag, which is removed on Kubernetes 1.24


joedborg (approved):
LGTM

PR #2870, opened 2022-01-21
Drop RemoveSelfLink=false feature flag

neoaggelos:
### Summary

The `RemoveSelfLink=false` feature flag is being removed in Kubernetes 1.24. We had kept that since it was required for our hostpath-provisioner to work.

This is no longer needed, so we can remove it, in time for 1.24.

The PR also adds a check for `microk8s inspect`, with steps for upgrading users that may encounter issues.


ktsakalozos (approved):
LGTM +1 thank you @neoaggelos

PR #2873, opened 2022-01-24
Silence the hostpath-provisioner error if the provisioner is not installed

ktsakalozos:
The error the inspection script prints when there are is not hostpath provisioner installed:
```
Error from server (NotFound): deployments.apps "hostpath-provisioner" not found
```


neoaggelos (approved):
good catch, thanks, LGTM

PR #2874, opened 2022-01-24
Enable hostpath-storage for s390x architectures

neoaggelos:
### Summary

Add `s390x` to the list of supported architectures for the hostpath-storage addon.


ktsakalozos (approved):
LGTM +1

PR #2892, opened 2022-01-28
Initial split of addon out of the main repository

ktsakalozos:
This PR needs to be merged after https://github.com/canonical/microk8s-addons/pull/1

Changes:
 - Removing the test-addons and test-upgrade from this repo.
 - The test execution scripts (test-distro.sh & github actions) fetch the addon when they run.
 - Updated snapcraft.yaml to fetch the addons on build time.
 - The enable/disable/status commands use the new path of the addons.


build-scripts/set-env-variables.sh
```diff
@@ -49,6 +49,10 @@ export K8S_DQLITE_TAG="${K8S_DQLITE_TAG:-v1.0.3}"
 
 export KUBE_SNAP_ROOT="$(readlink -f .)"
 
+export ADDONS_REPO="${ADDONS_REPO:-https://github.com/canonical/microk8s-addons}"
+export ADDONS_REPO_BRANCH="${ADDONS_REPO_BRANCH:-feature/initial-experimentation}"
```
ktsakalozos, line 53: This has to change right after the https://github.com/canonical/microk8s-addons/pull/1 is merged so we get the addons from the main branch.

PR #2894, opened 2022-01-31
bump calico to v3.21.4

MBcom:
Bump Calico version to v3.21.4  
  
This PR should maybe merged after https://github.com/ubuntu/microk8s/pull/2878 .  
We tested it in a linux-windows environment and we saw the error described in https://github.com/projectcalico/calico/issues/5539.  
We also had to run `kubectl set env daemonset/calico-node -n kube-system IP_AUTODETECTION_METHOD=can-reach=<one-of-our-node-IPs>` because nodes with multiple network connections  selected (automatically) the wrong IP/ network for Calico.  
This environment variable was set prior too therefore we aren't sure whether microk8s had set it or we did earlier too.

### DIFF
--- a/upgrade-scripts/000-switch-to-calico/resources/calico.yaml
+++ b/upgrade-scripts/000-switch-to-calico/resources/calico.yaml
@@ -69,97 +69,134 @@ spec:
     singular: bgpconfiguration
   scope: Cluster
   versions:
-    - name: v1
-      schema:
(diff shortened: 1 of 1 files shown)


ktsakalozos:
Thank you for this PR @MBcom the upgrade of calico is handled in https://github.com/ubuntu/microk8s/pull/2878

PR #2901, opened 2022-02-04
Modify ownership of tmp folder

### LINKED ISSUE #1009: Failed to enable helm
When running `microk8s.enable helm` on edge channel I get the following output:
```
Enabling Helm
Fetching helm version v2.16.0.
/snap/microk8s/1245/actions/enable.helm.sh: line 17: cd: /var/snap/microk8s/1245/tmp/helm: Permission denied
Failed to enable helm
```

zmraul:
Fixes #1009

This makes sure that the addons that use/create _tmp_ folder update the folder so it's on the `microk8s` group.


neoaggelos (changes requested):
Upon consideration, it is probably to ensure the directory exists and has correct permissions in the `configure` hook. See

https://github.com/ubuntu/microk8s/blob/1.23/snap/hooks/configure#L323-L328

and 

https://github.com/ubuntu/microk8s/blob/1.23/snap/hooks/configure#L338

Also, you should make sure that the directory exists (so add an `mkdir -p "${SNAP_DATA}/tmp}"`)

zmraul:
Thanks for the feedback! I updated the PR with the appropriate place to fix this bug.

neoaggelos (approved):
LGTM, thanks @zmraul

ktsakalozos:
Thank you @zmraul

PR #2917, opened 2022-02-14
deprecated --address flag removed in version 1.24

neoaggelos:
### Summary

`kube-scheduler` and `kube-controller-manager` have deprecated the `--address` flag and it is being removed in 1.24

Related logs:
```
Feb 07 11:37:32 ip-172-31-33-169 microk8s.daemon-kubelite[2706]: I0207 11:37:32.038243    2706 daemon.go:22] Starting Controller Manager
Feb 07 11:37:32 ip-172-31-33-169 microk8s.daemon-kubelite[2706]: Flag --address has been deprecated, This flag has no effect now and will be removed in v1.24.
Feb 07 11:37:32 ip-172-31-33-169 microk8s.daemon-kubelite[2706]: I0207 11:37:32.060648    2706 daemon.go:33] Starting Scheduler
Feb 07 11:37:32 ip-172-31-33-169 microk8s.daemon-kubelite[2706]: Flag --address has been deprecated, This flag has no effect now and will be removed in v1.24. You can use --bind-address instead.
```

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -431,6 +431,19 @@ then
   need_api_restart=true
 fi
 
+# scheduler --address flag is removed after 1.24
+if grep -e "--address=" ${SNAP_DATA}/args/kube-scheduler
+then
+  "${SNAP}/bin/sed" -i '/--address=/d' "$SNAP_DATA/args/kube-scheduler"
+fi
+
+# controller-manager --address flag is removed after 1.24
+if grep -e "--address=" ${SNAP_DATA}/args/kube-controller-manager
+then
+  "${SNAP}/bin/sed" -i '/--address=/d' "$SNAP_DATA/args/kube-controller-manager"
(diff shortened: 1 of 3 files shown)


snap/hooks/configure
```diff
@@ -431,6 +431,19 @@ then
   need_api_restart=true
 fi
 
+# scheduler --address flag is removed after 1.24
+if grep -e "--address=" ${SNAP_DATA}/args/kube-scheduler
+then
+  "${SNAP}/bin/sed" -i '/--address=/d' "$SNAP_DATA/args/kube-scheduler"
```
ktsakalozos, line 437: Minor change. I think we need a `need_api_restart=true` after this line too.

neoaggelos: this flag is already deprecated and has no effect, so I think we can avoid restarting the API server.

Issue #2921, opened 2022-02-14
Addons break out breaks snapcraft resuse


joedborg:
Since the addons break out, you cannot re-run `snapcraft build` without cleaning the build and starting it from scratch.  This really, really slows down the pace of local development.

```
+ mkdir addons
mkdir: cannot create directory ‘addons’: File exists
Failed to run 'override-build': Exit code was 1.
Run the same command again with --debug to shell into the environment if you wish to introspect this failure.
```

PR #2922, opened 2022-02-14
Delete addons dir if already created on previous build

joedborg:
Fixes https://github.com/ubuntu/microk8s/issues/2921


ktsakalozos (approved):
LGTM +1 thank you @joedborg

Issue #2920, opened 2022-02-14
Reboot does not retain host-access address


tookeyc:
Install fresh or upgrading to Microk8s 1.23 the host-access address added to the loopback interface is dropped upon server reboot.

Before reboot
![before](https://user-images.githubusercontent.com/59447548/153905660-6317d31a-c6f9-4413-9076-743f654e6893.jpg)

After reboot
![after](https://user-images.githubusercontent.com/59447548/153905722-cbc934db-7a13-4428-aef0-9715aef3718e.jpg)

Attempts to disable/enable host-access addon fails. Adding my own address to the loopback interface works. But this is causing HA issues.

I have attached the report as reqested.

[inspection-report-20220214_202308.tar.gz](https://github.com/ubuntu/microk8s/files/8062374/inspection-report-20220214_202308.tar.gz)

balchua:
Currently the command used to create the local host ip is ephemeral.  It maybe by design.  @ktsakalozos or @neoaggelos thoughts?
Thanks!

tookeyc:
The observed behaviour of host-access for microk8s 1.21 and 1.22 was the configured address was preserved on a reboot. And did not see anything in the release notes of 1.23 to suggest this behaviour had changed.

I am also seeing that when trying to disable the host-access addon that it fails to release even when using the force option.

At the moment I manually add the route after a reboot of the system, but this is far from a robust HA solution.

Any help on this would be very much appreciated.

balchua:
I see this in the logs.
```
Feb 14 19:23:55 worker1.dev-it.baxter.com microk8s.daemon-kubelite[3284]: + /snap/microk8s/2948/sbin/ip addr add '' dev lo label lo:microk8s
Feb 14 19:23:55 worker1.dev-it.baxter.com microk8s.daemon-kubelite[3284]: Error: any valid address is expected rather than "dev".
Feb 14 19:23:55 worker1.dev-it.baxter.com microk8s.daemon-kubelite[3284]: + echo 'Failed to enable host-access'
Feb 14 19:23:55 worker1.dev-it.baxter.com microk8s.daemon-kubelite[3284]: Failed to enable host-access
Feb 14 19:23:55 worker1.dev-it.baxter.com microk8s.daemon-kubelite[3284
```
This line is giving the problem

https://github.com/ubuntu/microk8s/blob/f4e78074e397bbf084ddaaf70a30ad255255421d/microk8s-resources/wrappers/run-kubelite-with-args#L173

[...]

PR #2924, opened 2022-02-15
Fix host-access IP on kubelite restart

neoaggelos:
### Summary

Closes https://github.com/ubuntu/microk8s/issues/2920

See https://github.com/ubuntu/microk8s/issues/2920#issuecomment-1040272348


commodis:
Great work guys!

When will this fix be released? Is it targeted for 1.24?

ktsakalozos:
The fix will be released with 1.24 and has been backported to 1.23. It will be available on 1.23 with the next patch release.

PR #2964, opened 2022-03-04
Go cluster agent

joedborg:
Merging @neoaggelos's work on the re-write of the cluster-agent in Go.  This removes around 40MB of resting consumption of RAM by removing Gunicorn.  It's likely to also increase the number of possible concurrent request handled by the cluster-agent.

I have tested this manually by creating a cluster.  But this includes a suite of new tests aimed at keeping the quality of our .go code high.

We may want to remove more files from `./scripts/cluster`.


joedborg:
Have also manually tested this with creating the cluster and adding Python API server nodes, as well as the inverse.  All working as expected.

joedborg:
Ready for review

cluster-agent/pkg/api/v2/join.go
```diff
@@ -0,0 +1,199 @@
...
+
+	if req.WorkerOnly {
+		if err := util.AddCertificateRequestToken(fmt.Sprintf("%s-kubelet", req.ClusterToken)); err != nil {
+			return nil, fmt.Errorf("failed adding certificate request token for kubelet: %w", err)
+		}
+		if err := util.AddCertificateRequestToken(fmt.Sprintf("%s-proxy", req.ClusterToken)); err != nil {
+			return nil, fmt.Errorf("failed adding certificate request token for kube-proxy: %w", err)
+		}
+
+		// TODO: list of control plane nodes
```
ktsakalozos, line 165: Let's not forget about this.

cluster-agent/pkg/api/v2/join.go
```diff
@@ -0,0 +1,199 @@
...
+
+	if err := util.CreateNoCertsReissueLock(); err != nil {
+		return nil, fmt.Errorf("failed to create lock file to disable certificate reissuing: %w", err)
+	}
+	response := &JoinResponse{
+		CertificateAuthority:       ca,
+		CallbackToken:              callbackToken,
+		APIServerPort:              util.GetServiceArgument("kube-apiserver", "--secure-port"),
+		APIServerAuthorizationMode: util.GetServiceArgument("kube-apiserver", "--authorization-mode"),
+		HostNameOverride:           hostname,
```
ktsakalozos, line 153: In the python implementation the HostNameOverride was returning the IP. Using the hostname is failing for me because it makes it to the dqlite config.

ktsakalozos, line 153: In the python implementation the HostNameOverride was returning the IP. Using the hostname is failing for me because it makes it to the dqlite config.

cluster-agent/pkg/api/v1/configure.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// Configure implements "POST /CLUSTER_API_V1/configure".
+func Configure(ctx context.Context, req ConfigureRequest) error {
+	if !util.IsValidSelfCallbackToken(req.CallbackToken) {
+		return fmt.Errorf("invalid callback token")
+	}
+	for _, service := range req.ConfigureServices {
+		if err := util.UpdateServiceArguments(service.Name, service.UpdateArguments, service.RemoveArguments); err != nil {
+			return fmt.Errorf("failed to update arguments of service %q: %w", service.Name, err)
```
ktsakalozos, line 62: Lets try to keep the same error messages. The error messages are printed on the other side in https://github.com/canonical/microk8s/blob/master/scripts/cluster/join.py#L236

cluster-agent/pkg/server/server.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// POST /v1/join
+	server.HandleFunc(fmt.Sprintf("%s/join", ClusterAPIV1), withMiddleware(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodPost {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		req := v1.JoinRequest{}
+		if err := UnmarshalJSON(r, &req); err != nil {
+			HTTPError(w, http.StatusBadRequest, err)
```
ktsakalozos, line 42: Lets keep the same return codes.

cluster-agent/pkg/server/server.go
```diff
@@ -0,0 +1,143 @@
...
+
+const (
+	// ClusterAPIV1 is the version 1 of the MicroK8s cluster API.
+	ClusterAPIV1 = "/cluster/api/v1.0"
+	// ClusterAPIV2 is the version 2 of the MicroK8s cluster API.
+	ClusterAPIV2 = "/cluster/api/v2.0"
+)
+
+// NewServer creates a new *http.ServeMux and registers the MicroK8s cluster agent API endpoints.
+func NewServer(timeout time.Duration, apiv1 *v1.API, apiv2 *v2.API) *http.ServeMux {
```
lferran, line 21: IMO this function seems to be doing to many things.
It could be re-written in a different way so each endpoint's error response logic is  more testable

PR #2974, opened 2022-03-11
scripts: fix the inspect script not to call /bin/mount

mardy:
There's no need to use /bin/mount, as we can read the info from /proc
ourselves.


ktsakalozos:
Merged in https://github.com/canonical/microk8s/pull/3237

PR #2975, opened 2022-03-11
scripts: remove all 'sudo' invocations from inspect.sh

mardy:
We are anyway recommending to run the inspect script with sudo (and
under strict confinement we cannor execute /bin/sudo anyway...).


ktsakalozos:
Addressed in https://github.com/canonical/microk8s/pull/3239, thank you @mardy

PR #2978, opened 2022-03-11
Fixes for git to work with strict confinement

neoaggelos:
### Summary

git uses a few helper paths and configuration files from the system. Make sure that we do not get failures when running in strict

Also, remove the git clone from the install hook (it is already there in the configure hook).


ktsakalozos (approved):
LGTM +1 thank you @neoaggelos

PR #2982, opened 2022-03-14
Do not load the community repo

ktsakalozos:
With this PR we stop loading the community repository by default. This was we do not clutter the UX with often irrelevant addons. If however we detect that an addon is in the community repository we prompt the user to enable the repository with a `microk8s enable community` call.

### DIFF
--- a/scripts/wrappers/common/utils.py
+++ b/scripts/wrappers/common/utils.py
@@ -369,7 +369,7 @@ def parse_xable_single_arg(addon_arg: str, available_addons: list):
             if is_community_addon(get_current_arch(), addon_name):
                 click.echo("To use the community maintained flavor enable the respective repository:")
                 click.echo("")
-                click.echo("    mcirok8s enable community")
+                click.echo("    microk8s enable community")
                 click.echo("")
 
             sys.exit(1)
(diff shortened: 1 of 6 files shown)


scripts/wrappers/common/utils.py
```diff
@@ -343,6 +366,12 @@ def parse_xable_single_arg(addon_arg: str, available_addons: list):
         matching_repos = [repo for (repo, addon) in available_addons if addon == addon_name]
         if len(matching_repos) == 0:
             click.echo("Addon {} was not found in any repository".format(addon_name), err=True)
+            if is_community_addon(get_current_arch(), addon_name):
+                click.echo("To use the community maintained flavor enable the respective repository:")
+                click.echo("")
+                click.echo("    mcirok8s enable community")
```
neoaggelos, line 372: I presume this is a typo

snap/snapcraft.yaml
```diff
@@ -560,7 +560,7 @@ parts:
         reference="$(echo ${line} | cut -f3 -d',')"
         git clone "${repository}" -b "${reference}" "addons/${name}"
       done
-      echo "${ADDONS_REPOS_ENABLED}" > addons/.auto-add
+      echo "core" > addons/.auto-add
```
neoaggelos, line 563: I'd rather this was changed in [set-env-variables.sh] instead (https://github.com/ubuntu/microk8s/blob/9536527e31d0d4d9c59f529d247c152625e68af5/build-scripts/set-env-variables.sh#L56)

neoaggelos (changes requested):

neoaggelos (approved):

balchua (approved):

PR #3022, opened 2022-04-01
Get the core addons for strict from the main addons branch

ktsakalozos:


neoaggelos (approved):
LGTM, let's merge once CI has a go at it

PR #3029, opened 2022-04-06
Allow for refreshing the server and front-proxy certs

ktsakalozos:
This PR introduces a `-e` or `--cert` argument in the refresh-certs command. With the new argument you can specify which certificate you want to refresh.

Also updated tox to be able to call unit tests on wrappers and non-wrapper scripts.


neoaggelos (changes requested):
Running `microk8s refresh-certs` produces:

```
root@k1:~# m refresh-certs
Taking a backup of the current certificates under /var/snap/microk8s/x1/var/log/ca-backup/
Creating new certificates
Unknown certificate to refresh
```

which does not help me at all as a user. Also, I'm not sure I quite like the backups under `anything/var/log`, why is that?

Further, updating `server.crt` restarts MicroK8s, including containerd, dqlite etc. I do not think we need this. We should probably only restart the required services (in this case, apiserver?)

Of course, we should still restart everything if the CA has changed.

neoaggelos (approved):

PR #3061, opened 2022-04-13
Fix ip counter in render csr template function (#3059)

### LINKED ISSUE #3059: Adding another IP addresses to csr.conf.template is conflicting with local ip address
To access mikrok8s deployd behind NAT by ip address additional address must be added to /var/snap/microk8s/current/certs/csr.conf.template the file looks like that:
```
[ alt_names ]
DNS.1 = kubernetes
DNS.2 = kubernetes.default
DNS.3 = kubernetes.default.svc
DNS.4 = kubernetes.default.svc.cluster
DNS.5 = kubernetes.default.svc.cluster.local
IP.1 = 127.0.0.1
IP.2 = 10.152.183.1
#MOREIPS
IP.3 = 10.245.162.247
```
After updating certs, csr.conf looks like that
```
[ alt_names ]
DNS.1 = kubernetes
DNS.2 = kubernetes.default
DNS.3 = kubernetes.default.svc
DNS.4 = kubernetes.default.svc.cluster
DNS.5 = kubernetes.default.svc.cluster.local
IP.1 = 127.0.0.1
IP.2 = 10.152.183.1
IP.3 = 10.5.3.112
IP.3 = 10.245.162.247
```
As a result 10.5.3.112 is excluded from the server certificate.

[inspection-report-20220412_140801.tar.gz](https://github.com/canonical/microk8s/files/8473668/inspection-report-20220412_140801.tar.gz)

err404r:
Fixes #3059

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -472,7 +472,11 @@ render_csr_conf() {
         local ips='' sep=''
         local -i i=3
         for IP_ADDR in $(echo "$IP_ADDRESSES"); do
-            ips+="${sep}IP.$((i++)) = ${IP_ADDR}"
+            local ip_id="IP.$((i++))"
+	    while grep '^'"${ip_id}" ${SNAP_DATA}/certs/csr.conf.template > /dev/null ; do
+              ip_id="IP.$((i++))"
+            done
+            ips+="${sep}${ip_id} = ${IP_ADDR}"
             sep='\n'
         done
         "$SNAP/bin/sed" -i "s/#MOREIPS/${ips}/g" ${SNAP_DATA}/certs/csr.conf.rendered


microk8s-resources/actions/common/utils.sh
```diff
@@ -470,7 +470,8 @@ render_csr_conf() {
     if ! [ "$IP_ADDRESSES" == "127.0.0.1" ] && ! [ "$IP_ADDRESSES" == "none" ]
     then
         local ips='' sep=''
-        local -i i=3
+        local ip_count=$(grep -ce '^IP' certs/csr.conf.template)
```
neoaggelos, line 473: I don't think this fixes the issue. In particular, the `IP.$number` lines in csr.conf.template need not be numbered, and also need not be in sequential order.

For example, the above does not work for this csr.conf.template:

```
....
IP.1 = $SOME_IP
IP.2 = $SOME_OTHER_IP
#MOREIPS#
IP.4 = $CUSTOM_IP
```

This is because grep would return 3, thus add lines with `IP.4`, overwriting our custom IP.

microk8s-resources/actions/common/utils.sh
```diff
@@ -470,7 +470,8 @@ render_csr_conf() {
...
+        local -i i=$((ip_count+1))
         for IP_ADDR in $(echo "$IP_ADDRESSES"); do
             ips+="${sep}IP.$((i++)) = ${IP_ADDR}"
```
neoaggelos, line 476: Continuing on the previous comment, I think it is better to add `microk8s-` (or similar) prefix to the IP addresses added automatically by MicroK8s, something like this:

```suggestion
            ips+="${sep}IP.microk8s-$((i++)) = ${IP_ADDR}"
```

Note that we also do something similar already for the MicroK8s charm, see https://github.com/canonical/charm-microk8s/blob/master/config.yaml#L98-L99

neoaggelos, line 476: Continuing on the previous comment, I think it is better to add `microk8s-` (or similar) prefix to the IP addresses added automatically by MicroK8s, something like this:

```suggestion
            ips+="${sep}IP.microk8s-$((i++)) = ${IP_ADDR}"
```

Note that we also do something similar already for the MicroK8s charm, see https://github.com/canonical/charm-microk8s/blob/master/config.yaml#L98-L99

neoaggelos (changes requested):
Thank you for this PR @err404r

I think the proposed fix may still cause issues in edge cases, see my comments for the reasoning and an alternative option.

Thanks!

microk8s-resources/actions/common/utils.sh
err404r: IP.mikrok8s-<number> looks like the best option, but according to OpenSSL docs.
Identifier must be numeric. https://www.openssl.org/docs/manmaster/man5/x509v3_config.html
So it is either undocumented feature or charm replace those values while render the config.

Also according to the docs numeric identifiers are not required to be sequential so the order of records is not important.

err404r:
Keeping in mind all thoughts above. I could propose another fix.

microk8s-resources/actions/common/utils.sh
```diff
@@ -472,7 +472,13 @@ render_csr_conf() {
         local ips='' sep=''
         local -i i=3
         for IP_ADDR in $(echo "$IP_ADDRESSES"); do
-            ips+="${sep}IP.$((i++)) = ${IP_ADDR}"
+            local ip_id="IP.$((i++))"
+            while ! [ '0' == "$(grep -c \"^${ip_id}\" ${SNAP_DATA}/certs/csr.conf.template)" ]
+            do
+              ip_id="IP.$((i++))"
+              echo ${ip_id}
```
neoaggelos, line 479: Is this leftover from debugging?

neoaggelos (commented):
I quote like this approach better, thanks @err404r.

Please can you look into one final comment?

microk8s-resources/actions/common/utils.sh
```diff
@@ -472,7 +472,13 @@ render_csr_conf() {
...
+            done
```
neoaggelos, line 480: Unfortunately this does not work for me, and I think it is related to the way the shell escapes the `^IP.$x` regex.

```suggestion
                while grep '^'"${ip_id}" ${SNAP_DATA}/certs/csr.conf.template > /dev/null ; do
                        ip_id="IP.$((i++))"
                done
```

neoaggelos (changes requested):
@err404r i think bash is doing something with the way it evaluates the quotes. Can you check if the suggested fix works for you?

microk8s-resources/actions/common/utils.sh
err404r: Here grep is not set to used regexp otherwise `.` in the IP would be a problem for IP.3 IP.33 will match also. According to grep documentation it supports ^ and $ without switching to regexp matching. But anyway looks like I pushed the wrong commit will double check.
Actually if in you setup grep is replaced with egrep to enable regexp by default this could be the case.

err404r:
Your suggestion is great, sorry not very familiar with github UI. So I just committed your version

microk8s-resources/actions/common/utils.sh
neoaggelos: good catch that `IP.33` would also match, but I do not think this is an issue (we might simply skip a few possible values).

ktsakalozos:
Thank you @err404r

PR #3077, opened 2022-04-19
Add `microk8s version` command

### LINKED ISSUE #3065: Add `microk8s --version` flag
## Goal
The purpose of this ticket is to add the `--version` flag to the main `microk8s` entrypoint.

## Motivation
As of now, we are using `snap info microk8s` and `microk8s.kubectl version` to know which versions of the snap and upstream k8s are installed, correspondingly. However, we are missing a convenient way to check, not only that, but also the versions for all microk8s components (cluster agent, kind, dqlite, etc)

We want to show the version for the following components:
- [x] MicroK8s snap
- [x] Upstream k8s
- [x] cni

## Example output
 ```bash
 $ microk8s --version      
MicroK8s v1.21.11 revision: 3058
  - K8s: 1.21-3
  - CNI: 4.3-beta
 ```

lferran:
### Description

fixes #3065 

The proposed implementation is the following:
- At build time, we write upstream versions (kubernetes and CNI so far) into a var file.
- `version` command reads from the file.
- Snap version and revision is queried from `snap list microk8s`

### DIFF
--- a/microk8s-resources/wrappers/microk8s.wrapper
+++ b/microk8s-resources/wrappers/microk8s.wrapper
@@ -27,6 +27,9 @@ elif [ "${APP}" == "inspect" ]; then
 elif [ "${APP}" == "help" ] || [ "${APP}" == "--help" ] || [ "$APP" == "-h" ]; then
     help
     readonly EXIT="0"
+elif [ "${APP}" == "--version" ]; then
+    "${SNAP}/microk8s-version.wrapper"
+    readonly EXIT="$?"
 else
     echo "'${APP}' is not a valid MicroK8s subcommand."
     help
--- a/scripts/wrappers/version.py
+++ b/scripts/wrappers/version.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python3
+
+import click
+import json
+from common.utils import snap_data
+from common.utils import run
+
+
+VERSIONS_FILE_NAME = "versions.json"
+VERSIONS_FILE_PATH = snap_data() / VERSIONS_FILE_NAME
+
+
+VERSIONS = None
+
+
+def get_upstream_version(upstream: str) -> str:
+    global VERSIONS
+
+    if VERSIONS is None:
+        # Cache versions
+        VERSIONS = _read_versions_file()
+
+    return VERSIONS[upstream]
+
+
+def _read_versions_file():
+    with open(VERSIONS_FILE_PATH, mode="r") as versions_file:
+        versions = json.loads(versions_file.read())
+        return versions
+
+
+def get_snap_version_data():
+    output = run("snap", "list", "microk8s")
+    snap_info = output.split("\n")[1]
+    version, revision = snap_info.split()[1:3]
+    return version, revision
+
+
+def print_versions() -> None:
+    version, revision = get_snap_version_data()
+    print(f"MicroK8s {version} revision: {revision}")
+
+    kube_version = get_upstream_version("kube")
+    print(f"  - K8s: {kube_version}")
+
+    cni_version = get_upstream_version("cni")
+    print(f"  - CNI: {cni_version}")
+
+
+@click.command(
+    context_settings={
+        "help_option_names": ["-h", "--help"],
+    }
+)
+def versions_command():
+    """
+    Shows version information for MicroK8s and its upstream components.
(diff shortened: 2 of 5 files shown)


microk8s-resources/wrappers/microk8s.wrapper
```diff
@@ -27,6 +27,9 @@ elif [ "${APP}" == "inspect" ]; then
 elif [ "${APP}" == "help" ] || [ "${APP}" == "--help" ] || [ "$APP" == "-h" ]; then
     help
     readonly EXIT="0"
+elif [ "${APP}" == "--version" ]; then
+    "${SNAP}/microk8s-version.wrapper"
+    readonly EXIT="$?"
```
neoaggelos, line 32: Is this to have both `microk8s version` and `microk8s --version` work?

scripts/wrappers/version.py
```diff
@@ -0,0 +1,63 @@
...
+    with open(VERSIONS_FILE_PATH, mode="r") as versions_file:
+        versions = json.loads(versions_file.read())
+        return versions
+
+
+def get_snap_version_data():
+    output = run("snap", "list", "microk8s")
+    snap_info = output.split("\n")[1]
+    version, revision = snap_info.split()[1:3]
+    return version, revision
```
neoaggelos, line 36: I'd rather we used `$SNAP_REVISION` for the revision. also, `snap list microk8s` is most likely going to be an issue in strict confinement.

scripts/wrappers/version.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def print_versions() -> None:
+    version, revision = get_snap_version_data()
+    print(f"MicroK8s {version} revision: {revision}")
+
+    kube_version = get_upstream_version("kube")
```
neoaggelos, line 43: Isn't it better to just load the file once, then read from the dict? 
e.g.

```
versions = json.load(version_file)
kube_version = version["kube"]
cni_version = version["cni"]
```

scripts/wrappers/version.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def _read_versions_file():
+    with open(VERSIONS_FILE_PATH, mode="r") as versions_file:
+        versions = json.loads(versions_file.read())
+        return versions
+
+
+def get_snap_version_data():
+    output = run("snap", "list", "microk8s")
```
neoaggelos, line 33: I think this is not going to work well with strict confinement. For the revision number, probably best to use the `SNAP_REVISION` environment variable. Also, the snap version is set to `$KUBE_VERSION` (see https://github.com/ubuntu/microk8s/blob/1e538046bac00df1b5ff5a1bc95ac11011d96480/snap/snapcraft.yaml#L608), which we already have

neoaggelos (changes requested):

scripts/wrappers/version.py
lferran: yup, indeed. much simpler 👍🏼

microk8s-resources/wrappers/microk8s.wrapper
lferran: Yes. I added it when we weren't sure between `--version` and `version`. Will remove it

snap/snapcraft.yaml
```diff
@@ -67,6 +67,8 @@ apps:
   kubectl:
     command: microk8s-kubectl.wrapper
     completer: kubectl.bash
+  version:
```
ktsakalozos, line 70: With this you add a new command. You will be able to call:
```
microk8s version
```
If I understand this correctly we want to add a `--version` argument to the already existing `microk8s` command.

ktsakalozos:
@lferran what would you say if we rewrote the microk8s.wrapper in python with click and add there the `--version` argument?

[1] https://github.com/canonical/microk8s/blob/master/microk8s-resources/wrappers/microk8s.wrapper

lferran:
> @lferran what would you say if we rewrote the microk8s.wrapper in python with click and add there the `--version` argument?
> 
> [1] https://github.com/canonical/microk8s/blob/master/microk8s-resources/wrappers/microk8s.wrapper

sounds good to me! I'll close this and go for rewriting the main entrypoint in python first

PR #3078, opened 2022-04-20
Adding `--version` flag to microk8s

lferran:
## Goal
The purpose of this ticket is to add the `--version` flag to the main `microk8s` entrypoint. It will also work when invoked with `microk8s version`, to be conformant with related tools like `microk8s ctr version`, `docker version` or `kubectl version`.

## Motivation
As of now, we are using `snap info microk8s` and `microk8s.kubectl version` to know which versions of the snap and upstream k8s are installed, correspondingly. We are missing a convenient and unified way to get version information.

We want to show the version for the following components:
- [x] MicroK8s snap
- [x] Upstream k8s
- [x] cni

## Example output
  ```bash
 $ microk8s version      
MicroK8s v1.21.11 revision 3058
 ```


lferran:
It seems that writing the versions.json file is not allowed
```bash
Writing versions at /snap/snapcraft/7201/versions.json ...
/bin/sh: 9: /root/parts/runc/src/set-env-variables.sh: cannot create /snap/snapcraft/7201/versions.json: Read-only file system
+ echo {"kube":"v1.23.5","cni":"v0.8.7"}
```
Alternatively, if there is a way to get the cni version from python, we could drop the file approach.

lferran:
I managed to install the snap locally and test it:
```bash
$ microk8s --version
MicroK8s v1.23.6 revision: x1
  - K8s: v1.23.6
  - CNI: v0.8.7
```
and also
```bash
$ sudo snap run --shell microk8s
root@pardal:~# cat $SNAP/versions.json
{"kube":"v1.23.6","cni":"v0.8.7"}
```

glebsts:
@lferran though you did a great job, thank you, mb there is a point to align this feature with other "get version" commands like `microk8s ctr version`, `docker version`, `kubectl version` etc, and make it `version`, not `--version`?

barrettj12:
> @lferran though you did a great job, thank you, mb there is a point to align this feature with other "get version" commands like `microk8s ctr version`, `docker version`, `kubectl version` etc, and make it `version`, not `--version`?

Why not support both?

microk8s-resources/wrappers/microk8s.wrapper
```diff
@@ -27,6 +27,9 @@ elif [ "${APP}" == "inspect" ]; then
 elif [ "${APP}" == "help" ] || [ "${APP}" == "--help" ] || [ "$APP" == "-h" ]; then
     help
     readonly EXIT="0"
+elif [ "${APP}" == "--version" ] || [ "${APP}" == "-v" ]; then
+    "${SNAP}/microk8s-version"
```
neoaggelos, line 31: If we changed the name of the script to `microk8s-version.wrapper`, we would basically also get the `microk8s version` flavor of the command for free.

scripts/wrappers/version.py
```diff
@@ -0,0 +1,36 @@
...
+
+def get_upstream_versions() -> str:
+    versions_file = Path(environ["SNAP"]) / "versions.json"
+    with open(versions_file, mode="r") as file:
+        versions = json.loads(file.read())
+        return versions
+
+
+def get_snap_version() -> str:
+    return environ["SNAP_VERSION"]
```
neoaggelos, line 16:
```suggestion
    return os.getenv("SNAP_VERSION", "(unknown)")
```

scripts/wrappers/version.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def get_snap_revision() -> str:
+    return environ["SNAP_REVISION"]
```
neoaggelos, line 20: Similarly here

tests/unit/test_version.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@patch("version.environ", {"SNAP_VERSION": "v1.21.11"})
+def test_get_snap_version():
+    assert get_snap_version() == "v1.21.11"
+
+
+@patch("version.environ", {"SNAP_REVISION": "3086"})
+def test_get_snap_revision():
+    assert get_snap_revision() == "3086"
```
neoaggelos, line 19: I would probably avoid mocking for this, and instead just set the environment variable

neoaggelos (commented):
Some minor comments

scripts/wrappers/version.py
lferran: isn't `SNAP_VERSION` a variable that we expect to be there always?

tests/unit/test_version.py
lferran: sure! Why though? Out of curiosity

neoaggelos: This way, the tests are too closely tied to the implementation (e.g. if the implementation were to use `os.getenv` instead, these tests would not work). 

Mocks are best suited for outputs/integration with external services. They also serve as a way of disconnecting the testing environment from the real one. This is useful when we want to mock the output of an external command, etc, but is probably a dis-service when checking environment (where no mocks are really needed, the test case itself is isolated).

lferran: makes total sense! thanks for sharing your thoughts

ktsakalozos (approved):
LGTM +1 thank you @lferran

PR #3107, opened 2022-04-28
Update dqlite to 1.10.0

ktsakalozos:


ktsakalozos:
I do not see memory spikes anymore. 
nospikes

If it is ok with you @neoaggelos lets merge this for 1.24.

neoaggelos (approved):
I also see that the dqlite cluster remains stable, I don't see any regressions in the logs or otherwise.

LGTM, let's add this in for 1.24

PR #3125, opened 2022-05-03
Validate newly added addon repositories

lferran:
### Description

This PR adds some validation checks when adding addon repositories:
 - Check that addons.yaml files is correct (according to [the addons docs](https://microk8s.io/docs/howto-addons))
 - Check that enable and disable hooks are present and have the right permissions.

### DIFF
--- a/scripts/wrappers/addons.py
+++ b/scripts/wrappers/addons.py
@@ -4,14 +4,14 @@ import os
 import shutil
 import subprocess
 import sys
-import jsonschema
-
-import click
-import yaml
 from pathlib import Path
-from common.utils import get_current_arch, snap_common
 from typing import List
 
+import click
+import jsonschema
+import yaml
+
+from common.utils import get_current_arch, snap_common
 
 GIT = os.path.expandvars("$SNAP/git.wrapper")
 
--- a/scripts/wrappers/addons.py
+++ b/scripts/wrappers/addons.py
@@ -218,16 +218,44 @@ def update(name: str):
         click.echo("Error: built-in repository '{}' cannot be updated".format(name), err=True)
         sys.exit(1)
 
+    commit_before_pull = git_current_commit(repo_dir)
+
     click.echo("Updating repository {}".format(name))
     subprocess.check_call([GIT, "pull"], cwd=repo_dir)
 
-    if not (repo_dir / "addons.yaml").exists():
-        click.echo(
-            "Error: repository '{}' does not contain an addons.yaml file".format(name), err=True
-        )
-        click.echo("Remove it with:", err=True)
-        click.echo("    microk8s addons repo remove {}".format(name), err=True)
+    try:
+        validate_addons_repo(repo_dir)
+    except RepoValidationError as err:
+        click.echo(err.message, err=True)
+        click.echo(f"Rolling back repository {name}")
+        git_rollback(commit_before_pull, repo_dir)
         sys.exit(1)
+
+
+class GettingGitCommitError(Exception):
+    def __init__(self, exit_code, stderr):
+        self.exit_code = exit_code
+        self.stderr = stderr
+
+
+def git_current_commit(repository: Path) -> str:
+    """
+    Returns the current commit hash of a given git repository
+    """
+    cmd = [GIT, "rev-parse", "--verify", "HEAD"]
+    process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=repository)
+    stdout, stderr = process.communicate()
+    if process.returncode != 0:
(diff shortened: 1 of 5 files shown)


scripts/wrappers/addons.py
```diff
@@ -4,11 +4,14 @@
 import shutil
 import subprocess
 import sys
+import jsonschema
```
lferran, line 7: I hope it's ok to add a new dependency :/ otherwise, I could try to check the contents of `addons.yaml` "manually"

scripts/wrappers/addons.py
```diff
@@ -18,6 +21,123 @@
...
+                            "properties": {
+                                "name": {"type": "string"},
+                                "description": {"type": "string"},
+                                "version": {"type": "string"},
+                                "check_status": {"type": "string"},
+                                "supported_architectures": {
+                                    "type": "array",
+                                    "items": {
+                                        "type": "string",
+                                        "enum": ["amd64", "arm64", "s390x"],
```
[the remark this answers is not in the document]
neoaggelos: it's ok to start with these, then add more as needed.

scripts/wrappers/addons.py
```diff
@@ -4,11 +4,14 @@
...
 import subprocess
 import sys
+import jsonschema
 
 import click
 import yaml
-
+from pathlib import Path
 from common.utils import get_current_arch, snap_common
+from typing import List
```
neoaggelos, line 13: let's use the following order for python imports

```py
import standard_library

import third_party

import our_own
```

scripts/wrappers/addons.py
```diff
@@ -18,6 +21,123 @@
...
+    """
+    Check that enable and disable hooks are present in the
+    repository, and that they have execute permissions.
+    """
+    for addon in get_addons_list(repo_dir):
+        addon_folder = repo_dir / "addons" / addon
+        for hook_name in ("enable", "disable"):
+            hook = addon_folder / hook_name
+            check_exists(hook)
+            check_is_executable(addon_folder)
```
neoaggelos, line 119: is this a typo?

```suggestion
            check_is_executable(hook)
```

scripts/wrappers/addons.py
```diff
@@ -18,6 +21,123 @@
...
+
+
+def get_addons_list(repo_dir: Path) -> List[str]:
+    contents = load_addons_yaml(repo_dir)
+    return [addon["name"] for addon in contents["microk8s-addons"]["addons"]]
```
neoaggelos, line 124: since we are validating the schema, are we sure that these do not crash at this point?

scripts/wrappers/addons.py
```diff
@@ -41,13 +161,7 @@ def add(name: str, repository: str, reference: str, force: bool):
...
     subprocess.check_call(["chgrp", "microk8s", "-R", repo_dir])
 
-    if not (repo_dir / "addons.yaml").exists():
-        click.echo(
-            "Error: repository '{}' does not contain an addons.yaml file".format(name), err=True
-        )
-        click.echo("Remove it with:", err=True)
-        click.echo("    microk8s addons repo remove {}".format(name), err=True)
-        sys.exit(1)
+    validate_addons_repo(repo_dir)
```
neoaggelos, line 164: if this fails in any way, i think it is best that we cleanup the folder instead of leaving it there. in general, i'd rather we didn't call sys.exit(1) in multiple place, but rather an exception that we can catch here, print a useful error message and then cleanup and exit.

neoaggelos (changes requested):

scripts/wrappers/addons.py
lferran: yes good catch! it's actually a typo. this makes me think I have to revisit the tests

scripts/wrappers/addons.py
lferran: Sounds like a much better idea 👍🏼

lferran:
@neoaggelos ready for re-review. 

Now that I am at it, should I add the same validation logic for the `update` command? 
In the case of an update failing to validate, should we remove the repo, or just roll-back the git pull?

neoaggelos:
> @neoaggelos ready for re-review. 
> 
> Now that I am at it, should I add the same validation logic for the `update` command? 
> In the case of an update failing to validate, should we remove the repo, or just roll-back the git pull?

Yes, reverting the update sounds better. You can probably do this by checking the commit hash before upgrading, then git reset to it if validation fails.

ktsakalozos (approved):
LGTM +1 thank you @lferran

neoaggelos (approved):
LGTM, great work with this @lferran

Issue #3085, opened 2022-04-22
calicp-kube-controllers can't get API Server: context deadline exceeded


atimin:
I have the problem after having installed microk8s on my Ubuntu 21.10 server:

```
sudo snap install microk8s --channel=1.23 --classic
```

Checked the pods and saw that one crushed:
```
$ microk8s.kubectl get pods -n kube-system
calico-node-c7h46                          1/1     Running            1 (7m38s ago)   10m
calico-kube-controllers-5ddf994775-gp8cv   0/1     CrashLoopBackOff   7 (34s ago)     10m
```

In logs , I see that something is wrong with the API Server:

```
$ microk8s.kubectl logs calico-kube-controllers-5ddf994775-gp8cv  -n kube-system
2022-04-22 13:26:15.311 [INFO][1] main.go 88: Loaded configuration from environment config=&config.Config{LogLevel:"info", WorkloadEndpointWorkers:1, ProfileWorkers:1, PolicyWorkers:1, NodeWorkers:1, Kubeconfig:"", DatastoreType:"kubernetes"}
W0422 13:26:15.312587       1 client_config.go:543] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
2022-04-22 13:26:15.313 [INFO][1] main.go 109: Ensuring Calico datastore is initialized
2022-04-22 13:26:25.313 [ERROR][1] client.go 261: Error getting cluster information config ClusterInformation="default" error=Get "https://10.152.183.1:443/apis/crd.projectcalico.org/v1/clusterinformations/default": context deadline exceeded
2022-04-22 13:26:25.313 [FATAL][1] main.go 114: Failed to initialize Calico datastore error=Get "https://10.152.183.1:443/apis/crd.projectcalico.org/v1/clusterinformations/default": context deadline exceeded
```

Result of `microk8s.inspect`

[inspection-report-20220422_132011.tar.gz](https://github.com/canonical/microk8s/files/8541366/inspection-report-20220422_132011.tar.gz)

balchua:
I see these these in the `kubelite` logs
```
Apr 22 13:19:39 drift-test-rig microk8s.daemon-kubelite[843]: I0422 13:19:39.899796     843 proxier.go:657] "Failed to load kernel module with modprobe, you can ignore this message when kube-proxy is running inside container without mounting /lib/modules" moduleName="ip_vs"
[...]

PR #3137, opened 2022-05-09
improve handling for br_netfilter module

neoaggelos:
### Summary

References #3085

- try to load using modprobe from the host, fallback to the snap binary if that fails
- if br_netfilter module loads successfully, make sure to remove the --proxy-mode=userspace flag from kube-proxy

### Notes

I can consistently reproduce the issue with CentOS 7. After merging, we need to backport this to 1.21-1.23 branches as well with priority.


ktsakalozos (approved):
LGTM +1 thank you @neoaggelos

PR #3140, opened 2022-05-10
[Backport 1.22] Bump the ingress controller version from 1.0.5 -> 1.2.0

lferran:


lferran:
tested it locally and it works.

PR #3164, opened 2022-05-25
Try to get the dashboard token in dashboard-proxy before creating a n…

ktsakalozos:


scripts/wrappers/dashboard_proxy.py
```diff
@@ -45,27 +78,13 @@ def dashboard_proxy():
...
-            break
-
-        time.sleep(5)
+    # Get token created by the dashboard enable script
+    token = get_token("microk8s-dashboard-token")
+    if not token:
+        # Create a token. This is needed in case the enable script is pre-1.25.
+        print("Create token for accessing the dashboard")
+        run([KUBECTL, "apply", "-f", "-"], input=SECRET_YAML.encode("ascii"))
+        token = get_token("microk8s-dashboard-proxy-token")
```
neoaggelos, line 87: Since we are doing this either way, we don't have to use a separate name for the secret. Should we rename `microk8s-dashboard-proxy-token` to `microk8s-dashboard-token`?

ktsakalozos: +1

PR #3172, opened 2022-05-29
Check the correct file for AppArmor confinement.

MFAshby:
Documentation:
https://www.kernel.org/doc/html/latest/admin-guide/LSM/index.html

Fixes:
https://github.com/canonical/microk8s/issues/3161

#### Testing

~Currently untested -- I need to get the necessary tooling working to build and test.~ Tested install and boot on Manjaro linux with kernel version 5.17.

#### Possible Regressions

Older kernels without the separate security module folders in /proc/.../attr might now fail to install microk8s.

#### Checklist

* [n/a] The introduced changes are covered by unit and/or integration tests.
I don't think it's applicable in this case

### DIFF
--- a/microk8s-resources/wrappers/run-containerd-with-args
+++ b/microk8s-resources/wrappers/run-containerd-with-args
@@ -4,8 +4,16 @@ set -ex
 
 source $SNAP/actions/common/utils.sh
 # Re-exec outside of apparmor confinement
-if ! is_strict && [ -d /sys/kernel/security/apparmor ] && [ "$(cat /proc/self/attr/current)" != "unconfined" ]; then
-    exec aa-exec -p unconfined -- "$0" "$@"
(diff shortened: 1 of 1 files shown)


ktsakalozos (changes requested):
Thank you for this PR @MFAshby 

It is not ok to break older kernels. We should keep the old check paths and extend them with the new ones.

neoaggelos (changes requested):
Just to ensure that we do not break any existing kernels, would it be preferable to have something like this instead?

```
if [ "$(cat /proc/self/attr/current)" != "unconfined" ] || "$(cat /proc/self/attr/apparmor/current)" != "unconfined" ]; then
   exec aa-exec -p unconfined -- "$0" "$@"
fi

neoaggelos (approved):

Issue #3175, opened 2022-05-30
microk8s.kubectl should be used for internal health checking as using the system kubectl may fail


nobuto-m:
#### Summary

`microk8s enable <addon>` may fail especially when microk8s' Kubernetes version and the system's kubectl version are different.

#### What Should Happen Instead?

`microk8s enable <addon>` works without an error.

#### Reproduction Steps

1. Install microk8s explicitly with 1.21 (for Kubeflow on microk8s)

        sudo snap install microk8s --classic --channel=1.21

1. Install kubectl with the latest version (for other Kubernetes)

       sudo snap install kubectl --classic

1. Enable addons

       microk8s enable dns storage ingress metallb:10.64.140.43-10.64.140.49

```
$ microk8s enable dns storage ingress metallb:10.64.140.43-10.64.140.49
Traceback (most recent call last):
  File "/snap/microk8s/3202/scripts/wrappers/enable.py", line 43, in <module>
    enable(prog_name="microk8s enable")
  File "/snap/microk8s/3202/usr/lib/python3/dist-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/snap/microk8s/3202/usr/lib/python3/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/snap/microk8s/3202/usr/lib/python3/dist-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/snap/microk8s/3202/usr/lib/python3/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/snap/microk8s/3202/scripts/wrappers/enable.py", line 36, in enable
    enabled_addons, _ = get_status(get_available_addons(get_current_arch()), True)
  File "/snap/microk8s/3202/scripts/wrappers/status.py", line 157, in get_status
    kube_output = kubectl_get("all")
  File "/snap/microk8s/3202/scripts/wrappers/common/utils.py", line 169, in kubectl_get
    return run("kubectl", kubeconfig, "get", cmd, "--all-namespaces", die=False)
  File "/snap/microk8s/3202/scripts/wrappers/common/utils.py", line 39, in run
    result.check_returncode()
  File "/snap/microk8s/3202/usr/lib/python3.6/subprocess.py", line 389, in check_returncode
    self.stderr)
[...]

PR #3176, opened 2022-05-31
Make sure to use microk8s.kubectl

nobuto-m:
If we call kubectl simply, the system installed kubectl has precedence.
That may lead to some weird issues like an error due to a discrepancy
between kubectl cli and kubernetes API versions.

Closes: #3175

#### Testing

1. prepare `kubectl` on the system that always fails

       $ sudo ln -s /bin/false /usr/local/bin/kubectl
       $ kubectl; echo $?
       1

1. install the patched snap

       $ sudo snap install --dangerous --classic ./microk8s.snap

1. execute addon enable command, which was failing in #3175

1. confirm the command now works.

#### Checklist

### DIFF
--- a/scripts/wrappers/common/utils.py
+++ b/scripts/wrappers/common/utils.py
@@ -13,7 +13,7 @@ import yaml
 
 LOG = logging.getLogger(__name__)
 
-kubeconfig = "--kubeconfig=" + os.path.expandvars("${SNAP_DATA}/credentials/client.config")
+KUBECTL = os.path.expandvars("$SNAP/microk8s-kubectl.wrapper")
 
 
 def get_current_arch():
@@ -188,14 +188,19 @@ def ensure_started():
 
 def kubectl_get(cmd, namespace="--all-namespaces"):
     if namespace == "--all-namespaces":
-        return run("kubectl", kubeconfig, "get", cmd, "--all-namespaces", die=False)
+        return run(KUBECTL, "get", cmd, "--all-namespaces", die=False)
     else:
-        return run("kubectl", kubeconfig, "get", cmd, "-n", namespace, die=False)
+        return run(KUBECTL, "get", cmd, "-n", namespace, die=False)
 
 
 def kubectl_get_clusterroles():
     return run(
-        "kubectl", kubeconfig, "get", "clusterroles", "--show-kind", "--no-headers", die=False
+        KUBECTL,
+        "get",
+        "clusterroles",
+        "--show-kind",
+        "--no-headers",
+        die=False,
     )
 
 


nobuto-m:
I exercised a manual testing as per the description, and confirmed working as expected for the record.

> #### Testing
> 
>     1. prepare `kubectl` on the system that always fails
>        ```
>        $ sudo ln -s /bin/false /usr/local/bin/kubectl
>        $ kubectl; echo $?
>        1
>        ```
> 
>     2. install the patched snap
>        ```
>        $ sudo snap install --dangerous --classic ./microk8s.snap
>        ```
> 
>     3. execute addon enable command, which was failing in [microk8s.kubectl should be used for internal health checking as using the system kubectl may fail #3175](https://github.com/canonical/microk8s/issues/3175)
> 
>     4. confirm the command now works.

nobuto-m:
I exercised a manual testing as per the description, and confirmed working as expected for the record.

> #### Testing
> 
>     1. prepare `kubectl` on the system that always fails
>        ```
>        $ sudo ln -s /bin/false /usr/local/bin/kubectl
>        $ kubectl; echo $?
>        1
>        ```
> 
>     2. install the patched snap
>        ```
>        $ sudo snap install --dangerous --classic ./microk8s.snap
>        ```
> 
>     3. execute addon enable command, which was failing in [microk8s.kubectl should be used for internal health checking as using the system kubectl may fail #3175](https://github.com/canonical/microk8s/issues/3175)
> 
>     4. confirm the command now works.

neoaggelos (changes requested):
Thank you for this PR @nobuto-m

This is problematic for a few reasons, namely:
1. the /snap/bin folder may not be available in PATH, so `microk8s.kubectl` will not be available.
2. `microk8s.kubectl` will have issues in strict confinement, so we actively avoid using it inside the snap
3. the extra `kubeconfig` parameter is unnecessary if using microk8s.kubectl.

Note that these are not issues with the PR, they all pre-exist in the current state of the code as well.

Further, this code looks quite old, there is since a much better way of calling kubectl. See https://github.com/ubuntu/microk8s/blob/master/scripts/wrappers/reset.py#L23 and   https://github.com/ubuntu/microk8s/blob/master/scripts/wrappers/reset.py#L163 for details.

In general, it should suffie to change "kubectl" to `$SNAP.microk8s-kubectl.wrapper`, and remove the extra `kubeconfig` file location (it's already set in the wrapper script)

nobuto-m:
Thank you for the feedback.

> This is problematic for a few reasons, namely:
> 
>     1. the /snap/bin folder may not be available in PATH, so `microk8s.kubectl` will not be available.

Are you sure on this? I've tested this as per the test case, and `microk8s.kubectl` was actually called successfully.
https://github.com/canonical/microk8s/blob/3223749b74dead7c230406328a5a0daa20b13488/scripts/wrappers/common/utils.py#L43

>     2. `microk8s.kubectl` will have issues in strict confinement, so we actively avoid using it inside the snap

If that's the case, the original code was even worse since it's calling the **system**'s kubectl instead of kubectl inside the snap.

>     3. the extra `kubeconfig` parameter is unnecessary if using microk8s.kubectl.

Agreed.

> Note that these are not issues with the PR, they all pre-exist in the current state of the code as well.
> 
> Further, this code looks quite old, there is since a much better way of calling kubectl. See https://github.com/ubuntu/microk8s/blob/master/scripts/wrappers/reset.py#L23 and https://github.com/ubuntu/microk8s/blob/master/scripts/wrappers/reset.py#L163 for details.

If the current code and/or the legacy way is problematic, would you like to refactor those before or after this PR? Or do you expect me to do such refactoring?

> In general, it should suffie to change "kubectl" to `$SNAP.microk8s-kubectl.wrapper`, and remove the extra `kubeconfig` file location (it's already set in the wrapper script)

nobuto-m:
I've pushed another commit to use microk8s-kubectl.wrapper and will do a manual test later.

nobuto-m:
I believe the failure as:
FAILED tests/test-upgrade.py::TestUpgrade::test_upgrade
is not related to my chance since it looks like the recent commits have the same problem. In the meantime, I've manually tested the built snap and it works for a fresh installation as per the test case.

Screenshot from 2022-06-04 15-57-41

neoaggelos:
> is not related to my chance since it looks like the recent commits have the same problem

Correct, this is unrelated to your changes.

neoaggelos (approved):
LGTM, thanks @nobuto-m

nobuto-m:
Thanks. In the meantime, would it be possible to backport this change down to 1.21 channel? 1.21 is required to deploy Kubeflow and the issue is obvious when microk8s is 1.21 and the system kubectl is newer like 1.24 when `kubectl get all` is called.

neoaggelos:
Yes, this needs to be backported to all supported stable tracks.

Note that with the release of 1.24, 1.21 is officially out of support, but I guess we can make an exception for this particular bugfix, given that KubeFlow is still stuck in 1.21

PR #3182, opened 2022-05-31
Always regenerate csr.conf

dud225:
Since commit 82683e41282aee0153f6818591e92037db7fd3b0, the tool
refresh_certs.py needs to be told which certificate to refresh.
However when asking to refresh the front proxy client or the server
certificate, the csr.conf is read as is without csr.conf.template being
taken into consideration contrary to what is done when all the
certificates are renewed.
This fixes this inconsistency by always regenerating the file csr.conf.

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -417,7 +417,7 @@ _gen_server_cert() {
 gen_server_cert() {
     # Generate the server certificate
     # The file csr.conf is checked to determine if a regeneration of the cert must be done.
-    if compare_csr_conf; then
+    if csr_conf_updated; then
         _gen_server_cert
     fi
 }
@@ -431,7 +431,7 @@ _gen_proxy_client_cert() {
 gen_proxy_client_cert() {
     # Generate the proxy client certificate
     # The file csr.conf is checked to determine if a regeneration of the cert must be done.
-    if compare_csr_conf; then
+    if csr_conf_updated; then
         _gen_proxy_client_cert
     fi
 }
@@ -460,7 +460,7 @@ produce_certs() {
         ${SNAP}/usr/bin/openssl req -x509 -new -sha256 -nodes -days 3650 -key ${SNAP_DATA}/certs/front-proxy-ca.key -subj "/CN=front-proxy-ca" -out ${SNAP_DATA}/certs/front-proxy-ca.crt
     fi
 
-    if compare_csr_conf; then
+    if csr_conf_updated; then
         _gen_server_cert
         _gen_proxy_client_cert
         echo "1"
@@ -497,10 +497,10 @@ render_csr_conf() {
     fi
 }
 
-compare_csr_conf() {
+csr_conf_updated() {
     # Compares the file csr.conf.rendered with csr.conf
     # Return
-    #  0 if no change
+    #  0 if different
     #  1 otherwise.
     render_csr_conf
 
@@ -510,9 +510,9 @@ compare_csr_conf() {
 
     if ! "${SNAP}/usr/bin/cmp" -s "${SNAP_DATA}/certs/csr.conf.rendered" "${SNAP_DATA}/certs/csr.conf"; then
         cp ${SNAP_DATA}/certs/csr.conf.rendered ${SNAP_DATA}/certs/csr.conf
-        return 1
+        return 0
     fi
-    return 0
+    return 1
 }
 
 get_node() {
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -408,19 +408,40 @@ get_ips() {
     fi
 }
 
-gen_server_cert() (
+_gen_server_cert() {
     export OPENSSL_CONF="/snap/microk8s/current/etc/ssl/openssl.cnf"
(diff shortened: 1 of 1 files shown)


microk8s-resources/actions/common/utils.sh
```diff
@@ -485,6 +486,24 @@ render_csr_conf() {
     fi
 }
 
+compare_csr_conf() {
+    # Compares the file csr.conf.rendered with csr.conf
+    # Return
+    #  0 if no change
+    #  1 otherwise.
```
ktsakalozos, line 493: This function will return "0 if no change" so when used in a bash if statement it will evaluate to true if there is no change. So if we have something like 
```
if compare_csr_conf ; then
  generate_cert
fi
```
we will generate a certificate if there is no change in the csr file. This does not sound right.

microk8s-resources/actions/common/utils.sh
```diff
@@ -397,19 +397,40 @@ get_ips() {
...
-gen_proxy_client_cert() (
+gen_server_cert() {
+    # Generate the server certificate
+    # The file csr.conf is checked to determine if a regeneration of the cert must be done.
+    if compare_csr_conf; then
+        _gen_server_cert
+    fi
+}
+
+_gen_proxy_client_cert() {
```
ktsakalozos, line 414: I am not sure we are doing the right thing here. 

Let's say someone wants to reproduce the server certificates. The refresh_certs.py script in [1] will `rm server.crt` and then will call this `gen_server_cert` function. If there is no change in the `csr.conf.template` (which is expected) we will not be recreating the server.crt. 


[1] https://github.com/canonical/microk8s/blob/master/scripts/wrappers/refresh_certs.py#L139

ktsakalozos (changes requested):
Thank you for the PR. Please see the comments.

microk8s-resources/actions/common/utils.sh
dud225: Sorry I got confused, fixed.

microk8s-resources/actions/common/utils.sh
dud225: Yes correct. I would propose to remove the csr.conf file in all the relevant functions in `refresh_certs.py` as it is done for the function _reproduce_all_root_ca_certs_: https://github.com/canonical/microk8s/blob/master/scripts/wrappers/refresh_certs.py#L110=
WDYT?

ktsakalozos:
> However when asking to refresh the front proxy client or the server
> certificate, the csr.conf is read as is without csr.conf.template being
> taken into consideration contrary to what is done when all the
> certificates are renewed.
> This fixes this inconsistency by always regenerating the file csr.conf.

@dud225 could you please help me understand what is the effect of this inconsistency? When does this inconsistency becomes an issue?

Normally, when you change the `csr.conf.template` the apiserver-kicker will pickup the change and will reproduce the respective certificates [1]. So I would expect that under normal circumstances when someone calls the refresh-certs command any changes in the `csr.conf.template` are already in the `csr.conf`.

[1] https://github.com/canonical/microk8s/blob/master/microk8s-resources/wrappers/apiservice-kicker#L77

dud225:
I may miss something but in the current code there isn't any reference to `csr.conf.template` in `produce_certs` so I think that any modification to it is always ignored.

ktsakalozos:
In the current code `produce_certs` calls `render_csr_conf` [1] that uses `csr.conf.template` to produce a  `csr.conf.rendered`. If there are any changes between the `csr.conf.rendered` and the `csr.conf` currently used the respective certificates are re-issues. Any changes to the `csr.conf.template` are picked up by the apiservice-kicker in [2].

[1] https://github.com/canonical/microk8s/blob/master/microk8s-resources/actions/common/utils.sh#L449
[2] https://github.com/canonical/microk8s/blob/master/microk8s-resources/wrappers/apiservice-kicker#L76

dud225:
Hello
You are correct, however AFAIC I'm not falling into this condition. [We actually discussed my configuration together](https://kubernetes.slack.com/archives/CAUNWQ85V/p1637776299227600), long story short my chart is leveraging Multus to create additional IPs and I do not want those to trigger an update of the certs. So I want to keep control of the certificates update by using the _refresh-certs_ command. 
In other words I had to set `--advertise-address 0.0.0.0` to disable the _apiserver-kicker_ from refreshing the certificates.

I want to report that `microk8s refresh-certs --cert server.crt` is actualling behaving differently than what `microk8s refresh-certs --cert ca.crt` does (which is similar to what the _apiserver-kicker_ does).

iamwillbar:
> In the current code `produce_certs` calls `render_csr_conf` [1] that uses `csr.conf.template` to produce a `csr.conf.rendered`. If there are any changes between the `csr.conf.rendered` and the `csr.conf` currently used the respective certificates are re-issues. Any changes to the `csr.conf.template` are picked up by the apiservice-kicker in [2].
> 
> [1] https://github.com/canonical/microk8s/blob/master/microk8s-resources/actions/common/utils.sh#L449 [2] https://github.com/canonical/microk8s/blob/master/microk8s-resources/wrappers/apiservice-kicker#L76

@ktsakalozos does that mean that the step in [this documentation](https://microk8s.io/docs/services-and-ports#heading--auth) that advises to run `refresh-certs` after changing `csr.conf.template` is unnecessary?

Related to this, I assume the `csr.conf.template` would have to be updated on all nodes, is that correct?

dud225:
closed in favor of #3740

PR #3193, opened 2022-06-02
Pass credentials over the content interface

ktsakalozos:
#### Summary
Use the content interface to pass the microk8s credentials to juju (or anyone else that may need them)

#### Changes
Update the content interface with the right path to the k8s credentials

#### Testing
```
snap install microk8s --channel=latest/edge/MK-530
```
Build a tester snap https://github.com/ktsakalozos/microk8s-content-tester.git and snap install it.
Connect the interface:
```
snap connect tester:microk8s  microk8s:microk8s
```
Get a shell into the tester snap:
```
sudo snap run --shell tester.hello
```
The client.config file is available:
```
ls -lh $SNAP_DATA/credentials/
```

#### Notes
This interface will be in the MicroK8s strict snap.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -653,4 +653,5 @@ slots:
     interface: content
     content: microk8s
     source:
-      read: [$SNAP/.microk8s-info/microk8s]
+      read:
+        - $SNAP_DATA/credentials


wallyworld (commented):
This looks great. I assume in the juju snap we could set the target like this:

```
target: $SNAP_DATA/microk8s
```
So that we would then be able to access the config file like this:

`cat $SNAP_DATA/microk8s/credentials/client.config`

We can do a test with the juju snap and confirm it all works.

ycliuhw (commented):
It seems this revision is broken.

```
$ sudo snap install microk8s --channel=latest/edge/MK-530

$ mctr images list
Arguments file /var/snap/microk8s/3377/args/ctr is missing.

$ microk8s.ctr -h
Arguments file /var/snap/microk8s/3377/args/ctr is missing.

**Comment by ycliuhw** [comment]
> It seems this revision is broken.
> 
> ```
> $ sudo snap install microk8s --channel=latest/edge/MK-530
> 
> $ mctr images list
> Arguments file /var/snap/microk8s/3377/args/ctr is missing.
> 
> $ microk8s.ctr -h
> Arguments file /var/snap/microk8s/3377/args/ctr is missing.
> ```
Ok, it's a permission issue. Run it with `sudo` works.

**Comment by ycliuhw** [comment]
Juju can read the client.config at `$SNAP_DATA/credentials/client.config` which is awesome!
But we have two issues here that needs to be solved later:
-  the image pre-fetch hook is broken currently. Because the hook script calls `microk8s.ctr`;
- Juju uses `microk8s status` to check if the required addons (`dns` and `storage`) are enabled;

**Comment by ktsakalozos** [comment]
Hi @ycliuhw,

Is the installation error you got reproducible or is it some kind of race condition? I suspect the purge of the microk8s snap left the credentials directory behind probably because it was in use. Was the content interface connected? Was juju using anything from the credentials directory?

Regarding the dns and storage addons you will now be able to query the storage classes and pods so you do not need to use the status command anymore.

Populating the containerd cache with jujud-operator and juju-db images should also probably go through k8s. That would require some form of a dummy deployment.

**Comment by wallyworld** [comment]
> Populating the containerd cache with jujud-operator and juju-db images should also probably go through k8s. That would require some form of a dummy deployment.

This is an optimisation rather than a functional requirement; it would need to be done at snap install time - when either juju or microk8s snap is installed, if the other one is there, then the containerd registry needs to have the jujud operator image stored to make subsequent juju bootstraps feel fast.  As you say, perhaps juju could create a dummy pod to pull the image and then delete it; this would have the same effect.

**Comment by ycliuhw** [comment]
> Hi @ycliuhw,
> 
> Is the installation error you got reproducible or is it some kind of race condition? I suspect the purge of the microk8s snap left the credentials directory behind probably because it was in use. Was the content interface connected? Was juju using anything from the credentials directory?
> 
> Regarding the dns and storage addons you will now be able to query the storage classes and pods so you do not need to use the status command anymore.
> 
> Populating the containerd cache with jujud-operator and juju-db images should also probably go through k8s. That would require some form of a dummy deployment.

Hi @ktsakalozos 
I was not able to reproduce it after that. I guess it was because of the previous removal with --purge. But I am not sure how it happened. Right, if we can not access the microk8s command anymore, we perhaps have to do that via k8s API.

**Comment by ycliuhw** [comment]
Hi @ktsakalozos 
I found the testing snap uses `snap_microk8s` but not the `microk8s` group.
Is this change for the testing snap only or for future stable microk8s releases as well?
```
You can either try again with sudo or add the user ubuntu to the 'snap_microk8s' group:

    sudo usermod -a -G snap_microk8s ubuntu
    sudo chown -f -R ubuntu ~/.kube

After this, reload the user groups either via a reboot or by running 'newgrp snap_microk8s'.
```

ktsakalozos:
> I found the testing snap uses `snap_microk8s` but not the `microk8s` group.

The strict snap will use the `snap_microk8s` group while the classic snap the `microk8s` one.

ktsakalozos:
Thinterface is already in use by Juju so I am merging it.

Issue #2402, opened 2021-07-03
Change the internal IP


ArseniiPetrovich:
Can I somehow change the IP that the node is using as internal IP? My nodes has an VPN interconnection, and it seems they communicate well through the VPN. However when it comes to setting up addons like DNS I'm having a "No route to host" issue. I tracked it down to microk8s trying to use the "Internal IPs" of nodes instead of using their VPN IPs.

ktsakalozos:
@ArseniiPetrovich could you provide an example with IPs in your setup?

Normally the pods get an IP and the CNI and kube-proxy makes sure the communication is established among them. At which point do you get the "No route to host" error?

ArseniiPetrovich:
https://drive.google.com/drive/folders/18GFyEQ9gf6seLBfM72rUbdBQGuUQ-PXp?usp=sharing

Here, 3 tarballs from all our nodes. Would that answer your question?
I want clusters to connect with `10.11.11.x` addresses, while they connect with their primary addresses, either internal or external ones.

ArseniiPetrovich:
So I get an error when doing `mk8s enable/disable dns`

ArseniiPetrovich:
FYI, one of the nodes is currently out of cluster, I'm testing different options to fix this, but this does not work with 2 nodes either.

afeiszli:
I just ran into the same issue attempting the same process. 
1. Installed private interfaces on all Ubuntu servers using 10.99.x.x addresses (using Netmaker though this shouldn't matter)
2. Installed MicroK8s on top
3. Used the add-node command and selected the private interface
4. Ran the join command on the next node
5. repeat steps 3-4 on all nodes
6. Install CoreDNS

At this point it fails trying to reach the first node. Running "microk8s kubectl get nodes -o wide" I see that all nodes are using default interface IP's as their "internal IP" and no "external IP" is set.

afeiszli:
Update: to set internal IP's you can run the following (run as root on each node):

echo --node-ip=$(ip address show dev YOUR_PRIVATE_INTERFACE | grep 'inet ' | awk -F ' ' '{print $2}' | sed 's/\/24//g') >> /var/snap/microk8s/current/args/kubelet; microk8s stop; microk8s start

Replacing YOUR_PRIVATE_INTERFACE with the private interface name. Idk if this is correct or causes any different issues, but it at least sets the internal IP correctly.

ArseniiPetrovich:
[...]

PR #3216, opened 2022-06-06
ensure kubelet --node-ip is set correctly when joining a cluster

### LINKED ISSUE #3184: Kubelet node-ip parameter overrided while joining and join don't work
#### Summary
When joining a node that have `--node-ip` parameter set in `/var/snap/microk8s/current/args/kubelet` this parameter is overrided and take the value of the joined master; the node never join and microk8s crash on new node.

In `join.py`, i suspect the problem to come from the values in `info["kubelet_args"]` given by `get_connection_info` function to be written in `kubelet` file by the `store_base_kubelet_args` function.

We may respect the actual `node-ip` parameter configured before join.

#### Initial problem
I have to use two network interface and the only way to make it work is to hardcode ip: `--node-ip` in `args/kubelet file` and `--advertise-address` in `args/kube-apiserver`.

neoaggelos:
#### Summary

References #2402
Closes #3184 

#### Changes

- Make sure to set the `--node-ip` kubelet flag properly when joining a cluster where the flag is set.

#### Testing

1. Install MicroK8s on a node, then configure `--node-ip` kubelet argument
2a. Install MicroK8s on a second node, then join as a control plane or worker-only node
2b. Install MicroK8s on a second node, disable HA-cluster, join node.

In all cases, the `--node-ip` of the joining node should be updated and set to the correct IP address. 

#### Possible Regressions

This only affects clusters where the `--node-ip` argument has been explicitly set by the administrator. It does not have implications by default.

#### Checklist

* [X] The introduced changes are covered by unit and/or integration tests.

#### Notes

This needs to be backported to all supported tracks

### DIFF
--- a/scripts/cluster/join.py
+++ b/scripts/cluster/join.py
@@ -550,6 +550,17 @@ def store_base_kubelet_args(args_string):
     try_set_file_permissions(args_file)
 
 
+def update_kubelet_node_ip(args_string, hostname_override):
+    """
+    Update the kubelet --node-ip argument if it was set on the node that we join.
+
+    :param args_string: the kubelet arguments
+    :param hostname_override: the source IP address used by the node when joining
+    """
+    if "--node-ip" in args_string:
+        set_arg("--node-ip", hostname_override, "kubelet")
+
+
 def replace_admin_token(token):
     """
     Replaces the admin token in the known tokens
@@ -822,7 +833,7 @@ def join_dqlite_worker_node(info, master_ip, master_port, token):
     store_cert("serviceaccount.key", info["service_account_key"])
 
     store_base_kubelet_args(info["kubelet_args"])
-
+    update_kubelet_node_ip(info["kubelet_args"], hostname_override)
     update_cert_auth_kubeproxy(token, info["ca"], master_ip, master_port, hostname_override)
     update_cert_auth_kubelet(token, info["ca"], master_ip, master_port)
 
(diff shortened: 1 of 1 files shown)


ktsakalozos (approved):
LGTM +1 thank you @neoaggelos

usersina:
Although this fixes the issue with the mismatching node-ip, the ingress controller will not work on the newly joined worker node and will be in a `Pending` state.

This means that ingress is accessible from the controller node public ip, but not from the worker nodes, FWIW.

To replicate this, assuming that I have:
- `kube-controller` at 192.168.1.1
- `worker-node` at 192.168.1.2

### kube-controller

```bash
# Enable the needed addons
microk8s enable \
    hostpath-storage dns rbac \
    metrics-server dashboard ingress
microk8s enable registry --size 40Gi

# Update the `--node-ip` so that future workers use the correct interface
sudo echo --node-ip=192.168.1.1 >> /var/snap/microk8s/current/args/kubelet
sudo snap restart microk8s

# Verify that the internal IP is being used with
microk8s.kubectl get nodes -o wide

# Prepare the join command
microk8s add-node
```

### worker-node
```bash
microk8s join 192.168.1.1:25000/some-very-long-string/shorter-string --worker
```

It's also worth mentioning that `microk8s disable ingress` will get stuck forever at this point. Also force terminating the controller ingress pod on `worker-node` causes it to be stuck in a terminating state.

neoaggelos:
Hi @usersina, which MicroK8s version are you using? I am not able to replicate the issue you are describing following these commands.

A pod that is stuck in Pending and then terminating is most likely due to a changed kubelet IP address, so the `--node-ip` must be playing into this, but I'm not sure how. Can you create an issue with more details? Thanks!

usersina:
@neoaggelos, there is an open issue for it already where [I documented what worked for me](https://github.com/canonical/microk8s/issues/2402#issuecomment-1460214658) in addition to the versions I've used.

PR #3238, opened 2022-06-10
Calling inspect script properly in strict

ktsakalozos:
Be able to call `microk8s.inspect` (with dot) in the strict snap.


berkayoz (approved):
LGTM!

PR #3239, opened 2022-06-10
Inspect command to work on both strict and classic

ktsakalozos:
#### Summary
With this PR the inspect script should work as is in both stric and classic. There are a few "if is_strict" branches in the inspect script and minor fixes in the wrappers calling the inspection script. Also there should not be any denials thrown anymore.

#### Changes
- "if is_strict" branches in the inspect script
- updates in the wrappers calling the inspection script.

#### Testing
Manual testing. Build the strict and classic snap, install it and call the inspection command.


berkayoz (approved):
LGTM!

PR #3249, opened 2022-06-14
Install latest candidate snapd to run our tests on GH

ktsakalozos:
#### Summary
Run our test on GH with snapd from candidate this way we will be catching potentially breaking changes originating from snapd.

#### Changes
Update the gh actions spec.

#### Testing
In the actions of this PR


neoaggelos (commented):
Should this not be put under a separate step, like so:

```
- name: Switch to candidate snapd
  run: sudo snap refresh snapd --chanel=latest/candidate
```

ktsakalozos:
> Should this not be put under a separate step

Done!

PR #3263, opened 2022-06-21
fix configure hook not running on old installs

neoaggelos:
Fix regression from https://github.com/canonical/microk8s/commit/5ff8682524d29037a299ff54d09a885954668ad9#diff-7f1ea137a27c441e1be1e0ddb37091c53b09335f3124c2b47c4792e82b15290dR7

`$SNAP_DATA/var/lock/installed.lock` will not be there on old installs, which prevents the configure hook from running.

Revert change to only check for the lock file on the strict branch.

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -4,14 +4,12 @@ set -eux
 
 source $SNAP/actions/common/utils.sh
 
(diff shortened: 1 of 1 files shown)


ktsakalozos (approved):
LGTM

PR #3274, opened 2022-06-24
Fixes in strict cluster testing

ktsakalozos:
Closes the gap between strict an master branches and make sure the test-cluster.py runs on strict.


berkayoz (approved):
LGTM!

PR #3287, opened 2022-06-30
Imporved installation of snap before testing addons

ktsakalozos:
Two fixes:
- remove the `--classic` flag on installing a locally built snap
- wait for the installation to finish. This fixes a race condition where the first unit test of test-addons fails because the cluster is not yet ready.


berkayoz (approved):
LGTM!

PR #3298, opened 2022-07-04
Allow for confinement flag

ktsakalozos:
#### Summary
Allow for a "confinement" property in the addons.yaml to hide/disable the addons not available in classic or strict. For backwards compatibility if the property is missing from the addon's section the addon is considered valid for both classic and strict.

#### Changes
- Move the scripts/cluster python scripts under scripts/wrappers so as to remove some duplicate code (`is_strict()`).
- Update the `get_available_addons` to take into account the new property.
- Add a unit test

#### Testing
Via unit testing.


scripts/wrappers/common/utils.py
```diff
@@ -247,15 +247,28 @@ def is_community_addon(arch, addon_name):
...
+    strict = is_strict()
     for dir in os.listdir(snap_common() / "addons"):
         try:
             addons_yaml = snap_common() / "addons" / dir / "addons.yaml"
             with open(addons_yaml, "r") as fin:
                 addons = yaml.safe_load(fin)
 
             for addon in addons["microk8s-addons"]["addons"]:
-                if arch in addon["supported_architectures"]:
+                to_be_added = True
```
neoaggelos, line 258: Since the `to_be_added` variable is initially true and becomes false as needed, isn't it better to simply have a `continue` statement when any of the predicates below are false?

scripts/wrappers/common/cluster/utils.py
```diff
@@ -13,17 +13,6 @@
 import socket
 
 
-def get_group():
```
neoaggelos, line 16: let's ensure this does not break https://github.com/ubuntu/microk8s/blob/master/scripts/wrappers/addons.py#L179

neoaggelos (changes requested):
I'm slightly anxious about the renaming of all these files here, but it is fine.

Left a small code related comment and we can proceed.

berkayoz (approved):
LGTM!

PR #3306, opened 2022-07-05
Let calico handle offloading

ktsakalozos:


microk8s-resources/wrappers/apiservice-kicker
```diff
@@ -122,6 +98,11 @@ do
     then
       echo "Reloading the calico CNI"
       if (is_apiserver_ready)  &&
+          if ! grep -q "ChecksumOffloadBroken" "${SNAP_DATA}/args/cni-network/cni.yaml" &&
```
neoaggelos, line 101: should this not be a kubectl call?

neoaggelos: is this guaranteed to be correct though/ what happens in a cluster?

ktsakalozos: In a multi-node clusters we have two types of nodes:
- nodes with the flag in their cni.yaml. These nodes will not talk to the API server.
- nodes that got refreshed and they do not have the flag set on their cni.yaml. The first node of this type will patch the `felixconfigurations` forcing all the nodes to disable offloading. Any subsequent patching attempt will have no effect as it will not alter the `felixconfigurations`. This kind of nodes does talk to the API server.

What do you have in mind?

neoaggelos (approved):

PR #3324, opened 2022-07-15
Refactor snapcraft build pipeline

neoaggelos:
### Summary

Refactor and cleanup the `snapcraft.yaml` file for MicroK8s. The purpose is to reduce the toil of maintaining the build pipeline, as well as improve the development iteration speed.

A short description of current issues:
- Many parts re-use the same source directory, which leads to corrupted builds when files are updated
- Many parts were using the project root as source, which caused issues when changing git branches
- Different components were built with different Go versions, causing inconstencies and increased build speeds
- Some Go components build scripts were written prior to the existence Go modules, changed up the `$GOPATH` and consistently caused issues when projects could no longer build.
- During development iterations, changing any file required all binaries to build from scratch, which took very long.
- Inconsistencies in the way upstream projects are fetched, and how patches are applied.

### Changes

- This PR addresses all of the above issues, plus makes a few quality of life additions. Each component is now built using a recipe described in `./build-scripts/components/$component`, which consists of a repository, a git reference, a list of patches, as well as the script that builds the binaries and copies them to the install directory. See the `./build-scripts/components/README.md` file for details.

- Build Helm (version 3.9.1) and embed in the snap. This greatly improves the UX of using the `microk8s helm` commands. Further, this means that Helm2 is now removed, both `microk8s helm` and `microk8s helm3` are using Helm 3.9.1

- Fix the autocompletion scripts for the `microk8s.kubectl` command (which were not working in some shells)

- Add autocompletion scripts for the `microk8s.helm` command

- Ensure autocompletion works for the `microk8s kubectl` and `microk8s helm` commands.

- Ensure all binaries are built with the `-s -w` flags to reduce the snap size. 

- Reduce the size of the snapcraft.yaml file, remove duplicate or stale properties, in order to make it easier to reduce the diff with the strict branch.

- Reduce the logic built into snapcraft.yaml, which allows re-using it externally, as well as make it easier to eventually upgrade the core snap from core18 to core20/core22

### Testing

As a result, the snap size has been reduced by about 20MBs (230MB -> 210MB), even with the addition of the Helm binaries.


CODEOWNERS
```diff
@@ -1,2 +1,4 @@
 *               @ktsakalozos
```
ktsakalozos, line 1: Now that the addons are removed from the source tree we can remove this file. The entire team is responsible for all the code.

build-scripts/components/cluster-agent/version.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+echo "main"
```
ktsakalozos, line 3: Just a note. We should start creating branches for the cluster agent.

microk8s-resources/wrappers/microk8s-helm.wrapper
```diff
@@ -1,15 +1,3 @@
 #!/bin/bash
 
-set -eu
-
-export PATH="$SNAP/usr/sbin:$SNAP/usr/bin:$SNAP/sbin:$SNAP/bin:$PATH"
-ARCH="$($SNAP/bin/uname -m)"
-export LD_LIBRARY_PATH="$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/$ARCH-linux-gnu:$SNAP/usr/lib/$ARCH-linux-gnu"
```
ktsakalozos, line 7: I am a bit nervous about dropping this from the classic snap.

microk8s-resources/wrappers/microk8s-helm3.wrapper
```diff
@@ -1,15 +1,3 @@
...
-export LD_LIBRARY_PATH="$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/$ARCH-linux-gnu:$SNAP/usr/lib/$ARCH-linux-gnu"
-
-if [ ! -f "${SNAP_DATA}/bin/helm3" ]; then
-  echo "Helm 3 not available, try enabling it with 'microk8s enable helm3'" >&2
-  exit 0
-fi
-
-source $SNAP/actions/common/utils.sh
-"${SNAP_DATA}/bin/helm3" --kubeconfig="${SNAP_DATA}"/credentials/client.config "$@"
+"${SNAP}/bin/helm" --kubeconfig="${SNAP_DATA}"/credentials/client.config "$@"
```
ktsakalozos, line 3: I think that by moving the helm binary from {SNAP_DATA} to {SNAP} we will make it report as disabled due to

microk8s-resources/wrappers/microk8s-helm.wrapper
neoaggelos: Dropped due to this, can re-add if we feel safer for possible future changes:
```
$ ldd ./build-scripts/.install/bin/helm 
	not a dynamic executable
```

microk8s-resources/wrappers/microk8s-helm3.wrapper
neoaggelos: https://github.com/canonical/microk8s-core-addons/pull/70

ktsakalozos: Community addons are also using the helm binary from  {SNAP_DATA}.

When users refresh to 1.25 those addons will probably fail because the addons repo do not refresh along. For backwards compatibility lets copy the helm binary on {SNAP_DATA} in the configure/post-refresh hook. WDYT?

CODEOWNERS
ktsakalozos: @neoaggelos please remove the CODEOWNERS file.

microk8s-resources/wrappers/microk8s-helm3.wrapper
neoaggelos: copy the binary, or can we get away with a symlink (pointing to `current`)?

build-scripts/build-component.sh
```diff
@@ -0,0 +1,56 @@
...
+  git clone "${GIT_REPOSITORY}" --depth 1 -b "${GIT_TAG}" "${COMPONENT_BUILD_DIRECTORY}"
+else
+  cd "${COMPONENT_BUILD_DIRECTORY}"
+  git checkout "${GIT_TAG}"
+fi
+
+cd "${COMPONENT_BUILD_DIRECTORY}"
+git config user.name "MicroK8s builder bot"
+git config user.email "microk8s-builder-bot@canonical.com"
+if echo "${VERSION}" | grep -e rc -e alpha -e beta; then
```
ktsakalozos, line 37: `VERSION` should be `GIT_TAG`.

PR #3343, opened 2022-07-22
Automatically update list of API servers in worker nodes

neoaggelos:
### Summary

Update list of control plane nodes in worker nodes.

### Changes

- Split reconcile hooks into common, control-plane only and worker-only
- Enable apiservice-kicker for worker-only nodes
- Add hook that list the active endpoints for the `kubernetes` service and updates the traefik configuration file accordingly. The `kubelet.config` is used for this purpose, which has sufficient permissions even in RBAC enabled clusters.
- It is possible to disable this by creating the `$SNAP_DATA/var/lock/traefik-no-update` lock file.

### Testing

- Add worker node to single-node cluster, then add a second control plane. After a while, the worker uses both control plane nodes. Afterwards, remove initial node, the worker node keeps working.
- Upgrade worker node from 1.23 to latest version. The apiservice-kicker should now start running on the worker node, and the list of control plane nodes is updated.

### Implementation notes

- The internal is currently set to 30 seconds, we might want to change to something in the range of 1-10 minutes.

- Example output of `kubectl get endpoints kubernetes -o=yaml`

  <summary><details>

```
apiVersion: v1
kind: Endpoints
metadata:
  creationTimestamp: "2022-07-22T11:06:51Z"
  labels:
    endpointslice.kubernetes.io/skip-mirror: "true"
  name: kubernetes
  namespace: default
  resourceVersion: "80"
  uid: 4b75c83c-33c4-412a-86c4-91a1501e8990
subsets:
- addresses:
  - ip: 10.0.3.180
  ports:
  - name: https
    port: 16443
    protocol: TCP
```
  </details></summary>


neoaggelos:
Replaced by https://github.com/canonical/microk8s/pull/3347

PR #3344, opened 2022-07-22
Follow improvements for snapcraft build pipeline

neoaggelos:
### Summary

Follow up PR for #3324 

### Changes

- Move logic of initializing addon repositories out of snapcraft.yaml
- Specify the build directory for each component, which prevents rebuilding all components when a single one is changed
- Build kubelite with CGO_ENABLED=0
- Build go-migrator with CGO_ENABLED=0
- Move `snapcraftctl set-version` in a separate build step for simplicity
- Ensure runc (which requires CGO) has `build-attributes: [no-patchelf]` set
- Remove `build-attributes: [no-patchelf]` from parts that do not need it (microk8s-addons, kubernetes)
- Build the containernetwork plugins into a single binary, further reducing the installation size (cni tools reduced from 65MB to 8MB).

### DIFF
--- a/build-scripts/addon-repositories.sh
+++ b/build-scripts/addon-repositories.sh
@@ -1,7 +1,0 @@
-#!/bin/bash
-
-ADDONS_REPOS="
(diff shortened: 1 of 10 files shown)


ktsakalozos:
LGTM +1

PR #3347, opened 2022-07-25
Use API server proxy from cluster-agent and remove traefik 

neoaggelos:
#### Summary

Remove traefik from the snap, use the API server proxy from https://github.com/canonical/microk8s-cluster-agent/pull/7

#### Changes

- Add a hook that runs during the `configure` hook, migrates the old traefik configs into the new format expected by the apiserver proxy command
- Remove traefik service, add apiserver-proxy
- Change dqlite worker node joins to setup properly

#### Testing

Tested the following scenarios:

1. Join worker node into an existing cluster, ensure it can connect properly and spin up new pods.
2. Join worker node into an existing cluster with multiple nodes. Ensure it automatically picks up all API servers.
3. Join worker node with traefik proxy, upgrade to new snap, ensure configuration is migrated properly and that kubelet comes up.
4. Add new control plane to cluster after worker joins, ensure the API server proxy picks it up and uses it.
5. Remove control plane node from cluster, ensure worker drops it from the configuration.

#### Notes

PR on the cluster-agent repository: https://github.com/canonical/microk8s-cluster-agent/pull/7


snap/snapcraft.yaml
```diff
@@ -190,13 +190,6 @@ parts:
     source: build-scripts/
     override-build: ./build-component.sh etcd
 
-  traefik:
```
ktsakalozos, line 193: What happens when you refresh from a node with traefik to a node with the new apiserver-proxy? What happens to the old daemon?

neoaggelos: During the snap refresh, the traefik service is stopped and removed, handled by snapd

```
Jul 25 11:48:01 u2 snapd[1543]: services.go:919: RemoveSnapServices - disabling snap.microk8s.daemon-traefik.service
```

Old snap:

```
# before upgrade
root@u2:~# ps -fea | grep traefik
root       16573       1  0 11:46 ?        00:00:00 /snap/microk8s/x1/bin/traefik --configfile /var/snap/microk8s/x1/args/traefik/traefik.yaml
```

```
# after upgrade
root@u2:~# ps -fea | grep apiserver-proxy
root       18968       1  0 11:48 ?        00:00:00 /snap/microk8s/x2/bin/cluster-agent apiserver-proxy --config /var/snap/microk8s/x2/args/apiserver-proxy-config --kubeconfig /var/snap/microk8s/x2/credentials/kubelet.config --refresh-interval 30s
root       19777     636  0 11:49 pts/1    00:00:00 grep --color=auto apiserver-proxy
```

ktsakalozos (approved):
lgtm +1

PR #3394, opened 2022-08-17
Handle brnaches that changed names for local repos

ktsakalozos:
#### Summary
The core and community addons repositories are packaged inside the $SNAP. When these repos are added they are git cloned under $SNAP_COMMON following the branch the repo has in the $SNAP. The branch of the repo in the $SNAP may change as we ship new releases ie, main, 1.24, 1.25 etc. This change will cause the `git pull` performed in the `microk8s addons repo update` to fail because the branch name we follow disappeared. In this case we need to re-add the repository. 

Fixes: https://github.com/canonical/microk8s/issues/3392

#### Changes
Update the `microk8s addons repo update` command to be able to detect if the name of the branch a repository followed has changed.

#### Testing
To reproduce:
```
sudo snap install microk8s --channel=latest/edge --classic
sudo snap refresh microk8s --channel=1.24/stable
sudo microk8s addons repo update core
```

#### Possible Regressions
Any changes the user had done in the repo will be overwritten if the repository is re-added.

### DIFF
--- a/scripts/wrappers/addons.py
+++ b/scripts/wrappers/addons.py
@@ -11,7 +11,7 @@ import click
 import jsonschema
 import yaml
 
-from common.utils import get_current_arch, snap_common, get_group
+from common.utils import get_current_arch, snap_common, get_group, snap
 
 GIT = os.path.expandvars("$SNAP/git.wrapper")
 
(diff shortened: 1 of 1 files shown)


scripts/wrappers/addons.py
```diff
@@ -211,10 +211,28 @@ def update(name: str):
...
+        # If the branch name we follow has not changed a simple git pull is enough
+        # If the branch name changed we need to git repo add --force
+        followed_branch_name = subprocess.check_output(
+            [GIT, "rev-parse", "--abbrev-ref", "HEAD"], cwd=repo_dir, stderr=subprocess.DEVNULL
+        ).decode()
+        snapped_branch_name = subprocess.check_output(
+            [GIT, "rev-parse", "--abbrev-ref", "HEAD"], cwd=remote_url, stderr=subprocess.DEVNULL
+        ).decode()
+        if followed_branch_name != snapped_branch_name:
+            add(name, remote_url, None, True)
```
neoaggelos, line 229: This path does not set `commit_before_pull`, so reverting the update if the validation below fails will crash

scripts/wrappers/addons.py
```diff
@@ -211,10 +211,28 @@ def update(name: str):
...
+    remote_url = subprocess.check_output(
+        [GIT, "remote", "get-url", "origin"], cwd=repo_dir, stderr=subprocess.DEVNULL
+    ).decode()
+    if remote_url.startswith(snap()):
+        # This is a repository that we have in the snap.
+        # If the branch name we follow has not changed a simple git pull is enough
+        # If the branch name changed we need to git repo add --force
+        followed_branch_name = subprocess.check_output(
+            [GIT, "rev-parse", "--abbrev-ref", "HEAD"], cwd=repo_dir, stderr=subprocess.DEVNULL
+        ).decode()
```
neoaggelos, line 224: I'd add a `.strip()`, but this is fine as well, we only want to compare branches.

neoaggelos (commented):
Left some initial comments

berkayoz (approved):

scripts/wrappers/addons.py
```diff
@@ -211,18 +239,30 @@ def update(name: str):
...
+        snapped_branch_name = subprocess.check_output(
+            [GIT, "rev-parse", "--abbrev-ref", "HEAD"], cwd=remote_url, stderr=subprocess.DEVNULL
+        ).decode()
+        if followed_branch_name != snapped_branch_name:
+            clone_and_validate(remote_url, repo_dir)
+        else:
+            pull_and_validate(name, repo_dir)
+    else:
+        pull_and_validate(name, repo_dir)
```
neoaggelos, line 265: nit; i am not very fond of this sort of if-else-if-else, but it's fine

neoaggelos (approved):

PR #3406, opened 2022-08-25
Release workflow improvements

ktsakalozos:
Here they are:
- Move the test upgrade into the main repository
- Call the test-upgrade from the GH actions
- Call the test-upgrade on the test-distro.sh
- Call test-clustering on the test-distro.sh
- Make the test-upgrade path test all releases from 1.19 (vs 1.16) until now.


neoaggelos (approved):
In general LGTM. Minor question, should we remove validations that only do sanity checks that do not have anything to do with the upgrades? (e.g. checking for the default address pool of metallb)

Not an issue, PR look fine to me as-is

ktsakalozos:
> should we remove validations that only do sanity checks that do not have anything to do with the upgrades? (e.g. checking for the default address pool of metallb

The initial intend for the upgrade tests were to reuse the validators of the test-addons.py tests. Now that tes-addons.py and test-upgrades.py are moved to different repos we can go into specializing them.

PR #3430, opened 2022-09-09
Set kubernetes version based on manifest, set snap version as eskd version 

berkayoz:
#### Summary

Set kubernetes version based on manifest, set snap version as eskd version

#### Changes

Added the ability to set the kubernetes version based on the fetched eksd manifest
Set snap version to eksd version instead of kubernetes version
Organized and cleaned up code

#### Testing

Manually tested by snapping and installing the snap.

#### Checklist

* [X] The introduced changes are covered by unit and/or integration tests.

### DIFF
--- a/build-scripts/components/kubernetes/version.sh
+++ b/build-scripts/components/kubernetes/version.sh
@@ -2,18 +2,10 @@
 
 KUBE_TRACK="${KUBE_TRACK:-1.22}"            # example: "1.24"
 KUBE_VERSION="${KUBE_VERSION:-}"        # example: "v1.24.2"
-
-if [ -e "${SNAPCRAFT_STAGE}/KUBE_VERSION" ]; then
-  KUBE_VERSION=$(cat "$SNAPCRAFT_STAGE/KUBE_VERSION")
-fi
+EKS_TRACK=$(echo "${KUBE_TRACK}" | cut -f1,2 -d'.' | tr -s . -) # example: "1-24"
 
 if [ -z "${KUBE_VERSION}" ]; then
-  if [ -z "${KUBE_TRACK}" ]; then
-    KUBE_VERSION="$(curl -L --silent "https://dl.k8s.io/release/stable.txt")"
-  else
-    KUBE_TRACK="${KUBE_TRACK#v}"
-    KUBE_VERSION="$(curl -L --silent "https://dl.k8s.io/release/stable-${KUBE_TRACK}.txt")"
-  fi
+  KUBE_VERSION=$(curl -L --silent https://raw.githubusercontent.com/aws/eks-distro/main/projects/kubernetes/kubernetes/${EKS_TRACK}/GIT_TAG)
 fi
 
 echo "${KUBE_VERSION}"
--- a/build-scripts/components/kubernetes/version.sh
+++ b/build-scripts/components/kubernetes/version.sh
@@ -3,6 +3,10 @@
 KUBE_TRACK="${KUBE_TRACK:-1.22}"            # example: "1.24"
 KUBE_VERSION="${KUBE_VERSION:-}"        # example: "v1.24.2"
 
+if [ -e "${SNAPCRAFT_STAGE}/KUBE_VERSION" ]; then
+  KUBE_VERSION=$(cat "$SNAPCRAFT_STAGE/KUBE_VERSION")
+fi
+
 if [ -z "${KUBE_VERSION}" ]; then
   if [ -z "${KUBE_TRACK}" ]; then
     KUBE_VERSION="$(curl -L --silent "https://dl.k8s.io/release/stable.txt")"
--- a/build-scripts/eksd/build.sh
+++ b/build-scripts/eksd/build.sh
@@ -2,7 +2,7 @@
 
 set -ex
 
-GIT_TAG="$("$SNAPCRAFT_PROJECT_DIR/build-scripts/components/kubernetes/version.sh")"
+source "$SNAPCRAFT_PROJECT_DIR/build-scripts/components/kubernetes/version.sh"
 INSTALL="${1}"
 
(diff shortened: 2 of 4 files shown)


build-scripts/eksd/build.sh
```diff
@@ -11,10 +11,15 @@ else
   (cd "${SNAPCRAFT_PART_BUILD}/eks-distro" && git fetch --all && git pull)
 fi
 
-EKS_TRACK=$(echo "${GIT_TAG}" | cut -f1,2 -d'.' | tr -s . - | cut -c2-)
+EKS_TRACK=$(echo "${KUBE_TRACK}" | cut -f1,2 -d'.' | tr -s . -)
 EKS_RELEASE=$(cat "${SNAPCRAFT_PART_BUILD}"/eks-distro/release/"${EKS_TRACK}"/production/RELEASE)
 
+snapcraftctl set-version "$KUBE_TRACK.$EKS_RELEASE"
```
neoaggelos, line 17: I don't think this is correct. For example, eks v1.23-5 release is running kubernetes version v1.23.9, https://github.com/aws/eks-distro/releases/tag/v1-23-eks-5

This would set the version to `1.23.5` instead, which is not correct. Is it not enough to instead change https://github.com/canonical/microk8s/pull/3430/files#diff-3e91c14d99827f2b9997c7495162c4e899f582f684827e688caa516849b906a2R12 to pull the kuberentes version from the eks branch instead of the upstream?

That way, we can set the kubernetes version to whatever the eks version needs, and no other changes would be needed.

neoaggelos (changes requested):

build-scripts/eksd/build.sh
berkayoz: We already pull the kubernetes version from the manifest and set it accordingly in the build.
This change is mostly for the jenkins process, I would say that the snap version of the `eksd` set to `1-23-5` is a better fit since it follows a different versioning and release cycle and basically a fork. For example the current eksd version is `1-23-eks-5`, not `v1.23.9`.

I'm not sure if we should have the kubernetes version as the snap version. If we can add the eksd version as a seperate metadata and adapt the jenkins process accordingly, I would be fine with either. If not, it would create the need to maintain 2 seperate jenkins processes, because they would have some significant changes. We can apply the same jenkins jobs to `-eksd` with this change.

build-scripts/eksd/build.sh
```diff
@@ -11,10 +11,15 @@ else
...
+
 curl "https://distro.eks.amazonaws.com/kubernetes-$EKS_TRACK/kubernetes-$EKS_TRACK-eks-$EKS_RELEASE.yaml" -o "${SNAPCRAFT_PART_BUILD}/kubernetes-$EKS_TRACK-eks-$EKS_RELEASE.yaml"
 
 mkdir -p $INSTALL/images
-python3 parse.py "${SNAPCRAFT_PART_BUILD}/kubernetes-$EKS_TRACK-eks-$EKS_RELEASE.yaml" "$INSTALL/images"
+python3 parse-manifest.py "${SNAPCRAFT_PART_BUILD}/kubernetes-$EKS_TRACK-eks-$EKS_RELEASE.yaml" "$INSTALL/images"
+
+export KUBE_VERSION=$(cat "$SNAPCRAFT_PART_BUILD/eksd-components/kubernetes/git-tag")
```
neoaggelos, line 24: This is all we really need, this should be moved to `version.sh`. We can even just fetch it from https://raw.githubusercontent.com/aws/eks-distro/main/projects/kubernetes/kubernetes/1-23/GIT_TAG, no need to clone any repositories.

build-scripts/components/kubernetes/version.sh
```diff
@@ -3,6 +3,10 @@
 KUBE_TRACK="${KUBE_TRACK:-1.22}"            # example: "1.24"
 KUBE_VERSION="${KUBE_VERSION:-}"        # example: "v1.24.2"
 
+if [ -e "${SNAPCRAFT_STAGE}/KUBE_VERSION" ]; then
+  KUBE_VERSION=$(cat "$SNAPCRAFT_STAGE/KUBE_VERSION")
+fi
+
 if [ -z "${KUBE_VERSION}" ]; then
   if [ -z "${KUBE_TRACK}" ]; then
     KUBE_VERSION="$(curl -L --silent "https://dl.k8s.io/release/stable.txt")"
```
neoaggelos, line 12: This is all we really need to do to set the version, no need to mess with passing version numbers around.
```suggestion
KUBE_TRACK="${KUBE_TRACK:-1-22}"            # example: "1-22"

KUBE_VERSION=$(curl -L --silent https://raw.githubusercontent.com/aws/eks-distro/main/projects/kubernetes/kubernetes/${KUBE_TRACK}/GIT_TAG)
```

neoaggelos: @berkayoz ^ see my suggestion, we can then cleanup the rest of the parts. similarly, you can update the `kubernetes-version` part to append the eks release suffix as needed.

build-scripts/eksd/parse-manifest.py
```diff
@@ -9,6 +11,14 @@
...
             with open(f"{sys.argv[2]}/images-list", "w+") as list:
                 for component in manifest["status"]["components"]:
+                    component_path = "{}/eksd-components/{}".format(
+                        os.environ["SNAPCRAFT_PART_BUILD"], component["name"]
+                    )
+                    Path(component_path).mkdir(parents=True, exist_ok=True)
+
+                    with open(f"{component_path}/git-tag", "w+") as tag_file:
+                        tag_file.write(component["gitTag"])
+
```
neoaggelos, line 21: I don't think we need this

neoaggelos (commented):
Two more comments and it should be OK

ktsakalozos:
> nit; should help with keeping the diff concise.

Sorry @neoaggelos  I missed that.

PR #3437, opened 2022-09-13
Add test for airgap deployments with offline registry

neoaggelos:
### Summary

Add tests for airgap deployments with an offline registry.

Attempt to re-use as much of the machinery in `test-distro.sh` as possible.


ktsakalozos (approved):
LGTM +1 thank you @neoaggelos

PR #3441, opened 2022-09-13
Ignore --hostname-override for control plane and etcd nodes

neoaggelos:
Followup for #3434, ensure that all join scenarios are handled


ktsakalozos (approved):
LGTM +1

Issue #3501, opened 2022-10-11
[microk8s 1.25] KubeletNotReady status error for single node cluster of jetson nano 


AjayKumar4:
```
Ready            False   Tue, 11 Oct 2022 15:27:38 +0100   Tue, 11 Oct 2022 15:12:09 +0100   KubeletNotReady              container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized
```

<img width="1462" alt="Screenshot 2022-10-11 at 15 29 44" src="https://user-images.githubusercontent.com/12373874/195119442-8faad595-cae3-4d3d-84dc-cd872aa297f5.png">

[inspection-report-20221011_152323.tar.gz](https://github.com/canonical/microk8s/files/9757002/inspection-report-20221011_152323.tar.gz)

Workaround
Downgrade the microk8s to 1.24 will solve this issue

neoaggelos:
Hi @AjayKumar4 

Thank you for reporting this. I see the following failure in the cluster-dump that you sent:

```
MountVolume.SetUp failed for volume \"bpffs\" : hostPath type check failed: /sys/fs/bpf is not a directory
```

Can you attempt the following:

```
sudo snap install microk8s --channel 1.25 --classic
sudo vim /var/snap/microk8s/current/args/cni-network/cni.yaml
```

Inside the file, find and remove the following two sections and remove them:

```yaml

# (around line 4532)
             - name: bpffs
               mountPath: /sys/fs/bpf


# (around line 4558)
        - name: bpffs
          hostPath:
            path: /sys/fs/bpf
            type: Directory
```

Afterwards, re-apply the file with:

```
sudo microk8s kubectl apply -f /var/snap/microk8s/current/args/cni-network/cni.yaml
```

AjayKumar4:
@neoaggelos 
Nodes status changed to ready but for some strange reason pods are not initialising.
```
Events:
  Type     Reason       Age                    From     Message
  ----     ------       ----                   ----     -------
  Warning  FailedMount  2m18s (x2 over 4m36s)  kubelet  (combined from similar events): Unable to attach or mount volumes: unmounted volumes=[bpffs], unattached volumes=[cni-log-dir host-local-net-dir kube-api-access-tfsx8 xtables-lock cni-bin-dir cni-net-dir lib-modules var-run-calico policysync var-lib-calico bpffs]: timed out waiting for the condition
[...]

PR #3505, opened 2022-10-13
Remove unused bpf mounts on Calico pods

neoaggelos:
Calico CNI has some unused mounts related to eBPF, which break installation on some platforms, see #3501.

Remove them, since they were already disabled in the past.


ktsakalozos (approved):
LGTM +1

PR #3542, opened 2022-11-07
Fix runc and migrator builds, 1.24

ktsakalozos:
Two fixes on the build process:
- rework the migrator part to use the 1.19 go lang that includes the os.WriteFile
- fix the runc build by setting  GO111MODULE=on


neoaggelos (approved):
LGTM for the 1.24 repo. This is cleaned up in newer branches.

PR #3543, opened 2022-11-07
Fix runc and migrator builds, 1.23

ktsakalozos:
Two fixes on the build process:

- rework the migrator part to use the 1.19 go lang that includes the os.WriteFile
- fix the runc build by setting GO111MODULE=on


neoaggelos (approved):
LGTM, please verify that the changes to `plugin` and `source` do not add any extra files to the resulting snap.

PR #3546, opened 2022-11-07
Port kubelite patches to 1.26-alpha.3

ktsakalozos:
Porting the pre-stable patches to 1.26.


build-scripts/components/kubernetes/pre-patches/0000-Kubelite-integration.patch
```diff
@@ -397,10 +398,10 @@ index 00000000000..667b24f68e6
 +	println("Stopping kubelite")
```
neoaggelos, line 398: this looks new, is it needed?

neoaggelos (approved):

build-scripts/components/kubernetes/pre-patches/0000-Kubelite-integration.patch
ktsakalozos: This must be a github behavior. Maybe the having a patch in the diff between PRs breaks the presentation. The line in question, for example, already exists, it is not new: https://github.com/canonical/microk8s/blob/master/build-scripts/components/kubernetes/pre-patches/0000-Kubelite-integration.patch#L397

PR #3536, opened 2022-11-02
Update containerd to v1.6.9

dalbani:
#### Summary
This MR updates `containerd` to v1.6.9, the latest release at the time of writing.
It includes a updated version of `continuity`, which should fix the issue I reported at https://github.com/canonical/microk8s/issues/3499.

#### Changes

Update of `containerd` from v1.6.6 to v1.6.9.

#### Checklist

* [x] Read the [contributions](https://github.com/canonical/microk8s/blob/master/CONTRIBUTING.md) page.
* [x] Submitted the [CLA form](https://ubuntu.com/legal/contributors/agreement), if you are a first time contributor.
* [x] The introduced changes are covered by unit and/or integration tests.

PR #3550, opened 2022-11-09
[Backport 1.25] Update containerd to v1.6.9

neoaggelos:
Backport #3536 to 1.25 branch

### DIFF
--- a/build-scripts/components/containerd/version.sh
+++ b/build-scripts/components/containerd/version.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-echo "v1.6.6"
+echo "v1.6.9"
--- a/build-scripts/components/runc/version.sh
+++ b/build-scripts/components/runc/version.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-echo "v1.1.2"
+echo "v1.1.4"


dalbani:
Yeah, backporting to 1.25 is awesome.

dalbani:
@neoaggelos  
Sorry if it's not the right place to ask, but is there some explanation somewhere over the relation between the Git repository and the Snap store in terms of versions?
Is this a manual process for a new Snap to be pushed or does it get triggered automatically?
In practice, I'm trying to figure out when this backport will eventually land in my local installation 😄

neoaggelos:
In general, after a stable git branch is updated (master, 1.25, 1.24, 1.23, etc), CI tests are run and a push is built and published to the respective edge channel. 

```
$ snap info microk8s | grep 1.25/
  1.25/stable:           v1.25.3         2022-11-10 (4128) 174MB classic
  1.25/candidate:        v1.25.3         2022-11-07 (4128) 174MB classic
  1.25/beta:             v1.25.3         2022-11-07 (4128) 174MB classic
  1.25/edge:             v1.25.4         2022-11-11 (4193) 174MB classic
```

Then, a second round of more extensive is testing is done on edge builds and they are promoted to beta and then stable. As a rule, we only push to stable a week after, to have more trust that things do not break.

Hope this helps to clear things out. This is also documented (in more detail) in https://microk8s.io/docs/setting-snap-channel. If there is something confusing, feel free to raise an issue and we can improve the docs.

dalbani:
Ok, thanks for the information. 
And to be very concrete, is there a way to determine which Git commit ID is related to which revision of a particular snap?

neoaggelos:
Though not the best UX, in general this can be tracked in the Launchpad builders. For example, this is the list of builds for the MicroK8s 1.25 branch https://launchpad.net/~microk8s-dev/+snap/microk8s-1.25.

You can click on each build to see the snap revision number that was published, as well as the repository hash. For example, https://launchpad.net/~microk8s-dev/+snap/microk8s-1.25/+build/1938571 is the amd64 build for the commit 83f7af12566b0051ab861c871086f52927f62443, and was published with the revision number 4193.

PR #3573, opened 2022-11-17
remove -f flag from suggested chown cmd

### LINKED ISSUE #3534: microk8s suggests command hiding error info
#### Summary

When you try to run microk8s with insufficient permissions, you get an error and suggestion:

```console
$ microk8s status
Insufficient permissions to access MicroK8s.
You can either try again with sudo or add the user runner to the 'snap_microk8s' group:

    sudo usermod -a -G snap_microk8s runner
    sudo chown -f -R runner ~/.kube

After this, reload the user groups either via a reboot or by running 'newgrp snap_microk8s'.
```

However, if you don't have a `~/.kube` file, the `chown` command will fail silently and suppress the error message. This can cause failures e.g. in GitHub runners / scripts with `set -e`, and the `-f` option will hide the error message. This makes it very hard to debug the failure as there are no log messages.

Suggest to the user not to use the `-f` flag in `chown`. There is no reason to suppress the error message.
```console
$ microk8s status
Insufficient permissions to access MicroK8s.
You can either try again with sudo or add the user runner to the 'snap_microk8s' group:

    sudo usermod -a -G snap_microk8s runner
    sudo chown -R runner ~/.kube

After this, reload the user groups either via a reboot or by running 'newgrp snap_microk8s'.
```

barrettj12:
#### Summary

When the user has insufficient permissions, we suggest commands to run. We shouldn't suggest using the `-f` flag in `chown`, because this suppresses error messages which might be useful to the user.

Fixes #3534

#### Changes

Remove `-f` flag in `chown` in the "insufficient permissions" error.

#### Testing

N/A

#### Possible Regressions

N/A

#### Checklist

#### Notes

N/A

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -18,7 +18,7 @@ exit_if_no_permissions() {
     echo "You can either try again with sudo or add the user $USER to the '${group}' group:" >&2
     echo "" >&2
     echo "    sudo usermod -a -G ${group} $USER" >&2
-    echo "    sudo chown -f -R $USER ~/.kube" >&2
+    echo "    sudo chown -R $USER ~/.kube" >&2
     echo "" >&2
     echo "After this, reload the user groups either via a reboot or by running 'newgrp ${group}'." >&2
     exit 1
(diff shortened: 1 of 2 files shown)


neoaggelos (approved):
LGTM, thanks @barrettj12

PR #3581, opened 2022-11-23
Update images to pull from registry.k8s.io instead of k8s.gcr.io

ktsakalozos:
#### Summary
Update our default images to pull from registry.k8s.io instead of k8s.gcr.io

#### Changes
In the pause container and the airgap test.


neoaggelos (approved):
A follow up PR is needed in the core addons as well.

sachinkumarsingh092 (approved):

PR #3583, opened 2022-11-23
Upgrade from core18 to core20

berkayoz:
#### Summary

Fixed python packaging, switched missing libraries and updated yaml bits

#### Changes

* Switch from core18 to core20
* Replaced btrfs-tools with btrfs-progs and libbtrfs-dev
* Upgraded libssl1.0.0 to libssl1.1
* Changed python packaging to work with core20

#### Testing

Snapped, installed and tested manually on 18.04, 20.04, 22.04 with workloads.

#### Possible Regressions

Initial assessment seems like the changed packages do not cause a problem, there might be things that hasn't come up yet.

#### Checklist

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -127,15 +127,15 @@ parts:
     plugin: autotools
     source: https://www.netfilter.org/projects/libnftnl/files/libnftnl-1.1.8.tar.bz2
     build-environment:
-    - LIBMNL_LIBS: $SNAPCRAFT_STAGE/usr/local/lib
+      - LIBMNL_LIBS: $SNAPCRAFT_STAGE/usr/local/lib
 
   iptables:
     after: [libnftnl]
     source: https://www.netfilter.org/projects/iptables/files/iptables-1.8.6.tar.bz2
     plugin: autotools
     build-environment:
-    - LIBMNL_LIBS: $SNAPCRAFT_STAGE/usr/local/lib
-    - LIBNFTNL_LIBS: $SNAPCRAFT_STAGE/usr/local/lib
+      - LIBMNL_LIBS: $SNAPCRAFT_STAGE/usr/local/lib
+      - LIBNFTNL_LIBS: $SNAPCRAFT_STAGE/usr/local/lib
     autotools-configure-parameters:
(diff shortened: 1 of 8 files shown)


snap/snapcraft.yaml
```diff
@@ -177,6 +185,7 @@ parts:
       - git
       - grep
       - hostname
+      - ufw
```
neoaggelos, line 188: Why do we include ufw in the snap?

berkayoz: We are setting the python path environment variable globally in the snap to use the one we ship which is Python 3.8. ufw uses Python, and the snap uses the hosts ufw binary on classic which can lead to errors due to python version mismatch. For example the ufw on 22.04 doesn't seem to be compatible with Python3.8.

So the idea is to include the ufw so we can avoid using the host binary. I'm open to suggestions on how we can handle this.

ktsakalozos:
LGTM but we need to wait to branch out the 1.26 release.

ktsakalozos:
@berkayoz when trying to upload the snap package on the store I got this error:
```
The store was unable to accept this snap.
  - package contains external symlinks: lib/systemd/system/zfs-import.service -> /dev/null, usr/local/bin/iptables-xml -> /usr/local/sbin/xtables-legacy-multi
```

berkayoz:
@ktsakalozos I'm thinking removing [this line](https://github.com/canonical/microk8s/blob/ad8b56bd970c75e59d603df146b6d703919935ea/snap/snapcraft.yaml#L142) could get rid of the last symlink issue, I don't think we do anything special for zfs except staging the `zfsutils-linux` package. I'm not sure how to fix it, why do we need this package? What is depending on it?

ktsakalozos:
@berkayoz for the zfs-import service we can have a 
```
    override-prime: |
      snapcraftctl prime
      rm -vf lib/systemd/system/zfs-import.service
```

But for the iptables-xml I am not sure yet. The iptables deb package is pulled from the apt repo due to `ufw`.  however we want to build the iptables binaries in the dedicated part we have.

ktsakalozos (approved):
LGTM +1 thank you @berkayoz

PR #3605, opened 2022-12-01
docs: add live star-history graph to README.md

tianzhou:


tianzhou:
> Thank you for this PR @tianzhou
> 
> Although the star history is an important metric for any project, I am not sure if it belongs in the README. I do not see it often in projects.

Sure thing. It's your call.

PR #3691, opened 2023-01-25
Run GPU tests on 1.26

ktsakalozos:
#### Summary
Run the GPU tests on the host since all tests are run in lxc containers

#### Changes
This work is in test-distro.sh

#### Testing
Manual


berkayoz (approved):
One minor comment but LGTM.

PR #3709, opened 2023-01-30
Init configurations during install or with `snap set microk8s config=<config>`

neoaggelos:
### Summary

Allow setting MicroK8s configuration using:

- `snap set microk8s config="$(cat config.yaml)"` commands.
- An existing configuration file on the disk during the snap installation (for now: `/etc/microk8s.yaml` and `/root/.microk8s.yaml`)

### Implementation Notes

1. Configuration is only copied to `$SNAP_COMMON/etc/microk8s/{init,snap-set}.yaml` and is applied by the apiservice-kicker service. This decouples any long-running actions or actions that require the API server to be running (e.g. enable addons) from the actual snap installation.
2. If applied successfully, the files are renamed to have a `.applied` suffix and are then ignored. In the future, we might want to remove them completely after applying.
3. Configuration is applied using https://github.com/canonical/microk8s-cluster-agent/pull/11. In fact, this PR is non-functional until the `cluster-agent init` command implementation is merged.
4. We will not include a `snap set microk8s config-file=config.yaml` variant, since relative paths cannot be resolved by the `configure` hook.
5. The configure hook _must_ take care of creating `snap-set.yaml` only if the configuration has not been applied yet. This is because the configure runs in multiple scenarios (e.g. snap refresh), which would cause the configuration to be applied every time.

### Other Notes

The strict branch (https://github.com/canonical/microk8s/blob/strict/microk8s-resources/wrappers/apiservice-kicker#L103-L107) currently contains this hard-coded code:

```bash
      if [ -e $SNAP_COMMON/etc/launcher/configuration/config.yaml ] &&
         $SNAP/usr/bin/python3 $SNAP/scripts/launcher.py $SNAP_COMMON/etc/launcher/configuration/config.yaml
      then
        mv $SNAP_COMMON/etc/launcher/configuration/config.yaml $SNAP_COMMON/etc/launcher/configuration/config.yaml.applied
      fi
```

Perhaps there is value in adjusting the code accordingly, so that any configurations coming from the content interfaces are instead copied to `$SNAP_COMMON/etc/microk8s/`, and then no extra code is needed.

### Testing

Example config file:

```yaml
---
addons:
  - name: storage
  - name: dns
  - name: rbac
    disable: true
```

- `write /etc/microk8s.yaml && snap install microk8s`
- `snap install microk8s && snap set microk8s config="$(cat config.yaml)"`

The actual configuration testing is currently only unit-tested in https://github.com/canonical/microk8s-cluster-agent/pull/11

### DIFF
--- a/microk8s-resources/microk8s.default.yaml
+++ b/microk8s-resources/microk8s.default.yaml
@@ -1,0 +1,5 @@
+# Default launch configuration for MicroK8s.
+---
+version: 0.1.0
+addons:
+  - name: dns
--- a/microk8s-resources/wrappers/apiservice-kicker
+++ b/microk8s-resources/wrappers/apiservice-kicker
@@ -81,6 +81,23 @@ do
     # Run reconcile hooks
     $SNAP/usr/bin/python3 $SNAP/scripts/run-lifecycle-hooks.py reconcile || true
 
+    # Apply configuration files from $SNAP_COMMON/etc/launcher/
+    match_config_files="${SNAP_COMMON}/etc/launcher/"*.yaml
+    for config in ${match_config_files}; do
+      # bash will return the regex if no files were found, in that case skip
+      if [ "${config}" == "${match_config_files}" ]; then
+        continue
+      fi
+
+      echo "Applying ${config}"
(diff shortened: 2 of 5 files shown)


ktsakalozos:
LGTM +1

PR #3749, opened 2023-02-09
Added tests for mixed version clusters

sachinkumarsingh092:
#### Summary

Added tests for mixed version clustering after we removed flags from previous versions which were causing the newer node to fail while joining the older node cluster.

#### Changes

Introduced a new class `TestUpgradeCluster` in `test-cluster.py` which creates 2 nodes of different versions and joins them.

#### Possible Regressions

Hopefully none!!

#### Checklist

### DIFF
--- a/tests/test-cluster.py
+++ b/tests/test-cluster.py
@@ -39,7 +39,8 @@ class VM:
             self.vm_name = attach_vm
 
     def setup(self, channel_or_snap):
-        """Setup the VM with the right snap.
+        """
+        Setup the VM with the right snap.
 
         :param channel_or_snap: the snap channel or the path to the local snap build
         """
@@ -413,7 +414,7 @@ class TestCluster(object):
                     continue
             break
 
-    def test_worker_noode(self):
+    def test_worker_node(self):
         """
         Test a worker node is setup
         """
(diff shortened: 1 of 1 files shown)


tests/test-cluster.py
```diff
@@ -463,3 +464,119 @@ def test_no_cert_reissue_in_nodes(self):
...
+
+class TestUpgradeCluster(object):
+    @pytest.fixture(autouse=True, scope="module")
+    def setup_old_versioned_cluster(self):
+        """
+        Provision VMs of the previous version and form the cluster.
+        """
+        try:
+            print("Setting up cluster of an older version")
+            track, risk_level = channel_to_test.split('/')
```
ktsakalozos, line 477: What would the track be if we are on strict on eksd?

tests/test-cluster.py
```diff
@@ -463,3 +464,119 @@ def test_no_cert_reissue_in_nodes(self):
...
+            if track == "latest":
+                track = "1.26"   # TODO: remove this hardcoding
+            track = float(track) - 0.01
+            older_version = str(track) + '/' + risk_level
```
ktsakalozos, line 481: For risk level lets stick with stable so we do not need to deal with any issues on edge

tests/test-cluster.py
```diff
@@ -463,3 +464,119 @@ def test_no_cert_reissue_in_nodes(self):
...
+            print("Old version is {}".format(older_version))
+
+            type(self).VM = []
+            if not reuse_vms:
+                size = 3
```
ktsakalozos, line 486: We do not need an HA cluster for this test. A single node would be enough.

tests/test-cluster.py
sachinkumarsingh092: How do we tackle track being `latest` rather than a server version?

ktsakalozos: The latest track we should use the old stable track. For example now the current stable is 1.26, the latest track should build snaps that must cluster with the 1.25 versioned clusters. Have a look at https://github.com/canonical/microk8s/blob/master/tests/test-upgrade-path.py#L36 on how we figure out the current stable release.

tests/test-cluster.py
sachinkumarsingh092: Do we want to mix those clusters, the eksd or strict with the normal one?

ktsakalozos: normally we would not need to.

sachinkumarsingh092: So, these tracks will be mixed with the older version of themselves. Like 1.24-eksd with 1.23-eksd and so on.

tests/test-cluster.py
sachinkumarsingh092: Changing size to 1

tests/test-cluster.py
```diff
@@ -463,3 +465,139 @@ def test_no_cert_reissue_in_nodes(self):
...
+                track = str(float(track.split("-")[0]) - 0.01) + "-" + track.split("-")[1]
+            else:
+                track = float(track) - 0.01
+            older_version = str(track) + "/" + "stable"
+            print("Old version is {}".format(older_version))
+
+            type(self).VM = []
+            if not reuse_vms:
+                size = 1
+                for i in range(0, size):
```
ktsakalozos, line 508: If we have size of 1 why do we need the fore loop?

tests/test-cluster.py
```diff
@@ -463,3 +465,139 @@ def test_no_cert_reissue_in_nodes(self):
...
+                    print("Waiting for machine {}".format(i))
+                    vm.run("/snap/bin/microk8s.status --wait-ready --timeout 120")
+                    self.VM.append(vm)
+            else:
+                for vm_name in reuse_vms:
+                    vm = VM(backend, vm_name)
+                    vm.setup(older_version)
+                    self.VM.append(vm)
+
+            # Form cluster
```
ktsakalozos, line 521: If we have a cluster of size one we do not need to join nodes. The cluster is already formed.

tests/test-cluster.py
sachinkumarsingh092: For testing only, removed now.

tests/test-cluster.py
```diff
@@ -463,3 +465,121 @@ def test_no_cert_reissue_in_nodes(self):
...
+                    r = requests.get(release_url)
+                    if r.status_code == 200:
+                        last_stable_str = r.content.decode().strip()
+                        last_stable_str = last_stable_str.replace("v", "")
+                        last_stable_str = ".".join(last_stable_str.split(".")[:2])
+                        break
+                    else:
+                        time.sleep(3)
+                        attempt += 1
+
```
ktsakalozos, line 491: If we do not manage to get to https://dl.k8s.io/release/stable.txt we should not go on with the test.

tests/test-cluster.py
```diff
@@ -463,3 +465,121 @@ def test_no_cert_reissue_in_nodes(self):
...
+        # Wait for nodes to be ready
+        print("Waiting for two node to be Ready")
+        attempt = 0
+        while attempt < 10:
+            try:
+                connected_nodes = vm_older_version.run("/snap/bin/microk8s.kubectl get no")
+                num_nodes = connected_nodes.count(b" Ready")
+                if num_nodes != 2:
+                    time.sleep(5)
+                    continue
```
ktsakalozos, line 578: Shouldn't we increase the `attempts` counter here?

sachinkumarsingh092: The attempt counter is for how many times there was an error while listing the nodes rather than how many times there was an attempt to list the nodes.
We do a similar thing for other tests as well.

tests/test-cluster.py
sachinkumarsingh092: Makes sense. Fixed.

tests/test-cluster.py
ktsakalozos: Sure but if for some reason we break the test and the node does not get to Ready the test will just hung instead of timeout with an error. Lets fail after 5 mins or something like that.

sachinkumarsingh092: Fixed.

ktsakalozos (approved):
LGTM +1

PR #3750, opened 2023-02-10
Build libdqlite and libraft with debuginfo

cole-miller:
As discussed on Mattermost with @ktsakalozos.

Signed-off-by: Cole Miller <cole.miller@canonical.com>


ktsakalozos:
Thank you for this @cole-miller

PR #3775, opened 2023-02-17
Cluster agent built for FIPS

ktsakalozos:
Same as k8s-dqlite and k8s services. Building the cluster-agent with cgo support and having an env file to load the right libs before starting the service.

### DIFF
--- a/build-scripts/components/cluster-agent/build.sh
+++ b/build-scripts/components/cluster-agent/build.sh
@@ -4,5 +4,6 @@ export INSTALL="${1}/bin"
 export GOEXPERIMENT=opensslcrypto
 mkdir -p "${INSTALL}"
 
-make cluster-agent
+go mod tidy
+CGO_ENABLED=1 go build -o cluster-agent ./main.go
 cp cluster-agent "${INSTALL}"
--- a/build-scripts/components/cluster-agent/build.sh
+++ b/build-scripts/components/cluster-agent/build.sh
@@ -4,6 +4,5 @@ export INSTALL="${1}/bin"
 export GOEXPERIMENT=opensslcrypto
 mkdir -p "${INSTALL}"
 
-go mod tidy
 CGO_ENABLED=1 go build -o cluster-agent ./main.go
 cp cluster-agent "${INSTALL}"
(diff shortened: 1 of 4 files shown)


microk8s-resources/default-args/cluster-agent-env
```diff
@@ -0,0 +1,5 @@
+# In a node where FIPS is enabled, ie /proc/sys/crypto/fips_enabled is set to 1
+# Force FIPS to be used with. If not set GOFIPS is implicetely selected.
+export GOFIPS=0
```
neoaggelos, line 3: Should this be commented as well?

neoaggelos (changes requested):
Can we have the non-fips parts into master instead? Namely, the `cluster-agent-env` file.

build-scripts/components/cluster-agent/build.sh
```diff
@@ -4,5 +4,6 @@ export INSTALL="${1}/bin"
 export GOEXPERIMENT=opensslcrypto
 mkdir -p "${INSTALL}"
 
-make cluster-agent
+go mod tidy
```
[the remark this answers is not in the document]
ktsakalozos: Just blindly copied the Makefile. I removed it.

microk8s-resources/default-args/cluster-agent-env
ktsakalozos: No. The idea here is that even in a fips capable system we need to start normally (ignore FIPS) and then if the user needs to use FIPS she would need to point MicroK8s to the right libs.

microk8s-resources/wrappers/run-cluster-agent-with-args
```diff
@@ -14,6 +14,13 @@ export LD_LIBRARY_PATH="$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/$ARCH-linux-gnu:$SNAP/
 export LD_LIBRARY_PATH=$SNAP_LIBRARY_PATH:$LD_LIBRARY_PATH
 export OPENSSL_CONF="$SNAP/etc/ssl/openssl.cnf"
 
+set -a
+if [ -e "${SNAP_DATA}/args/cluster-agent-env" ]
+then
+  . "${SNAP_DATA}/args/cluster-agent-env"
+fi
+set +a
+
```
neoaggelos, line 23: We should have this on `master` as well, not only `feature/fips`

neoaggelos (approved):
LGTM, repeating my one comment for cluster-agent-env

PR #3776, opened 2023-02-17
Split testing steps into functions

berkayoz:
#### Summary

Split testing steps into functions so they can be callable as wanted.

#### Changes

Moved testing steps in seperate files with functions under `libs`

#### Testing

Manually tested.

#### Checklist


ktsakalozos:
Thank you for this PR @berkayoz could we have a way to call each test on its own from test-distro.sh? The end goal was/is to have each test in its own jenkins stage.

tests/test-airgap.sh
```diff
@@ -1,102 +1,22 @@
...
-
-airgap_wait_for_pods() {
-  container="$1"
-
-  lxc exec $container -- bash -c "
-    while ! microk8s kubectl wait -n kube-system ds/calico-node --for=jsonpath='{.status.numberReady}'=1; do
-      echo waiting for calico
-      sleep 3
-    done
+#!/usr/bin/env bash
```
ktsakalozos, line 1: Shouldn't this file be moved under libs as we do with the rest of the files?

berkayoz:
> Thank you for this PR @berkayoz could we have a way to call each test on its own from test-distro.sh? The end goal was/is to have each test in its own jenkins stage.

Thanks for the review @ktsakalozos 

You are correct, the end goal is to have each test suite in its own stage. However for better visibility, we can also put each part of the suite for example `setup_test_addons`, `test_core_addons`, `post_test_addons` into separate jenkins steps and call them like
```groovy
sh '''
  #!/usr/bin/env bash
  source tests/libs/addons.sh
  post_test_addons
'''

sh '''
  #!/usr/bin/env bash
  source tests/libs/addons.sh
  test_core_addons
'''

sh '''
  #!/usr/bin/env bash
  source tests/libs/addons.sh
  post_test_addons
'''
```
This will help with identifying which step has failed in a testing suite without requiring us to do digging through logs since the jenkins ui will highlight it directly. We probably can parse this data easier to view our performance/standing in tests.

There is some additional work required for this, which is parameterizing the functions instead of environment variables since steps don't share environments. Once a step is done, the environment is cleared. But we can use Jenkins variables to share and manage things like lxd container names we use.

PR #3780, opened 2023-02-20
Require root for microk8s addons commands

neoaggelos:
### Summary

Closes https://github.com/canonical/microk8s/issues/3730

The root cause behind the issue is the way git treats repositories where the current user is different from the owner of the repository.

Temporarily work around the issue by requiring root, which also makes sense from a security perspective.


ktsakalozos (approved):
LGTM +1

PR #3783, opened 2023-02-20
Make openssl we use configurable

ktsakalozos:
#### Summary
Allow users to set where the openssl should be used from.

#### Changes
References to openssl are patched to use an openssl_call function in utils.sh. openssl_call uses openssl from a user provided path or defaults to the openssl in the snap.

#### Testing
Manual.

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -583,34 +583,31 @@ get_ips() {
 }
 
 gen_server_cert() (
-    export OPENSSL_CONF="${SNAP}/etc/ssl/openssl.cnf"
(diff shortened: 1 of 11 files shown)


neoaggelos:
Perhaps instead of a utility function in `common/utils.sh`, we can do this a `/snap/microk8s/current/openssl.wrapper`? similar to what we have for `git.wrapper`? https://github.com/canonical/microk8s/blob/d5b30e8dc07773309b3638c2b05e590431778365/microk8s-resources/wrappers/git.wrapper#L8-L9

ktsakalozos:
Thank you for the review @neoaggelos. Switched to an openssl.wrapper, it is cleaner this way.

neoaggelos (approved):
LGTM!

PR #3806, opened 2023-02-28
Improve chained ops warning

ktsakalozos:
Improve the warning messaging because the "Chained operations" are our internal code name for enabling/disabling addons in one command.


berkayoz (approved):
LGTM!

PR #3809, opened 2023-03-01
Prevent concurrent addon enable/disable operations

neoaggelos:
### Summary

Use a lockfile to prevent `microk8s enable/disable` operations from running concurrently.

Draft until we consider the new UX acceptable.

### DIFF
--- a/scripts/wrappers/common/utils.py
+++ b/scripts/wrappers/common/utils.py
@@ -1,3 +1,4 @@
+import fcntl
 import getpass
 import json
 import os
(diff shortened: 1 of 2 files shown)


scripts/wrappers/common/utils.py
```diff
@@ -460,6 +483,12 @@ def xable(action: str, addon_args: list):
 
     addons = parse_xable_addon_args(addon_args, available_addons)
 
+    # NOTE(neoaggelos): This environment value is always passed on to the addon enable/disable
```
ktsakalozos, line 486: Wouldn't it be cleaner if the env variable was set right before calling the unprotected_xable (at https://github.com/canonical/microk8s/pull/3809/files#diff-cea8894ac12cb68765874a03fba91e79d22c596d86633886b5c02abe52829448R448)?

PR #3811, opened 2023-03-02
Containerd and runc build with CGO

ktsakalozos:
#### Summary
Build containerd and runc with CGO enabled so it can use FIPS certified libs

#### Changes
Changes in the build scripts and the env containerd runs under.

#### Testing
Manual. Enable GOFIPS and give it a wrong lib path to verify it is trying to find libs from there.

### DIFF
--- a/build-scripts/components/containerd/build.sh
+++ b/build-scripts/components/containerd/build.sh
@@ -4,6 +4,7 @@ INSTALL="${1}/bin"
 mkdir -p "${INSTALL}"
 
 for bin in ctr containerd containerd-shim containerd-shim-runc-v1 containerd-shim-runc-v2; do
+  export SHIM_CGO_ENABLED=1
   make "bin/${bin}"
   cp "bin/${bin}" "${INSTALL}/${bin}"
 done
--- a/build-scripts/components/runc/build.sh
+++ b/build-scripts/components/runc/build.sh
@@ -3,5 +3,6 @@
 export INSTALL="${1}/bin"
 mkdir -p "${INSTALL}"
 
+export SHIM_CGO_ENABLED=1
 make BUILDTAGS="seccomp apparmor" EXTRA_LDFLAGS="-s -w"
 cp runc "${INSTALL}/runc"
--- a/build-scripts/components/runc/build.sh
+++ b/build-scripts/components/runc/build.sh
@@ -3,6 +3,6 @@
 export INSTALL="${1}/bin"
 mkdir -p "${INSTALL}"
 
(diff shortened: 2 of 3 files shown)


build-scripts/components/containerd/build.sh
```diff
@@ -4,6 +4,7 @@ INSTALL="${1}/bin"
 mkdir -p "${INSTALL}"
 
 for bin in ctr containerd containerd-shim containerd-shim-runc-v1 containerd-shim-runc-v2; do
+  export SHIM_CGO_ENABLED=1
```
neoaggelos, line 7: While we are here, can we also set `STATIC=1` to get statically linked binaries?

build-scripts/components/runc/build.sh
```diff
@@ -3,5 +3,6 @@
 export INSTALL="${1}/bin"
 mkdir -p "${INSTALL}"
 
+export SHIM_CGO_ENABLED=1
 make BUILDTAGS="seccomp apparmor" EXTRA_LDFLAGS="-s -w"
```
neoaggelos, line 7: While we're here, can we also link `runc` statically?

```suggestion
make BUILDTAGS="seccomp apparmor" EXTRA_LDFLAGS="-s -w" static
```

neoaggelos (commented):
Looks good, two comments to build statically while we are at it.

Can we keep common things in `master` instead of `feature/fips`?

build-scripts/components/runc/build.sh
```diff
@@ -3,5 +3,6 @@
 export INSTALL="${1}/bin"
 mkdir -p "${INSTALL}"
 
+export SHIM_CGO_ENABLED=1
```
berkayoz, line 6: I'm not sure if this is used by runc. The runc makefile seems to use `CGO_ENABLED` as seen [here](https://github.com/opencontainers/runc/blob/69225fa9199431bf3656766c8a1274473e4bb923/Makefile#L49-L56). Maybe we can include both lines just to be safe.

berkayoz (commented):
It looks good overall, I have a small question.

ktsakalozos:
> Looks good, two comments to build statically while we are at it.

@neoaggelos lets work on the static building of these components in the respective card

> Can we keep common things in master instead of feature/fips?

I do not think anything from this PR should go in the master at this point. The goal here is to show that fips is possible. This PR breaks even the classic confinement. We will have to wait for a base fips core and not resort to this kind of hacks.

neoaggelos (approved):

berkayoz (approved):

PR #3812, opened 2023-03-02
Update airgap tests to use launch configurations

neoaggelos:
### Summary

Update airgap tests to use launch configurations.

As a side-chore, cleanup spacing and line endings.


tests/libs/airgap.sh
```diff
@@ -30,6 +30,22 @@ function setup_airgap_registry_mirror() {
   local TO_CHANNEL=$4
 
   create_machine "$NAME" "$DISTRO" "$PROXY"
+
+  lxc exec "$NAME" -- bash -c "
+    mkdir -p /root/snap/microk8s/common
+    echo '
+---
+version: 0.1.0
+extraKubeletArgs:
```
ktsakalozos, line 39: the dns addon should be setting these args, why do we set them explicitly here?

neoaggelos: skips an extra restart and lease timeouts for the kubernetes services, saves a few minutes per test

tests/libs/airgap.sh
```diff
@@ -96,26 +107,39 @@ function setup_airgapped_microk8s() {
...
-      cp hosts.toml /var/snap/microk8s/current/args/certs.d/$registry/hosts.toml
-    done
-
-    sudo snap restart microk8s.daemon-containerd
+  lxc exec "$NAME" -- bash -c '
+  mkdir -p /root/snap/microk8s/common
+  echo "
+---
+version: 0.1.0
+extraKubeletArgs:
```
ktsakalozos, line 115: Same as above.

tests/libs/airgap.sh
```diff
@@ -178,20 +202,16 @@ done
...
   push_images_to_registry "$REGISTRY_NAME"
-  echo "5/7 -- Install MicroK8s on an airgap environment"
+  echo "4/5 -- Install MicroK8s on an airgap environment (using registry mirror)"
   setup_airgapped_microk8s "$AIRGAPPED_NAME" "$DISTRO" "$PROXY" "$TO_CHANNEL"
-  echo "6/7 -- Configure MicroK8s registry mirrors"
-  configure_airgapped_microk8s_mirrors "$REGISTRY_NAME" "$AIRGAPPED_NAME"
-  echo "7/7 -- Wait for airgap MicroK8s to come up"
-  test_airgapped_microk8s "$AIRGAPPED_NAME"
+  echo "5/5 -- Wait for airgapped MicroK8s to come up"
+  airgap_wait_for_pods "$AIRGAPPED_NAME"
```
ktsakalozos, line 214: Could we have a "function" in the function definition as we do with the rest?

tests/libs/airgap.sh
neoaggelos: see https://github.com/canonical/microk8s/pull/3812#discussion_r1123488038

PR #3861, opened 2023-03-21
Fix kubelet arguments for 1.27

neoaggelos:
### Summary

Fixes

```
Mar 21 06:42:43 u1 microk8s.daemon-kubelite[69528]: Error: failed to parse kubelet flag: unknown flag: --container-runtime
Mar 21 06:42:43 u1 microk8s.daemon-kubelite[69528]: F0321 06:42:43.601537   69528 daemon.go:57] Kubelet exited failed to parse kubelet flag: unknown flag: --container-runtime
Mar 21 06:42:43 u1 microk8s.daemon-kubelite[69528]: Flag --containerd has been deprecated, This is a cadvisor flag that was mistakenly registered with the Kubelet. Due to legacy concerns, it will follow the standard CLI deprecation timeline before being removed.
```

Also contains a fix for the `skip_opt_in_local_config` helper function. This is because removing the `container-runtime` flag would also remove the `container-runtime-endpoint` flag, which is not desired.

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -192,7 +192,9 @@ skip_opt_in_local_config() {
     # argument $2 is the configuration file under $SNAP_DATA/args
     local opt="--$1"
     local config_file="$SNAP_DATA/args/$2"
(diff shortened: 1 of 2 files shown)


ktsakalozos (approved):
LGTM +1

PR #3874, opened 2023-03-27
Add conditions when using etcd instead of dqlite

sachinkumarsingh092:
#### Summary

MicroK8s status takes a long time when dqlite is not running. This is an issue because microk8s status in control plane nodes will always attempt to query the status of the dqlite database.

This is false in the following conditions:
- non-HA setup; etcd is used instead and dqlite is not running
- custom setup; an external etcd data store is used, not dqlite

Now we check the etcd endpoints configured in apiserver and act accordingly:

(a) if dqlite, keep current behaviour

(b) if non-ha, print etcd single-node address

(c) if etcd, print list of members (perhaps verify status with etcdctl member list)

#### Testing

tested locally

#### Checklist

### DIFF
--- a/scripts/wrappers/common/utils.py
+++ b/scripts/wrappers/common/utils.py
@@ -163,16 +163,16 @@ def get_etcd_info():
 
 
 def is_external_etcd():
-    external_etcd = 0
+    external_etcd = True
     kube_apiserver_args = os.path.expandvars("${SNAP_DATA}/args/kube-apiserver")
     with open(kube_apiserver_args, "r") as f:
         kube_apiserver_args_content = f.read()
         for line in kube_apiserver_args_content.split("\n"):
-            # All these variables should be present for an external etcd config
-            if "external-etcd" or "etcd-cafile" or "etcd-certfile" or "etcd-keyfile" in line:
-                external_etcd += 1
+            if "var/kubernetes/backend/kine.sock" in line:
+                external_etcd = False
+                break
 
-    return external_etcd == 4
+    return external_etcd
 
 
 def is_cluster_locked():
--- a/scripts/wrappers/common/utils.py
+++ b/scripts/wrappers/common/utils.py
@@ -154,10 +154,12 @@ def get_etcd_info():
         for line in kube_apiserver_args_content.split("\n"):
             if "etcd-servers" in line:
                 # list all etcd endpointss
-                for endpoint in line.split("=")[1].split(","):
-                    ip_port = endpoint.split("//")[1]
-                    etcd_endpoints.append(ip_port)
-                break
+                if "=" in line:
+                    for endpoint in line.split("=")[1].split(","):
+                        if "//" in endpoint:
+                            ip_port = endpoint.split("//")[1]
+                            etcd_endpoints.append(ip_port)
+                    break
 
     return etcd_endpoints
 
--- a/scripts/wrappers/common/utils.py
+++ b/scripts/wrappers/common/utils.py
@@ -153,28 +153,37 @@ def get_etcd_info():
         etcd_endpoints = []
         for line in kube_apiserver_args_content.split("\n"):
             if "etcd-servers" in line:
+                server_url = get_argumets(line)
                 # list all etcd endpointss
-                if "=" in line:
-                    for endpoint in line.split("=")[1].split(","):
-                        if "//" in endpoint:
-                            ip_port = endpoint.split("//")[1]
-                            etcd_endpoints.append(ip_port)
-                    break
+                for endpoint in server_url.split(","):
+                    if "//" in endpoint:
+                        ip_port = endpoint.split("//")[1]
+                        etcd_endpoints.append(ip_port)
+                break
 
     return etcd_endpoints
 
 
+def get_argumets(etcd_servers_arg):
+    server_url = None
+    parts = etcd_servers_arg.split("=")
+    if len(parts) == 2:
+        # Argument has an equals sign, e.g. "--etcd-servers=http://10.0.0.1:2379"
+        server_url = parts[1]
+    elif len(parts) == 1:
+        # Argument has a space, e.g. "--etcd-servers http://10.0.0.1:2379"
+        server_url = etcd_servers_arg.split("--etcd-servers")[1].strip()
+    return server_url
+
+
 def is_external_etcd():
-    external_etcd = True
     kube_apiserver_args = os.path.expandvars("${SNAP_DATA}/args/kube-apiserver")
     with open(kube_apiserver_args, "r") as f:
         kube_apiserver_args_content = f.read()
         for line in kube_apiserver_args_content.split("\n"):
             if "var/kubernetes/backend/kine.sock" in line:
-                external_etcd = False
-                break
-
-    return external_etcd
+                return False
+    return True
 
 
 def is_cluster_locked():
--- a/scripts/wrappers/common/utils.py
+++ b/scripts/wrappers/common/utils.py
@@ -146,6 +146,35 @@ def get_dqlite_info():
     return info
 
 
+def get_etcd_info():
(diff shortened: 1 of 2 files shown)


scripts/wrappers/common/utils.py
```diff
@@ -146,6 +146,35 @@ def get_dqlite_info():
...
+    external_etcd = 0
+    kube_apiserver_args = os.path.expandvars("${SNAP_DATA}/args/kube-apiserver")
+    with open(kube_apiserver_args, "r") as f:
+        kube_apiserver_args_content = f.read()
+        for line in kube_apiserver_args_content.split("\n"):
+            # All these variables should be present for an external etcd config
+            if "external-etcd" or "etcd-cafile" or "etcd-certfile" or "etcd-keyfile" in line:
+                external_etcd += 1
+
+    return external_etcd == 4
```
neoaggelos, line 175: What is the reasoning behind this? This does not look right. First, the condition is wrong, since it evaluates to:
```
("external-etcd") or ("etcd-cafile") or ("etcd-certfile") or ("etcd-keyfile" in line)
```
which always evaluates to true, since the first expression is a non-empty string.

Further, I don't think this is the right way to check for an external etcd, since cert authentication is optional?

I would instead suggest a check similar to (the relevant utils are already implemented):

```python
"var/kubernetes/backend/kine.sock:12379" not in get_argument(kube-apiserver, etcd-servers)
```

scripts/wrappers/status.py
```diff
@@ -34,9 +36,14 @@ def print_pretty(isReady, enabled_addons, disabled_addons):
     console_formatter = "{:>3} {:<20} # ({}) {}"
     if isReady:
         print("microk8s is running")
+        etcd_endpoints = get_etcd_info()
         if not is_ha_enabled():
             print("high-availability: no")
-        else:
+            if len(etcd_endpoints) > 0:
```
neoaggelos, line 42: more pythonic

```suggestion
            if etcd_endpoints:
```

scripts/wrappers/status.py
```diff
@@ -59,6 +66,11 @@ def print_pretty(isReady, enabled_addons, disabled_addons):
 
             print("{:>2}{} {}".format("", "datastore master nodes:", masters))
             print("{:>2}{} {}".format("", "datastore standby nodes:", standby))
+        else:
+            if len(etcd_endpoints) > 0:
```
neoaggelos, line 70: same here

neoaggelos (changes requested):

scripts/wrappers/common/utils.py
sachinkumarsingh092: Not using `12379` should be fine, right? The rest of the address should be fine.
Made the canges.

scripts/wrappers/common/utils.py
```diff
@@ -146,6 +146,35 @@ def get_dqlite_info():
...
 
+def get_etcd_info():
+    kube_apiserver_args = os.path.expandvars("${SNAP_DATA}/args/kube-apiserver")
+    with open(kube_apiserver_args, "r") as f:
+        kube_apiserver_args_content = f.read()
+        etcd_endpoints = []
+        for line in kube_apiserver_args_content.split("\n"):
+            if "etcd-servers" in line:
+                # list all etcd endpointss
+                for endpoint in line.split("=")[1].split(","):
```
neoaggelos, line 157: this will cause the `get_etcd_info` to fail if any single line does not have `=`.

scripts/wrappers/common/utils.py
```diff
@@ -146,6 +146,35 @@ def get_dqlite_info():
...
+                    ip_port = endpoint.split("//")[1]
```
neoaggelos, line 158: again, this breaks if the string is not formatted as expected, and it must not.

scripts/wrappers/status.py
```diff
@@ -59,6 +66,11 @@ def print_pretty(isReady, enabled_addons, disabled_addons):
 
             print("{:>2}{} {}".format("", "datastore master nodes:", masters))
             print("{:>2}{} {}".format("", "datastore standby nodes:", standby))
+        else:
+            if etcd_endpoints:
+                print("{:>2}{}".format("", "datastore endpoints:"))
+                for endpoint in etcd_endpoints:
+                    print("{:>3}{}".format("", endpoint))
```
neoaggelos, line 73:
```suggestion
        elif etcd_endpoints:
            print("{:>2}{}".format("", "datastore endpoints:"))
            for endpoint in etcd_endpoints:
                print("{:>3}{}".format("", endpoint))
```

neoaggelos (changes requested):

scripts/wrappers/common/utils.py
sachinkumarsingh092: Added a check for this.

scripts/wrappers/common/utils.py
sachinkumarsingh092: For this too.

scripts/wrappers/common/utils.py
```diff
@@ -146,6 +146,37 @@ def get_dqlite_info():
...
+    kube_apiserver_args = os.path.expandvars("${SNAP_DATA}/args/kube-apiserver")
+    with open(kube_apiserver_args, "r") as f:
+        kube_apiserver_args_content = f.read()
+        etcd_endpoints = []
+        for line in kube_apiserver_args_content.split("\n"):
+            if "etcd-servers" in line:
+                # list all etcd endpointss
+                if "=" in line:
+                    for endpoint in line.split("=")[1].split(","):
+                        if "//" in endpoint:
```
neoaggelos, line 159: As mentioned earlier, this is valid configuration but is not picked up from the current function
```
--etcd-servers http://10.0.0.1:2379
```

scripts/wrappers/common/utils.py
```diff
@@ -146,6 +146,37 @@ def get_dqlite_info():
...
+    external_etcd = True
+    kube_apiserver_args = os.path.expandvars("${SNAP_DATA}/args/kube-apiserver")
+    with open(kube_apiserver_args, "r") as f:
+        kube_apiserver_args_content = f.read()
+        for line in kube_apiserver_args_content.split("\n"):
+            if "var/kubernetes/backend/kine.sock" in line:
+                external_etcd = False
+                break
+
+    return external_etcd
```
neoaggelos, line 177:
```suggestion
    return True
```

neoaggelos (approved):

scripts/wrappers/status.py
```diff
@@ -34,9 +36,14 @@ def print_pretty(isReady, enabled_addons, disabled_addons):
     console_formatter = "{:>3} {:<20} # ({}) {}"
     if isReady:
         print("microk8s is running")
+        etcd_endpoints = get_etcd_info()
         if not is_ha_enabled():
             print("high-availability: no")
-        else:
+            if etcd_endpoints:
+                print("{:>2}{}".format("", "datastore endpoints:"))
+                for endpoint in etcd_endpoints:
```
ktsakalozos, line 44: Here is what this is printing:
```
microk8s is running
high-availability: no
  datastore endpoints:
   127.0.0.1:12379
addons:
  enabled:
    helm                 # (core) Helm - the package manager for Kubernetes
    helm3                # (core) Helm 3 - the package manager for Kubernetes
  disabled:
.....
```

Have a look at the indentation of 127.0.0.1. I bet people are parsing this output and this will break them.

sachinkumarsingh092: This is intentional, as there can be multiple URLs like this:
```
--etcd-servers=https://10.139.219.7:2379,https://10.139.219.148:2379,https://10.139.219.100:2379
```

ktsakalozos: We are missing a space before `127.0.0.1`.

sachinkumarsingh092: Done!

PR #3897, opened 2023-04-04
Added cidr guards on join and flag to disable cni management

berkayoz:
Added a check on join to see if node CIDRs are matching
Added an optional flag to let the users disable calico management

Tested manually by building [microk8s-cluster-agent#28](https://github.com/canonical/microk8s-cluster-agent/pull/28) and pointing the snap to use the built binary.

### DIFF
--- a/scripts/wrappers/join.py
+++ b/scripts/wrappers/join.py
@@ -28,7 +28,7 @@ from common.cluster.utils import (
     mark_no_cert_reissue,
     restart_all_services,
     get_token,
-    get_cluster_cidr
+    get_cluster_cidr,
 )
 
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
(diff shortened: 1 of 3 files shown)


scripts/wrappers/join.py
```diff
@@ -760,6 +761,14 @@ def join_dqlite(connection_parts, verify=False, worker=False):
...
     )
 
+    # Get the cluster_cidr from kube-proxy args
+    cluster_cidr = get_cluster_cidr()
+
+    if "cluster_cidr" in info and info["cluster_cidr"] != cluster_cidr:
+        print("Joining cluster failed. CIDR for the nodes does not match.")
+        print(f"Cluster CIDR: {info['cluster_cidr']} -- Node CIDR:  {cluster_cidr}")
+        exit(4)
+
```
neoaggelos, line 771: Is it not enough to just update the cluster-cidr in the local node kube-proxy?

ktsakalozos: +1 lets try to update the cidr of the joining node with the one we get from the cluster we are joining.

scripts/wrappers/join.py
```diff
@@ -760,6 +761,14 @@ def join_dqlite(connection_parts, verify=False, worker=False):
         worker=worker,
     )
 
+    # Get the cluster_cidr from kube-proxy args
+    cluster_cidr = get_cluster_cidr()
+
+    if "cluster_cidr" in info and info["cluster_cidr"] != cluster_cidr:
```
ktsakalozos, line 767: Shouldn't this be present in the join_etcd ?

PR #3900, opened 2023-04-05
Finalize launch configurations work

neoaggelos:
### Summary

Sibling PR for https://github.com/canonical/microk8s-cluster-agent/pull/29

Skip restarting of all services after successful join of dqlite master nodes. This is because dqlite and kubelite have already restarted, the rest of the services are OK.

- Launch configurations are no longer applied by the apiservice kicker, but the cluster agent, see https://github.com/canonical/microk8s-cluster-agent/pull/29/commits/52a56de98278cc696d36b75512cc53336dadd46e
- No need to restart all services after joining a cluster with dqlite as control plane node.


scripts/wrappers/join.py
```diff
@@ -728,8 +727,6 @@ def update_dqlite(cluster_cert, cluster_key, voters, host):
             waits -= 1
     print(" ")
 
-    restart_all_services()
```
ktsakalozos, line 731: I think that if we want to remove this line we should start/restart the API server after the k8s-dqlite service is up, or else we risk the API server timing out on k8s-dqlite and not starting at all. I guess it is a race condition we will not see often.

Issue #3927, opened 2023-04-17
microk8s status --wait-ready improvements 


beliaev-maksim:
hello

`microk8s status --wait-ready` can be slightly improved.

1. when timeout is added (eg `microk8s status --wait-ready --timeout 150`) it should return non-zero on exit if timed out
2. it should wait for all addons to be up. For example, if metallb is selected, microk8s says that it is ON. However, I see that pods are not available. (to reproduce just block quay.io in your proxy). So, wait ready really need to wait for all addon pods to start

neoaggelos:
Thanks for reporting this @beliaev-maksim. I want to provide some more context on this and write down some thoughts that I have been experimenting with. I would like to link this issue with #2572 

I think this is related to the "simple" nature of addons as they have been historically implemented in MicroK8s. Currently, MicroK8s does a very simplistic check to see if an addon is enabled: look for a specific resource in the cluster (e.g. a daemonset has been created) or a file in the local node.

This allows `microk8s status` to quickly determine which addons are "enabled" and which are "disabled". Obviously, this calls for false positives (e.g. if you manually deploy a resource that conflicts with what the respective addon deploys, the addon is shown as enabled).

An important note before moving on: The purpose of `microk8s status` is to very strictly check whether the MicroK8s cluster is up and running, along with the status of the underlying datastore. `microk8s status` is not at all involved with whether any particular addon has been enabled properly or is functioning as expected. Further, doing this in a careless manner would lead to bigger issues, since the `microk8s enable/disable` commands are using `microk8s status` internally to check that the Kubernetes services are up before proceeding.

---

I believe that what is described here reads to me more like a `readiness` check. An addon may be enabled (e.g. the respective daemonset exists) but not yet ready (e.g. the pods have not come up).

Now, a readiness check might vary greatly between addons. For others it might be just a `kubectl wait --for=condition=ready $pod`, whereas for others like `rbac` it might be something more involved (e.g. do an apiserver request impersonating a read-only user and expect it to fail.

One more important note, a `readiness` check is not equivalent to or meant to replace the addon tests. The `readiness` check should simply probe the respective addon resources and check whether they are ready to serve requests.

To that end, I would recommend two additions:

- Allow individual addons to have a `ready` hook, which executes a readiness check probe (with a timeout). If the addon is ready, the probe is expected to exit with status code 0. For simplicity and backwards compatibility, if an addon does not have a readiness check hook, then it is implicitly assumed to be ready.

[...]

PR #3937, opened 2023-04-18
exit with error code 1 if cluster is not ready after timeout

beliaev-maksim:
covers part of #3927
* exit with error code 1 if cluster is not ready after timeout
* refactor wait_for_ready to eliminate loop calculation
* refactor variable name according to PEP8

### DIFF
--- a/scripts/wrappers/status.py
+++ b/scripts/wrappers/status.py
@@ -211,7 +211,6 @@ if __name__ == "__main__":
     if wait_ready:
         is_ready = wait_for_ready(timeout)
         if not is_ready:
-            print("Cluster is not ready after {} seconds.".format(timeout))
             sys.exit(1)
     else:
         is_ready = is_cluster_ready()
--- a/scripts/wrappers/status.py
+++ b/scripts/wrappers/status.py
@@ -1,5 +1,6 @@
 #!/usr/bin/python3
 import argparse
+import sys
 
 from common.utils import (
     exit_if_no_permission,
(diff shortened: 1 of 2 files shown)


beliaev-maksim:
@ktsakalozos please have a look :)

scripts/wrappers/status.py
```diff
@@ -208,26 +209,29 @@ def ha_cluster_formed(info):
     yaml_short = args.yaml
 
     if wait_ready:
-        isReady = wait_for_ready(timeout)
+        is_ready = wait_for_ready(timeout)
+        if not is_ready:
```
ktsakalozos, line 213: Users are probably already parsing this command's output so I would not change that. Instead I would just change the return code in the case of is_ready != true

beliaev-maksim: do you mean `print("Cluster is not ready after {} seconds.".format(timeout))` ?

I do not think we have any message now. And I believe it will be good to emphasize that behaviour has changed and now we return non-zero

ktsakalozos: It is not ok to change the output of a script because users may be already parsing it in a very specific way. Here is an example of how this could go wrong. Let's assume a user has asked `--format yaml` and the `wait_for_ready` times out. Please do not change the out put of the status command.

scripts/wrappers/status.py
```diff
@@ -208,26 +209,28 @@ def ha_cluster_formed(info):
...
+            sys.exit(1)
```
ktsakalozos, line 214: This exit should be at the end of the script so we keep the same output.

beliaev-maksim: @ktsakalozos are we good to return 1 if `is_cluster_ready()` also False ?

beliaev-maksim: see last commit

ktsakalozos:
Thank you for this work @beliaev-maksim . Lets try to put it in with the 1.28 release. 

Could you also make sure you have signed the CLA?
```
Checking the following users on GitHub:
- beliaev-maksim ✕ (issue checking CLA status [HttpError: User does not exist or is not a member of the organization])

Checking the following users on Launchpad:
- beliaev.m.s@gmail.com ✕ (has not signed the CLA)
```

beliaev-maksim:
@ktsakalozos 
by now I did it twice :)
image

PR #3961, opened 2023-05-11
Allow argument to configure cadvisor containerd connection

ktsakalozos:
#### Summary
The observability addon fails to pull some metrics on containers. We allow the containerd argument so cadvisor knows where to search for the running containers.

#### Changes
Allow `--containerd` argument and put it in the default arguments.

#### Testing
Manual


berkayoz (commented):
Looks good overall, one small comment in #3962.

PR #3962, opened 2023-05-11
1.27 - Allow argument to configure cadvisor containerd connection

ktsakalozos:
#### Summary
Backport https://github.com/canonical/microk8s/pull/3961 to 1.27

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -133,6 +133,14 @@ then
   "$SNAP/bin/sed" -i 's@requestheader-client-ca-file=${SNAP_DATA}/certs/ca.crt@requestheader-client-ca-file=${SNAP_DATA}/certs/front-proxy-ca.crt@g' ${SNAP_DATA}/args/kube-apiserver
 fi
 
+# In 1.27 we removed the --containerd argument of cubelet. This however is used by cadvisor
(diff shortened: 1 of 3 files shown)


snap/hooks/configure
```diff
@@ -133,6 +133,14 @@ then
   "$SNAP/bin/sed" -i 's@requestheader-client-ca-file=${SNAP_DATA}/certs/ca.crt@requestheader-client-ca-file=${SNAP_DATA}/certs/front-proxy-ca.crt@g' ${SNAP_DATA}/args/kube-apiserver
 fi
 
+# In 1.27 we removed the --containerd argument of cubelet. This however is used by cadvisor
```
berkayoz, line 136: Shouldn't this also be in the PR for the main branch?

berkayoz (approved):
Looks good overall, one small question.

snap/hooks/configure
ktsakalozos: @berkayoz the patch in the `configure` hook is for users following the 1.27 track. With this patch we add the `--containerd` argument. The master branch feeds the latest/edge and latest/beta channels. These channels are meant to be used for testing and ephemeral deployments so we do not need to apply this patch there. The `--containerd` will land on such deployments as a default argument.

PR #3979, opened 2023-05-19
fixes iptables rule with custom CIDR

### LINKED ISSUE #3971: Mismatch between changing the pods CIDR config and the automatic iptables rules
#### Summary
Following [this guide](https://microk8s.io/docs/change-cidr) to change the pods CIDR instruct to modify two files `/args/kube-proxy` and `/args/cni-network/cni.yaml`. And it works fine.

But I also encountered the same issues as in #266 after rebooting my nodes. Somehow on only one node, the traffic from the cluster couldn't reach anything. iptables was reset to `-P FORWARD DROP`.

microk8s had added an automatic rule to allow forward traffic on the node. Using `iptables -S`:
```
-A FORWARD -s 10.1.0.0/16 -m comment --comment "generated for MicroK8s pods" -j ACCEPT
-A FORWARD -d 10.1.0.0/16 -m comment --comment "generated for MicroK8s pods" -j ACCEPT
```

Problem is, this is the default CIDR, and not the changed one.

Looking at the [code adding this rule here ](https://github.com/canonical/microk8s/blob/3ad1294c8c5f4ae8b08c9510d3975006358e8780/microk8s-resources/wrappers/run-kubelite-with-args#L91) it takes his cidr information from either `/args/kubelet` or `/args/flannel-network-mgr-config`, which is not the same as the files mentioned in the guide above. The flannel config file still have the old 10.1.0.0 value.

No clash between the CIDR configuration location, so that following the guide to change CIDR also set the correct iptables rules.

Dunge:
#### Summary
Fixes #3971

I used the kube-proxy arg file to get the proper CIDR.
I decided to also leave the existing files as a fallback when kube-proxy doesn't have the value.

#### Testing
I DID NOT TEST.
I do not have the capabilities to build and deploy a cluster from code.
Change should be straightforward enough.

#### Possible Regressions
Existing iptables rules with the wrong IP will remains.

#### Checklist

### DIFF
--- a/microk8s-resources/wrappers/run-kubelite-with-args
+++ b/microk8s-resources/wrappers/run-kubelite-with-args
@@ -81,10 +81,14 @@ else
 fi
 
 ## Kubelet configuration
-pod_cidr="$(cat $SNAP_DATA/args/kubelet | grep "pod-cidr" | tr "=" " "| gawk '{print $2}')"
+pod_cidr="$(cat $SNAP_DATA/args/kube-proxy | grep "cluster-cidr" | tr "=" " "| gawk '{print $2}')"
 if [ -z "$pod_cidr" ]
 then
-  pod_cidr="$(jq .Network $SNAP_DATA/args/flannel-network-mgr-config | tr -d '\"')"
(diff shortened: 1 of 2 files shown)


neoaggelos (approved):
LGTM!

PR #3982, opened 2023-05-20
Observability fix 1.27 strict

ktsakalozos:
#### Summary
Porting two fixes to 1.27-strict:
- the observability fix
- the gh actions trigger

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -133,6 +133,14 @@ then
   "$SNAP/bin/sed" -i 's@requestheader-client-ca-file=${SNAP_DATA}/certs/ca.crt@requestheader-client-ca-file=${SNAP_DATA}/certs/front-proxy-ca.crt@g' ${SNAP_DATA}/args/kube-apiserver
 fi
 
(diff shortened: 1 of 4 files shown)


snap/hooks/configure
```diff
@@ -133,6 +133,14 @@ then
   "$SNAP/bin/sed" -i 's@requestheader-client-ca-file=${SNAP_DATA}/certs/ca.crt@requestheader-client-ca-file=${SNAP_DATA}/certs/front-proxy-ca.crt@g' ${SNAP_DATA}/args/kube-apiserver
 fi
 
+# In 1.27 we removed the --containerd argument of kubelet. This however is used by cadvisor
+# to gather some metrics. We add it back here.
+if ! grep -e "\-\-containerd" "${SNAP_DATA}/args/kubelet"
+then
+  refresh_opt_in_config containerd \${SNAP_COMMON}/run/containerd.sock kubelet
```
neoaggelos, line 140: Should it not be this instead?

```suggestion
  refresh_opt_in_local_config containerd \${SNAP_COMMON}/run/containerd.sock kubelet
```

neoaggelos (changes requested):

PR #4013, opened 2023-06-02
Transition to certificate based auth from token based auth

ktsakalozos:
#### Summary
By default issue certificates for authentication. Replace the token based approach we have but keep backwards compatibility.

#### Changes
When initializing the cluster x509 certs are created and placed in the kubeconfig files.
When joining a cluster use the `auth_token` filed to decide if we are joining a cluster that is using tokens and configure properly the cluster.

#### Testing
Unit test to check the two join paths (token and certificate). 
The clustering tests check the join operations.
Manual testing.

#### Possible Regressions
Yes!

### DIFF
--- a/build-scripts/components/cluster-agent/version.sh
+++ b/build-scripts/components/cluster-agent/version.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-echo "main"
+echo "MK-1179/x509-auth"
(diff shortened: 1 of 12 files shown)


ktsakalozos:
This test is expected to fail because this PR gets all CIS checks passing while the cis-hardening addon expects one test to fail. As soon as this PR is merged we will need to update the cis-hardening addon.

berkayoz (approved):
LGTM overall.

neoaggelos (approved):

PR #4020, opened 2023-06-06
Permissions fixes in CIS hardened deployments

ktsakalozos:
#### Summary
This PR makes sure that when we create or edit files and we have applied CIS hardening we keep the file ownership group root and set access permissions to 600

#### Notes
This PR builds on top of https://github.com/canonical/microk8s/pull/4013 so it has to be merged after it.

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -9,6 +9,16 @@ get_microk8s_group() {
   fi
 }
 
+get_microk8s_or_cis_group() {
+  if [ -e $SNAP_DATA/var/lock/cis-hardening ]
+  then
+    echo "root"
+  else
+    get_microk8s_group
+  fi
+}
+
+
 exit_if_no_permissions() {
   # test if we can access the default kubeconfig
   if [ ! -r $SNAP_DATA/credentials/client.config ]; then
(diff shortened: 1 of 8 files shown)


microk8s-resources/actions/common/utils.sh
```diff
@@ -9,6 +9,16 @@ get_microk8s_group() {
...
+get_microk8s_or_cis_group() {
+  if [ -e $SNAP_DATA/var/lock/cis-hardening ]
+  then
+    echo "root"
+  else
+    get_microk8s_group
+  fi
+}
+
+
```
neoaggelos, line 21: I would much rather we kept this as `get_microk8s_group`. Reading `get_microk8s_or_cis_group` in a foreign place might be misleading.

Would we ever need the actual group if we have applied CIS hardening?

neoaggelos (approved):
LGTM, just a minor comment

microk8s-resources/actions/common/utils.sh
ktsakalozos: > Would we ever need the actual group if we have applied CIS hardening?

If we have applied CIS hardening the group of the files we creates should be root and not microk8s/snap_microk8s. This means that whenever we create a file we would either need to check if we are in CIS or push this check into a function like `get_microk8s_or_cis_group`. In practice this is not in many places, but I thought it would be better if we keep this check in a single place.

neoaggelos: Should it not be just `get_microk8s_group` then that takes CIS into account?

ktsakalozos: Ah, we have kept the `get_microk8s_group` as it was. It is still used in some places.

PR #4023, opened 2023-06-07
Fixed remove-node exception and disabled leave command on single node…

### LINKED ISSUE #4006: Non-HA single node, leaving node removes all pods
#### Summary
Iam running the microk8s in ubuntu in no high availability i.e. there is a single node onto the same machine where it is installed. I updated the certificates and then i issue following command:
`sudo microk8s leave`

It give following messages:

```
Generating new cluster certificates.
Waiting for node to start. 
```

and then i stopped microk8s and then started again, the node appeared  however all of my pods / namespaces are gone, how to recover?

All the pods should retain

#### Reproduction Steps
Already described above

#### Can you suggest a fix?

[https://wetransfer.com/downloads/ee891f9d62bd9ffd7fdac2f9597e638f20230529135310/bf5d38484b8a54a7107a5447153c884820230529135327/7b32a3](url)

berkayoz:
Fixed remove-node exception and disabled leave command on single node clusters

Closes #4006

### DIFF
--- a/scripts/wrappers/remove_node.py
+++ b/scripts/wrappers/remove_node.py
@@ -8,7 +8,7 @@ import sys
 import click
 import netifaces
 
-from ipaddress import ip_address, IPv4Address
+from ipaddress import ip_address
 
 from common.cluster.utils import (
     try_set_file_permissions,
--- a/scripts/wrappers/remove_node.py
+++ b/scripts/wrappers/remove_node.py
@@ -26,11 +26,11 @@ cluster_dir = "{}/var/kubernetes/backend".format(snapdata_path)
 def remove_dqlite_node(node, force=False):
     try:
         # If node is an IP address, find the node name.
-        is_node_ip = False
+        is_node_ip = True
         try:
-            is_node_ip = True if ip_address(node) is IPv4Address else False
+            ip_address(node)
         except ValueError:
-            pass
+            is_node_ip = False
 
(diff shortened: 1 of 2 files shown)


sachinkumarsingh092 (approved):
Tested locally an it works fine. LGTM.

scripts/wrappers/remove_node.py
```diff
@@ -25,7 +25,13 @@
 def remove_dqlite_node(node, force=False):
     try:
         # If node is an IP address, find the node name.
-        if type(ip_address(node)) is IPv4Address:
+        is_node_ip = False
+        try:
+            is_node_ip = True if ip_address(node) is IPv4Address else False
+        except ValueError:
+            pass
```
neoaggelos, line 32: Perhaps this is an easier way to write the same?

```suggestion
        is_node_ip = True
        try:
            ip_address(node)
        except ValueError:
            is_node_ip = False
```

neoaggelos (approved):
LGTM, just a minor comment

neoaggelos (changes requested):
Sorry, wrong button

scripts/wrappers/remove_node.py
berkayoz: Oh, I missed the `type()` check. Thanks for the review!

neoaggelos (approved):

PR #4056, opened 2023-06-29
Add label to disable juju webhooks on node objects

neoaggelos:
### Summary

Add `model.juju.is/disable-webhook=true` node label when the kubelet is trying to register the node.


neoaggelos:
From further testing, it appears that this is not really needed.

PR #4071, opened 2023-07-07
Add test for dual-stack

sachinkumarsingh092:
#### Summary
Add the tests for dual-stack configuration. For the test, we configure a `lxd` container with the required launch configuration. We then deploy a nginx deployment and services and ping on its IPv6 address to verify that IPv6 addresses are also supported.

### DIFF
--- a/tests/test-cluster.py
+++ b/tests/test-cluster.py
@@ -8,8 +8,10 @@ import requests
 import signal
 import subprocess
 from os import path
+from pathlib import Path
 from utils import (
     is_strict,
+    is_ipv6_configured,
 )
 
 
@@ -23,6 +25,7 @@ channel_to_test = os.environ.get("CHANNEL_TO_TEST", "latest/stable")
 backend = os.environ.get("BACKEND", None)
 profile = os.environ.get("LXC_PROFILE", "lxc/microk8s.profile")
 
+TEMPLATES = Path(__file__).absolute().parent / "templates"
 
 class VM:
     """
@@ -44,7 +47,7 @@ class VM:
             self.attached = True
             self.vm_name = attach_vm
 
-    def setup(self, channel_or_snap):
+    def setup(self, channel_or_snap, launch_configs=None):
         """
         Setup the VM with the right snap.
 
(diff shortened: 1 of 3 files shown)


ktsakalozos:
The lint check is failing.

tests/test-cluster.py
```diff
@@ -83,6 +86,28 @@ def _setup_lxc(self, channel_or_snap):
...
             time.sleep(20)
+
+            # Set launch configurations before installing microk8s
+            if launch_configs is not None:
+                print("Setting launch configurations")
+                cmd_prefix = "/snap/bin/lxc exec {}  -- script -e -c".format(self.vm_name).split()
+                cmd = ["mkdir -p /root/snap/microk8s/common/"]
+                subprocess.check_output(cmd_prefix + cmd)
+
+                file_path = '/microk8s.yaml'
```
ktsakalozos, line 97: Are we sure we are writing the right path to the right place? https://microk8s.io/docs/add-launch-config

tests/test-cluster.py
```diff
@@ -53,16 +56,16 @@ def setup(self, channel_or_snap):
         if (path.exists("/snap/bin/multipass") and not self.backend) or self.backend == "multipass":
             print("Creating mulitpass VM")
             self.backend = "multipass"
-            self._setup_multipass(channel_or_snap)
+            self._setup_multipass(channel_or_snap, launch_configs)
```
ktsakalozos, line 59: Where is the multipass setup update that takes into account this variable? I only see the lxc one.

tests/test-cluster.py
```diff
@@ -513,6 +538,105 @@ def test_no_cert_reissue_in_nodes(self):
             assert "no-cert-reissue" in lock_files.decode()
 
 
+class TestDualStack(object):
+    @pytest.mark.skipif(
+        not is_ipv6_configured and backend == "multipass",
```
ktsakalozos, line 543: Why do we skip this test on multipass? I read multipass has ipv6 support.

ktsakalozos:
> The lint check is failing.

Fixed

PR #4079, opened 2023-07-12
Add dual-stack upgrade and multinode tests

sachinkumarsingh092:
### Summary
Added tests for dual-stack during clustering and upgrading of the cluster. The tests for clustering are added as a part of TestCluster to save redundancy of code for similar operations. 
For upgrading, we deploy a dual stack enabled cluster and then refresh/upgrade it to a newer revision and check if it still supports dual stack.


tests/test-cluster.py
```diff
@@ -29,7 +29,7 @@
 extraCNIEnv:
   IPv4_SUPPORT: true
   IPv4_CLUSTER_CIDR: 10.3.0.0/16
-  IPv4_SERVICE_CIDR: 10.153.183.0/24
```
ktsakalozos: Why did you change that to 152? Is there a problem with 1.53?

sachinkumarsingh092: It was causing problems when we enabled DNS addon. This was because the [DNS IP](https://github.com/canonical/microk8s-core-addons/blob/main/addons/dns/enable#L50) is not in the range of this cidr.

tests/test-cluster.py
```diff
@@ -558,28 +573,63 @@ def test_no_cert_reissue_in_nodes(self):
             lock_files = vm.run("ls /var/snap/microk8s/current/var/lock/")
             assert "no-cert-reissue" in lock_files.decode()
 
+    @pytest.mark.skipif(
+        # If the host system does not have IPv6 support or is not configured for IPv6, it won't be able to create VMs with IPv6 connectivity.
+        backend != "lxc" or not is_ipv6_configured,
```
ktsakalozos, line 578: Multipass also supports IPv6, why are we explicit to lxc here?

tests/test-cluster.py
```diff
@@ -558,28 +573,63 @@ def test_no_cert_reissue_in_nodes(self):
...
+                    print(resp)
+                    break
+            except subprocess.CalledProcessError as e:
+                print("Error occurred during the request:", str(e))
+                raise
+            attempt -= 1
+            time.sleep(2)
+
 
 class TestDualStack(object):
```
ktsakalozos, line 624: Do we need this TestDualStack anymore? The test was moved to TestCluster so why do keep this class?

tests/test-cluster.py
sachinkumarsingh092: Due to the use of commands like these (https://github.com/canonical/microk8s/pull/4079/files#diff-8266c515875225e782094da6227c5ca720828f786e9f4c793b12e025e88f2e7aR585) which assume a lxc container for tests.

tests/test-cluster.py
```diff
@@ -259,14 +274,14 @@ def setup_cluster(self):
                 for i in range(0, size):
                     print("Creating machine {}".format(i))
                     vm = VM(backend)
-                    vm.setup(channel_to_test)
```
ktsakalozos, line 262: This is the place where you should check for IPv6 support and not inside the VM class.

tests/test-cluster.py
```diff
@@ -259,14 +274,14 @@ def setup_cluster(self):
...
+                    vm.setup(channel_to_test, launch_config)
                     print("Waiting for machine {}".format(i))
                     vm.run("/snap/bin/microk8s.status --wait-ready --timeout 120")
                     self.VM.append(vm)
             else:
                 for vm_name in reuse_vms:
                     vm = VM(backend, vm_name)
-                    vm.setup(channel_to_test)
+                    vm.setup(channel_to_test, launch_config)
```
ktsakalozos, line 284: Same as above.

tests/test-cluster.py
```diff
@@ -24,6 +24,17 @@
 channel_to_test = os.environ.get("CHANNEL_TO_TEST", "latest/stable")
 backend = os.environ.get("BACKEND", None)
 profile = os.environ.get("LXC_PROFILE", "lxc/microk8s.profile")
+launch_config = """---
+version: 0.1.0
+extraCNIEnv:
+  IPv4_SUPPORT: true
+  IPv4_CLUSTER_CIDR: 10.3.0.0/16
+  IPv4_SERVICE_CIDR: 10.152.183.0/24
```
ktsakalozos, line 32: Why we changed to 10.152.183.0 from 10.153.183.0?

tests/test-cluster.py
```diff
@@ -87,7 +98,9 @@ def _setup_lxc(self, channel_or_snap, launch_configs=None):
                 ).split()
             )
             time.sleep(20)
-            self._load_launch_configuration_lxc(launch_configs)
+            if is_ipv6_configured:
```
ktsakalozos, line 101: Do not do this here. Do the ipv6 check when you call the setup() method

tests/test-cluster.py
```diff
@@ -87,7 +98,9 @@ def _setup_lxc(self, channel_or_snap, launch_configs=None):
...
+                self._load_launch_configuration_lxc(launch_configs)
```
ktsakalozos, line 102: I also need to point out a blind spot you have. There is a second backend, multipass.

tests/test-cluster.py
sachinkumarsingh092: Same as above (https://github.com/canonical/microk8s/pull/4079#discussion_r1263715249)

tests/test-upgrade.py
```diff
@@ -106,6 +108,32 @@ def test_upgrade(self):
             except CalledProcessError:
                 print("Will not test the metallb addon")
 
+        if is_ipv6_configured:
+            launch_config = """---
+version: 0.1.0
+extraCNIEnv:
+  IPv4_SUPPORT: true
+  IPv4_CLUSTER_CIDR: 10.3.0.0/16
+  IPv4_SERVICE_CIDR: 10.152.183.0/24
```
ktsakalozos, line 117: please change this to 10.153.183.0 so it matches the extraSANs

tests/test-upgrade.py
```diff
@@ -106,6 +108,32 @@ def test_upgrade(self):
             except CalledProcessError:
                 print("Will not test the metallb addon")
 
+        if is_ipv6_configured:
```
ktsakalozos, line 111: Applying of the launch configuration needs to go when we first install the snap so we make sure it survives the refresh. If you put it here it will get ignored by the refresh.

tests/test-cluster.py
sachinkumarsingh092: Yeah but that is never triggered: https://github.com/canonical/microk8s/blob/6f488f61794fdd0ae30f626c8810acbfd0274768/tests/test-cluster.py#L188C9-L188C45. I'll fix this too

tests/test-cluster.py
ktsakalozos: Then we should implement this push in the VM class and we make the test below backend agnostic (as with the rest of the test) .

tests/test-cluster.py
ktsakalozos: We can drop the single node tests if we have multinode ones.

tests/test-cluster.py
ktsakalozos: This is a good catch. Lets create a follow up card to make the dns addon get the subnet from cni-env.

.github/workflows/build-snap.yml
```diff
@@ -125,6 +125,8 @@ jobs:
     steps:
       - name: Checking out repo
         uses: actions/checkout@v3.3.0
+      - name: Setup tmate session
```
ktsakalozos, line 128: Do we need this? Maybe we can keep it for debugging sessions

tests/templates/ingress.yaml
```diff
@@ -18,7 +18,7 @@ spec:
         app: microbot
     spec:
       containers:
-        - image: cdkbot/microbot-$ARCH
+        - image: cdkbot/microbot-amd64
```
ktsakalozos, line 21: Please do not commit this.

tests/test-cluster.py
ktsakalozos: Now that we updated the core addon the 10.153.183.0 should work. Lets Try this.

tests/test-cluster.py
```diff
@@ -87,7 +98,9 @@ def _setup_lxc(self, channel_or_snap, launch_configs=None):
                 ).split()
             )
             time.sleep(20)
+
```
ktsakalozos, line 101: Do we need these empty lines?

.github/workflows/build-snap.yml
sachinkumarsingh092: It blocks the pipeline, I'll comment it out.

tests/templates/ingress.yaml
sachinkumarsingh092: It always happens at local execution. Removing it.

tests/test-cluster.py
sachinkumarsingh092: Makes the separation b/w different steps easily visible imo.

tests/test-cluster.py
sachinkumarsingh092: Yes trying that.

tests/test-cluster.py
```diff
@@ -24,6 +24,17 @@
...
+version: 0.1.0
+extraCNIEnv:
+  IPv4_SUPPORT: true
+  IPv4_CLUSTER_CIDR: 10.3.0.0/16
+  IPv4_SERVICE_CIDR: 10.153.183.0/24
+  IPv6_SUPPORT: true
+  IPv6_CLUSTER_CIDR: fd02::/64
+  IPv6_SERVICE_CIDR: fd99::/108
+extraSANs:
+  - 10.152.183.1"""
```
ktsakalozos, line 37: This will also need to change

sachinkumarsingh092: `extraSANs` to 10.153.x.x?

PR #4094, opened 2023-07-20
Set sane defaults for fs and asyncio parameters

berkayoz:
Set sane defaults for fs and asyncio parameters.

Does not work in strict, not enough permissions to modify any of these values/files or do `sysctl` calls. 
`system-files` interface might let us create one of these files, although this should be discussed with the snapd team first.

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -53,6 +53,28 @@ then
   fi
 fi
 
+if ! is_strict
+then
+  # set sysctl parameters 
+  if ! grep -q '^fs.aio-max-nr=' /etc/sysctl.d/10-microk8s.conf
+  then 
+      echo "fs.aio-max-nr=1048576" >> /etc/sysctl.d/10-microk8s.conf
+  fi
+
+  if ! grep -q '^fs.inotify.max_user_instances=' /etc/sysctl.d/10-microk8s.conf
+  then 
+      echo "fs.inotify.max_user_instances=1024" >> /etc/sysctl.d/10-microk8s.conf
+  fi
+
+  if ! grep -q '^fs.inotify.max_user_watches=' /etc/sysctl.d/10-microk8s.conf
+  then 
+      echo "fs.inotify.max_user_watches=1048576" >> /etc/sysctl.d/10-microk8s.conf
+  fi
+
+  # reload sysctl parameters to take effect
+  sysctl --system
+fi
+
 # If the addons directory is missing from SNAP_COMMON, then we are upgrading from an older MicroK8s version.
 if [ ! -d "${SNAP_COMMON}/addons" ]
 then
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -53,28 +53,6 @@ then
   fi
 fi
 
-if ! is_strict
-then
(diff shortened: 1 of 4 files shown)


snap/hooks/configure
```diff
@@ -53,6 +53,28 @@ then
   fi
 fi
 
+if ! is_strict
+then
+  # set sysctl parameters 
+  if ! grep -q '^fs.aio-max-nr=' /etc/sysctl.d/10-microk8s.conf
+  then 
+      echo "fs.aio-max-nr=1048576" >> /etc/sysctl.d/10-microk8s.conf
```
ktsakalozos, line 61: What if the user has set this or any of the other limits to something larger somewhere else?

ktsakalozos:
Please rebase your code, there was a fix in the build of the snap that is causing the tests to fail.

ktsakalozos:
Placing this code in the configure hook makes me a bit worried. It is the first time we introduce sysctl as a dependency.  There are two sysctl, one on the host and one in the snap. I guess the safe to use is the one on the host as that would match the kernel on the system we run on, would this require us to have the LD_LIBRARY_PATH pointing to the libs on the host? Are we 100% certain the sysctl would always be on the host? Maybe this is extreme. If this fails on the configure hook the snap will silently stay in an endless refresh loop, trying to refresh every 6 hours.

Maybe we could try to do this when we start the services instead. For example, in the wrapper script that starts k8s-dqlite we could do something like
```
if skip-aio-tune.lock not set
  if current aio limit is too low
    try set aio
  fi
fi 
```
This would allow us to at least get the execution and potential failures in a visible location that we could debug and turn off if something goes wrong. WDYT?

snap/hooks/configure
berkayoz: How this works is that there are multiple locations with different precedence. The `/etc/sysctl.conf` takes the highest precedence amongst the files. The precedence goes as
* `/usr/lib/sysctl.d/*.conf` and `/etc/sysctl.d/*.conf`. The precedence is set by the prefix number (higher number - higher precedence)
* `/etc/sysctl.conf`

I've set it to lowest priority available, so any other file should override this if there are any customizations in that regard. Although there is a small possibility for anyone to create a file that could have a lower precedence. 

I've thought about checking first with `sysctl -n fs.aio-max-nr` and comparing if the value is larger but this does not guarantee the set number is actually persistent at reboot etc. 

Maybe we can `grep` and check the paths above with value comparisons just to be safe. I'll try to add this.

berkayoz:
I agree about moving out of the configure hook, I'll move it to the `k8s-dqlite` wrapper. For the path of the `sysctl` command, I'm not sure if the path of the `sysctl` would matter, as both binaries would just try to update parameters under `/proc/sys/`. This seems to some extent is kernel independent.

One other thing would be is that I believe `sysctl` as a tool might not be available on the host as well, in that case we'd like to use or atleast fallback to the bundled `sysctl`.  And seems like our `bash-utils` part already stage the  `procps` package so we've been bundling `sysctl` for a while. 

I'll address the comments and rebase the PR. Thanks for the review!

PR #4099, opened 2023-07-20
Rework cluster certificates generation

neoaggelos:
fixes a number of issues/bugs:

- properly update --secure-port on control plane nodes that join
- always ensure kubelet certificates include subject alternate names for hostname and IP
- remove duplicate python code for certificate generation
- misc cleanup and tidy code
- cleanup OPENSSL_CONF env configuration

Requires https://github.com/canonical/microk8s-cluster-agent/pull/41

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -1061,21 +1061,20 @@ server_cert_check() {
   openssl x509 -in "$SNAP_DATA"/certs/server.crt -outform der | sha256sum | cut -d' ' -f1 | cut -c1-12
 }
 
-########################################################################
-# Description:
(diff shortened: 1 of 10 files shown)


microk8s-resources/actions/common/utils.sh
```diff
@@ -1077,6 +1061,87 @@ server_cert_check() {
   openssl x509 -in "$SNAP_DATA"/certs/server.crt -outform der | sha256sum | cut -d' ' -f1 | cut -c1-12
 }
 
+########################################################################
```
ktsakalozos, line 1064: I know that the formatting is inconsistent in this file but could we please follow the function documentation scheme of for example `create_kubeconfig_x509`, where the description and arguments go under the function name?

PR #4100, opened 2023-07-20
Remove the validation server skip k8s patch

ktsakalozos:


neoaggelos (approved):
LGTM, we need this for 1.27-strict as well

PR #4104, opened 2023-07-21
Configure kube-apiserver-kubelet-client certificate

neoaggelos:
### Summary

Follow up of #4099 

Make sure that we properly configure the `kube-apiserver-to-kubelet` client, so that commands like `kubectl logs` and `kubectl port-forward` authenticate and work properly.

See `kube-apiserver-to-client` in https://kubernetes.io/docs/setup/best-practices/certificates/#certificate-paths

With this, the hack for kube-apiserver-to-kubelet role and role binding we used is no longer needed.

### Testing

- `echo '--authorization-mode=Webhook' | sudo tee -a /var/snap/microk8s/current/args/kubelet`
- `microk8s enable rbac`
- `sudo snap restart microk8s`
- `kubectl port-forward $anything`

Without change:

```bash
root@t1:~# k create deploy nginx --image nginx
root@t1:~# k port-forward deploy/nginx 80
error: error upgrading connection: unable to upgrade connection: Forbidden (user=127.0.0.1, verb=create, resource=nodes, subresource=proxy)
root@t1:~# 
```

After change

```bash
root@t1:~# k create deploy nginx --image nginx
Forwarding from 127.0.0.1:80 -> 80
Forwarding from [::1]:80 -> 80
```


ktsakalozos (approved):
LGTM +1

PR #4136, opened 2023-08-10
Added inotify limits collection and updated the suggested parameter c…

berkayoz:
Added inotify limits collection and updated the suggested parameter change command


ktsakalozos:
LGTM +1

PR #4184, opened 2023-08-30
Refactor handling of pods with hostPath mounts from SNAP_DATA

neoaggelos:
### Summary

Refactor how we restart pods that mount paths from SNAP_DATA, as this is needed across snap restarts

### Changes

Some pods might contain hostPath mounts for directories under `/var/snap/microk8s/current/...`. A common case of this is the Calico CNI pods.

When the Microk8s snap refreshes, the `/var/snap/microk8s/current/` symlink changes to point to the new revision, and the original directory is eventually unlinked from the filesystem. This means we have to recreate the pods after a snap refresh.

In the past, we have done this by rolling out a restart of Calico (if installed), Cilium (if installed) or KubeOVN (if installed), via hooks.

Instead, we now use the `kill-host-pods.py` script. This script retrieves the list of pods running on the cluster, filters then against a list of criteria and deletes them if they match.

- We ensure that we only delete pods from the current host (by making sure that the container IDs of the pods are known to containerd)
- We use the kubelet kubeconfig file, since worker nodes do not have a client kubeconfig file available.


microk8s-resources/default-hooks/reconcile.d/10-cni-restart
```diff
@@ -1,38 +0,0 @@
...
-fi
-
-if  ! [ -e "${SNAP_DATA}/var/lock/skip-offloading" ] &&
-    ! [ -e "${SNAP_DATA}/var/lock/offloading-patch-applied" ] &&
-      [ -e "${SNAP_DATA}/args/cni-network/cni.yaml" ] &&
-      [ -e "${SNAP_DATA}/var/lock/ha-cluster" ] &&
-      [ -e "${SNAP_DATA}/var/lock/cni-loaded" ] &&
-      [ -e "${SNAP_DATA}/var/lock/no-flanneld" ] &&
-    ! grep -q "ChecksumOffloadBroken" "${SNAP_DATA}/args/cni-network/cni.yaml" &&
-    "${KUBECTL}" patch felixconfigurations default --patch '{"spec":{"featureDetectOverride":"ChecksumOffloadBroken=true"}}' --type=merge
```
ktsakalozos, line 34: This patching is not needed anymore?

scripts/kill-host-pods.py
```diff
@@ -7,45 +7,69 @@
...
+
+
+def post_filter_has_snap_data_mounts(pod) -> bool:
+    """
+    Return true if a pod definition contains any volumes that mount paths
+    from `/var/snap/microk8s/current/...`, e.g. CNI pods
+    """
+    for volume in pod["spec"].get("volumes", []):
+        hostpath_volume = volume.get("hostPath", {})
+        if hostpath_volume.get("path", "").startswith(SNAP_DATA_CURRENT):
```
ktsakalozos, line 39: Given that $SNAP_DATA would also be eventually removed, shouldn't we also check for `/var/snap/microk8s/any-revision/` ?

scripts/kill-host-pods.py
```diff
@@ -7,45 +7,69 @@
...
-
-            subprocess.check_call(cmd)
+        if not post_filter_has_known_containers(pod, containers):
+            continue
+        if with_snap_data_mounts and not post_filter_has_snap_data_mounts(pod):
+            continue
+
+        meta = pod["metadata"]
+        cmd = [*KUBECTL, "delete", "pod", "-n", meta["namespace"], meta["name"]]
+        if dry_run:
```
ktsakalozos, line 69: I think that in the case of a dry run we should not call the delete command

neoaggelos: If this is a dry run, then we prepend `echo` in the command to simply print it out instead

scripts/kill-host-pods.py
neoaggelos: I'd say we only really care about `/var/snap/microk8s/current`. `/var/snap/microk8s/common` paths for example do not have the same issue.

I don't we mount pods with specific `/var/snap/microk8s/$revision` anywhere, why do you think we need this?

microk8s-resources/default-hooks/reconcile.d/10-cni-restart
neoaggelos: Not anymore, for some releases now this has been handled by the Calico manifest.

scripts/kill-host-pods.py
ktsakalozos: Missed that.

microk8s-resources/default-hooks/reconcile.d/10-cni-restart
ktsakalozos: Did we force deployments to move to the calico versions that address this issue?

scripts/kill-host-pods.py
ktsakalozos: In the past we used to put into the manifests "$SNAP_DATA". I do not see this anymore, we use the path with "current". Do you think it would be much more work to catch those cases too? At least as a fail safe?

neoaggelos: i'd much rather we leave them as-is. especially in a cluster setting, this would not be a good idea.

microk8s-resources/default-hooks/reconcile.d/10-cni-restart
neoaggelos: yes. deployments from 1.26 onwards have the right cni.yaml manifest. for deployments starting from older versions, this hook has run already while upgrading and can be therefore safely removed

PR #4192, opened 2023-09-03
Change executables to use full-paths in utils.sh

berkayoz:
Changed executables to use full path while being called in utils.sh

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -67,7 +67,7 @@ is_service_expected_to_start() {
 set_service_not_expected_to_start() {
   # mark service as not starting
   local service="$1"
-  run_with_sudo touch ${SNAP_DATA}/var/lock/no-${service}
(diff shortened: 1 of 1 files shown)


sachinkumarsingh092 (commented):
Can we assume utilities like `systemctl` or `snapctl` to be always present in `SNAP` path?

berkayoz:
> Can we assume utilities like `systemctl` or `snapctl` to be always present in `SNAP` path?

I believe snapctl is only available inside the snap, I'm not sure if systemctl has to go through the host. 
In addition making full paths seem a bit messy especially for readability.

Something like
```bash
do_something() {
    PATH="$SNAP/usr/bin:$SNAP/bin:$SNAP/usr/sbin:$SNAP/sbin"
}
```
to set the path for functions and in run with sudo
```bash
run_with_sudo() {
...
sudo -E "PATH=$PATH" sh -c "$@"
}
```
to pass it to sudo environment might be cleaner.

PR #4205, opened 2023-09-15
Improve handling of necessary system pod restarts on snap refresh

neoaggelos:
### Summary

Minor improvements on https://github.com/canonical/microk8s/pull/4184

### Changes

- Only restart pods with an owner reference, to protect standalone pods from being deleted and not recreated.
- Also restart pods that mount paths that are symlinks for `$SNAP_DATA` paths


ktsakalozos (approved):
LGTM +1

PR #4208, opened 2023-09-17
Symlinks for CNI conf and binary directories

neoaggelos:
### Summary

Create symlinks for `/etc/cni/net.d` and `/opt/cni/bin` directories (if possible), to ensure compatibility with standard Kubernetes CNI paths.

### Notes

- For strict this will use bind mounts instead


ktsakalozos (approved):
LGTM +1

PR #4211, opened 2023-09-19
Spread test for multiple distros

berkayoz:
Spread test for multiple distros


ktsakalozos:
containerd on centos 7 fails to create the containers with:
```
failed to create containerd container: get apparmor_parser version: apparmor_parser resolves to executable in current directory
```

One way to work around this is to add a `disable-apparmor = true` in containerd config, see https://github.com/containerd/containerd/issues/7547

PR #4256, opened 2023-10-18
Only verify partial CA chain

bschimke95:
If an intermediate CA is used, we cannot verify the whole chain since the CA needs to be self-signed for `openssl verify` to work.

We only want to verify that the server certificate is actually issued by the parent CA (either a root CA or intermediate), thus verifying the partial chain is sufficient.

See also: https://stackoverflow.com/a/60831762

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -713,9 +713,10 @@ produce_certs() {
 
 ensure_server_ca() {
     # ensure the server.crt is issued by ca.crt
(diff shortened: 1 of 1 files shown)


bschimke95:
I manually verified that this fixed the "server.crt regenerates every 5s" issue.
As a result the fingerprint verification succeeds and joining a node to a cluster also works.

neoaggelos (approved):
I would prefer letting the user provide the full CA chain, and use that if possible.

This is acceptable, though, given our immediate requirements.

PR #4268, opened 2023-10-26
temporarily use latest/stable snapd to unblock CI

neoaggelos:
### Summary

Temporarily use latest/stable snapd to unblock CI


ktsakalozos:
LGTM +1

PR #4270, opened 2023-10-27
Add FIPS and strict patches

neoaggelos:
### Summary

Work in progress


ktsakalozos:
I am not totally sold on this. I like how we check that the patches continue to apply without errors.

I do not like that these patches live in the master branch. It would be best if they are kept in their respective branches. The force-push is also not good, we are losing the history.

PR #4271, opened 2023-10-30
Cleanup unused snap dependencies

neoaggelos:
### Summary

Do a pass over dependencies that are not required for the snap to work. Adjust minor parts of the code accordingly.

### Changes

- Remove unused `aufs-tools`, `socat`, `squashfs-tools` packages
- Use `ufw` from the host (when available)
- Cleanup `containerd` part (remove unused stage file filters, drop conntrack and libs as they are not required, containerd is now a static binary)
- Drop perl and perl5 scripts (unused, only pulled due to git dependency)
- Drop python wheel files from result snap
- Remove mutiple doc-related files from the snap

### Result

As a result of this work, the amd64 snap size is reduced from 180MB to 160mb (about ~12% size reduction)

### DIFF
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -1348,8 +1348,10 @@ increase_sysctl_parameter() {
 }
 
 use_snap_env() {
-  # Configure LD_LIBRARY_PATH
-  export LD_LIBRARY_PATH="$SNAP_LIBRARY_PATH:$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/ceph:${LD_LIBRARY_PATH:-}"
+  # Configure PATH, LD_LIBRARY_PATH and PYTHONPATH
+  export PATH="$SNAP/usr/bin:$SNAP/bin:$SNAP/usr/sbin:$SNAP/sbin:$REAL_PATH"
+  export LD_LIBRARY_PATH="$SNAP_LIBRARY_PATH:$SNAP/lib:$SNAP/usr/lib:$SNAP/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET:$SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/ceph:${REAL_LD_LIBRARY_PATH:-}"
+  export PYTHONPATH="$SNAP/usr/lib/python3.8:$SNAP/lib/python3.8/site-packages:$SNAP/usr/lib/python3/dist-packages"
 
   # Python configuration
   export PYTHONNOUSERSITE=false
@@ -1370,6 +1372,11 @@ use_snap_env() {
   # Configure XDG_RUNTIME_DIR
   export XDG_RUNTIME_DIR="${SNAP_COMMON}/run"
   mkdir -p "${XDG_RUNTIME_DIR}"
+}
+
+run_host() {
+  # Run a command using the host execution environment
+  PATH="$REAL_PATH" LD_LIBRARY_PATH="$REAL_LD_LIBRARY_PATH" PYTHONPATH="$REAL_PYTHONPATH" "${@}"
 }
 
 # check if this file is run with arguments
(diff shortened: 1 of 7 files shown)


microk8s-resources/actions/common/utils.sh
```diff
@@ -1372,6 +1374,11 @@ use_snap_env() {
   mkdir -p "${XDG_RUNTIME_DIR}"
 }
 
+run_host() {
```
bschimke95, line 1377: Wouldn't this break the snap confinement, as it assumes that the underlying host has the program available?

tests/lxc/install-deps/images_archlinux
```diff
@@ -26,7 +26,7 @@ pip3 install pytest requests pyyaml
 n=0
 until [ $n -ge 7 ]
 do
-  sudo snap install core18 && break  # substitute your command here
+  sudo snap install core20 && break  # substitute your command here
```
bschimke95, line 29: why this version bump here and in other places? Seems unrelated to the cleanup.

microk8s-resources/actions/common/utils.sh
neoaggelos: indeed, this is only used on non-strict code paths

tests/lxc/install-deps/images_archlinux
neoaggelos: this is actually a leftover from our upgrade core18 -> core20. the `core18` installed here has no effect

snap/snapcraft.yaml
```diff
@@ -171,35 +165,38 @@ parts:
       - python3-rados
       - python3-rbd
       - sed
-      - socat
```
ktsakalozos, line 174: socat was introduced to address the following error: https://github.com/canonical/microk8s/pull/94

could you please add some comment showing this is not needed anymore?

microk8s-resources/wrappers/run-kubelite-with-args
```diff
@@ -108,18 +108,14 @@ then
 fi
 
 #UFW configuration
```
ktsakalozos, line 110: I see with this PR we are also changing our UFW approach. We try to find UFW from the host and remove it from the snap. Could we please have this in a separate PR with justification on why this is better?

tests/lxc/install-deps/ubuntu_16.04
```diff
@@ -8,4 +8,4 @@ apt-get install python3-pip docker.io -y
 pip3 install importlib-metadata==2.1.0 pytest requests pyyaml sh
 # Attempting to address https://forum.snapcraft.io/t/lxd-refresh-cause-container-socket-error/8698
 # if core is to be installed by microk8s it fails
-snap install core18 | true
```
ktsakalozos, line 11: It might be worth completely removing this line and seeing if this problem still exists.

snap/snapcraft.yaml
neoaggelos: `socat` is not needed for some time now https://github.com/containerd/cri/pull/1470:

```
root@u1:~# find /snap/microk8s/current | grep socat
root@u1:~# which socat
root@u1:~# k port-forward deploy/nginx 8080:80 &
Forwarding from 127.0.0.1:8080 -> 80
Forwarding from [::1]:8080 -> 80
root@u1:~# curl localhost:8080
Handling connection for 8080
<!DOCTYPE html>
<html>
<head>
<title>Welcome to nginx!</title>
<style>
html { color-scheme: light dark; }
body { width: 35em; margin: 0 auto;
font-family: Tahoma, Verdana, Arial, sans-serif; }
</style>
</head>
<body>
<h1>Welcome to nginx!</h1>
<p>If you see this page, the nginx web server is successfully installed and
working. Further configuration is required.</p>

<p>For online documentation and support please refer to
<a href="http://nginx.org/">nginx.org</a>.<br/>
Commercial support is available at
<a href="http://nginx.com/">nginx.com</a>.</p>

<p><em>Thank you for using nginx.</em></p>
</body>
</html>
```

tests/lxc/install-deps/ubuntu_16.04
neoaggelos: I would rather not do this as part of this PR.

The installed `core18` is only unused (when installing `microk8s`, `core20` will be pulled instead). We should not need this line, but I'm pro keeping it.

microk8s-resources/wrappers/run-kubelite-with-args
neoaggelos: reverted

PR #4273, opened 2023-10-31
Fix microk8s import images for offline clusters

neoaggelos:
### Summary

Fixes https://github.com/canonical/microk8s/issues/4253

### Changes

- Retry listing the cluster nodes until success instead of waiting for a node with status == Ready


ktsakalozos:
LGTM +1

PR #4328, opened 2023-12-04
Fix find-resolv-conf.py from breaking with scoped IPv6 addresses

neoaggelos:
### Summary

References https://github.com/canonical/microk8s/issues/4327

### Changes

Python 3.8 (from core20) does not support scoped IPv6 addresses. Make sure to handle the case gracefully.

### Notes

This should be backported to 1.28, 1.27, 1.26 branches after merging


bschimke95 (approved):
LGTM

PR #4360, opened 2024-01-10
Do not set --hostname-override on kube-proxy when joining a cluster

### LINKED ISSUE #4345: NodePort services with externalTrafficPolicy set to Local don't work on worker nodes
#### Summary
NodePort services with externalTrafficPolicy set to Local don't work on worker nodes, they only work on master nodes.  Please reference bug for another example:  https://github.com/canonical/microk8s/issues/2971
Setting the externalTrafficPolicy to Cluster fixes the iptables issues, however this will not allow for actual sourceip to be accessible.

All nodes, including worker nodes should respond on the nodeport as configured and send traffic to the local pod.

#### Reproduction Steps

install daemonset with nodeport service enabled, with externalTrafficPolicy set to Local.

#### Introspection Report
[inspection-report-20231221_150119.tar.gz](https://github.com/canonical/microk8s/files/13743270/inspection-report-20231221_150119.tar.gz)

You can see various log entries related to the iptables configuration:
 for example: 
-A KUBE-EXTERNAL-SERVICES -p tcp -m comment --comment "istio-ingress-int/istio-internal-ingress-gw:status-port has no local endpoints" -m addrtype --dst-type LOCAL -m tcp --dport 31081 -j DROP

neoaggelos:
### Summary

Closes #4345

Needs backport on all supported tracks

`kube-proxy` must be configured to use the same node name as `kubelet`, otherwise services with Local external traffic policy fail to forward traffic (since node name of endpoints does not match the node name of the proxy).

### Description

In the past (for non HA), we would set `--hostname-override` flags on both kube-proxy and kubelet services. In recent times, we do not need to do this, and we end in a situation where kubelet is running with its proper hostname, but kube-proxy still tries to find a node with the IP as name (which does not exist).

For more details, see https://github.com/canonical/microk8s/issues/4345#issuecomment-1871073150

### Change

- Clear the `--hostname-override` flag on the kube-proxy when joining a cluster. We would only do this if we set this flag for kubelet as well, but we don't.

### Testing

Locally

### DIFF
--- a/scripts/wrappers/join.py
+++ b/scripts/wrappers/join.py
@@ -312,7 +312,7 @@ def create_kubeconfig(token, ca, master_ip, api_port, filename, user):
         try_set_file_permissions(config)
 
 
-def update_kubeproxy(token, ca, master_ip, api_port, hostname_override):
+def update_kubeproxy(token, ca, master_ip, api_port):
     """
     Configure the kube-proxy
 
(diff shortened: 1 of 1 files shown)


bschimke95 (approved):
LGTM

PR #4379, opened 2024-01-25
Retries in the nodes check in setup_old_versioned_cluster

ktsakalozos:
Reduce flakiness in testing of setup_old_versioned_cluster


ktsakalozos:
Tests are passing now.

PR #4469, opened 2024-03-26
Initialise 1.29-strict branch

bschimke95:


.github/workflows/build-snap.yml
```diff
@@ -2,6 +2,7 @@ name: Build and test MicroK8s snap
 
 on:
   - push
+  - pull_request
```
neoaggelos, line 5: just on pull request, otherwise this runs CI jobs twice

neoaggelos (approved):

PR #4473, opened 2024-03-29
fix: don't error if pod has no containerStatuses (#4471)

### LINKED ISSUE #4471: KeyError in kill-host-pods.py
https://github.com/canonical/microk8s/blob/f9eb294558bf1103a3109504452077af059af384/scripts/kill-host-pods.py#L24

It is possible for a pod _not_ to have a `containerStatus` field, e.g. when it is in "Pending" state because no compatible node is available for it to be scheduled onto.  If `kill-host-pods.py` runs when there are one or more pods in this condition, it fails with the following traceback:

```
Traceback (most recent call last):
  File "/snap/microk8s/6364/scripts/kill-host-pods.py", line 104, in <module>
    main()
  File "/snap/microk8s/6364/usr/lib/python3/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/snap/microk8s/6364/usr/lib/python3/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/snap/microk8s/6364/usr/lib/python3/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/snap/microk8s/6364/usr/lib/python3/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/snap/microk8s/6364/scripts/kill-host-pods.py", line 88, in main
    if not post_filter_has_known_containers(pod, containers):
  File "/snap/microk8s/6364/scripts/kill-host-pods.py", line 24, in post_filter_has_known_containers
    for container in pod["status"]["containerStatuses"] or []:
KeyError: 'containerStatuses'
```

Two possible fixes:

1. use `.get` instead of `[]`
2. add another layer of try/except around the whole `for` loop

ianroberts:
#### Summary

`kill-host-pods.py` iterates over all pods in all namespaces looking for pods that are running on the current node.  But if any pods in the iteration do not have `containerStatuses` in their pod status structure (e.g. if they're Pending waiting for a valid node to schedule to) then this currently causes the script to crash with a `KeyError`.  This PR changes the code to treat a missing `containerStatuses` the same as an empty one, so the pod is correctly skipped rather than the whole script crashing.

Closes #4471

#### Changes

- changed `pod["status"]["containerStatuses"]` (which raises an error if `containerStatuses` is missing) to `pod["status"].get("containerStatuses")` (which simply returns `None`).

#### Testing

Tested locally with a copy of the script, I believe the change is trivial enough not to require automated tests beyond this.

#### Checklist

### DIFF
--- a/scripts/kill-host-pods.py
+++ b/scripts/kill-host-pods.py
@@ -21,7 +21,7 @@ def post_filter_has_known_containers(pod, containers: list) -> bool:
     Return true if any of the container IDs on the pod match the list of
     containers passed on the second argument.
     """
-    for container in pod["status"]["containerStatuses"] or []:
+    for container in pod["status"].get("containerStatuses") or []:
         try:
             _, container_id = container["containerID"].split("containerd://", 2)
             if container_id in containers:


neoaggelos (approved):
LGTM, thanks a lot for the catch @ianroberts !

neoaggelos:
Would you also mind signing the CLA? And I believe we should also backport this change to older branches as well

ianroberts:
> Would you also mind signing the CLA? And I believe we should also backport this change to older branches as well

I did just sign the CLA, though the email address I've used is a secondary email on my launchpad account (`iroberts-usfd`) rather than the preferred one.

neoaggelos:
The CLA check fails with:

```
Checking the following users on Launchpad:
- ian@$DOMAIN ✕ (has no Launchpad account)
```

Can you amend the commit and use the email that you signed the CLA with?

ianroberts:
> The CLA check fails with:
> 
> ```
> Checking the following users on Launchpad:
> - ian@$DOMAIN ✕ (has no Launchpad account)
> ```
> 
> Can you amend the commit and use the email that you signed the CLA with?

That's what I mean - the email I signed the CLA with _is_ `ian@$DOMAIN`, and that email address _is_ associated with my launchpad account, but it was a secondary rather than primary email address.

I've just edited my launchpad profile to swap the primary and secondary addresses, can you run the CLA checker again now?

ianroberts:
Sorry, my bad, I had added the address to my _Ubuntu one_ account but not directly to my _Launchpad_ account, it should be there now.  But I'm not showing as a member of https://launchpad.net/~contributor-agreement-canonical, maybe I need to fill in the CLA again now the right emails are all linked?

neoaggelos:
Yes, can you please sign the CLA with the email address used in the commit? Thank you!

ianroberts:
I already had, twice... just filled in the form again, let's see if it takes this time.

ianroberts:
This bug is still present - I've just tried again to sign the CLA, can you see whether it has succeeded this time?

jmrr:
I'd like to chip in with some collateral damage this issue has produced in our setup.

We've had a major outage in production and almost another one (we caught this early and knew how to intervene), due to:

1. A snap refresh caused the creation of `"${SNAP_DATA}/var/lock/snapdata-mounts-need-reload"` flag file to trigger a `calico-node` DaemonSet restart
2. Lack of `containerStatuses` key caused the `KeyError` exception in the `kill-host-pods.py` script described in this PR.
3. Script fails with error, preventing `microk8s-resources/default-hooks/reconcile.d/10-pods-restart` outer script from removing the `"${SNAP_DATA}/var/lock/snapdata-mounts-need-reload"` file
4. `dqlite` was flooded with events due to the `calico-node` pods restart loop
5. At some point, `dqlite` couldn't cope, corrupted, and the whole cluster performance degraded.

ianroberts:
Rebased on latest master

ianroberts:
I say only "kind of" superseded, because the _default_ hook scripts in 1.32 have been changed to only look for pods on the current node, but these are only the default scripts that get installed if there are no hooks already present in `/var/snap/microk8s/common` from a previous version.  So it's only truly fixed in a clean install of 1.32; for users who want to upgrade from an earlier version it is necessary to manually copy in the updated hooks from `/snap/microk8s/current` to `/var/snap/microk8s/common` in order to benefit from the fix.

scripts/kill-host-pods.py
```diff
@@ -21,7 +21,7 @@ def post_filter_has_known_containers(pod, containers: list) -> bool:
     Return true if any of the container IDs on the pod match the list of
     containers passed on the second argument.
     """
-    for container in pod["status"]["containerStatuses"] or []:
+    for container in pod["status"].get("containerStatuses") or []:
```
HomayoonAlimohammadi, line 24: nit: Maybe we can change this to:
```suggestion
    for container in pod["status"].get("containerStatuses", []):
```

ianroberts: Can `containerStatuses` ever be present-but-null in the YAML, or is it always either present-and-valid or not present at all?

I wrote this as `.get("containerStatuses") or []` because that will handle both cases (missing or present-but-null) and treat them as an empty array.  Using `.get("containerStatuses", [])` will give you an empty array if `containerStatuses` is missing, but will return `None` if it is set-but-null (and in turn raise an error when you try to iterate over a `None` value).

HomayoonAlimohammadi:
@ianroberts That's right, sorry I committed the change since your response was not loaded for some reason.
You're right, your approach catches both. AFAIK that shouldn't ever happen, but it's a nice defensive programming style.

HomayoonAlimohammadi:
I believe the failing addon tests are not related to this PR. Should we merge this one or wait for a fix to rebase this PR on top of it @louiseschmidtgen?

louiseschmidtgen:
@HomayoonAlimohammadi The issue is resolved can you re-run the workflow please?

Issue #4361, opened 2024-01-10
Microk8s v1.29 snap installation failed on plain Debian 12.4


TecIntelli:
#### Summary
The last days I noticed that the installation of MicroK8s v1.29/stable (6364) failed on a new (plain) Debian 12.4 system (tested on AWS EC2 with default Debian 12 image provided by AWS). After a few tests I can summarize the following behavior:

- Installation of MicroK8s v1.28/stable (6089) on described Debian system via snap works like expected
[microk8s-1.28_6089-inspection-report-20240110_103728.tar.gz](https://github.com/canonical/microk8s/files/13886670/microk8s-1.28_6089-inspection-report-20240110_103728.tar.gz)

- Installation of MicroK8s v1.29/stable (6364) on described Debian system via snap failed and `microk8s inspect` responsed with: 
```
admin@ip-172-31-16-112:~$ microk8s status
microk8s is not running. Use microk8s inspect for a deeper inspection.
admin@ip-172-31-16-112:~$ microk8s inspect
Inspecting system
Inspecting Certificates
Inspecting services
  Service snap.microk8s.daemon-cluster-agent is running
  Service snap.microk8s.daemon-containerd is running
  Service snap.microk8s.daemon-kubelite is running
  Service snap.microk8s.daemon-k8s-dqlite is running
  Service snap.microk8s.daemon-apiserver-kicker is running
  Copy service arguments to the final report tarball
Inspecting AppArmor configuration
Gathering system information
  Copy processes list to the final report tarball
  Copy disk usage information to the final report tarball
  Copy memory usage information to the final report tarball
  Copy server uptime to the final report tarball
  Copy openSSL information to the final report tarball
  Copy snap list to the final report tarball
  Copy VM name (or none) to the final report tarball
  Copy current linux distribution to the final report tarball
  Copy asnycio usage and limits to the final report tarball
  Copy inotify max_user_instances and max_user_watches to the final report tarball
  Copy network configuration to the final report tarball
Inspecting kubernetes cluster
  Inspect kubernetes cluster
sudo: unable to resolve host ip-172-31-16-112: Name or service not known
sudo: unable to resolve host ip-172-31-16-112: Name or service not known
sudo: unable to resolve host ip-172-31-16-112: Name or service not known
sudo: unable to resolve host ip-172-31-16-112: Name or service not known
sudo: unable to resolve host ip-172-31-16-112: Name or service not known
sudo: unable to resolve host ip-172-31-16-112: Name or service not known
[...]

PR #4503, opened 2024-04-18
Add systemd delegate config if cpuset is missing

berkayoz:
Add delegate config if cpuset is missing, possible workaround for #4361
The issue is likely to be upstream dependency update on runc/libcontainers as 1.28 does not suffer from this.
An updated kernel(6.1+, tested with 6.5) also does not seem to suffer from this condition.

### DIFF
--- a/microk8s-resources/wrappers/run-kubelite-with-args
+++ b/microk8s-resources/wrappers/run-kubelite-with-args
@@ -125,6 +125,17 @@ then
   fi
 fi
 
+if ! is_strict && (systemctl show snap.microk8s.daemon-kubelite.service -p NRestarts | grep -qv "NRestarts=0") && (grep -qv cpuset /sys/fs/cgroup/cgroup.subtree_control)
+then
+  mkdir -p /etc/systemd/system/snap.microk8s.daemon-kubelite.service.d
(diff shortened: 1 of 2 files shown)


neoaggelos (approved):
LGTM. Tis a terrible hack, but it should work

microk8s-resources/wrappers/run-kubelite-with-args
```diff
@@ -125,6 +125,20 @@ then
...
+   (grep -qv cpuset /sys/fs/cgroup/cgroup.subtree_control) &&
+   [ ! -e /etc/systemd/system/snap.microk8s.daemon-kubelite.service.d/delegate.conf ]
+then
+  mkdir -p /etc/systemd/system/snap.microk8s.daemon-kubelite.service.d
+  tee /etc/systemd/system/snap.microk8s.daemon-kubelite.service.d/delegate.conf > /dev/null <<EOF
+[Service]
+Delegate=yes
+EOF
+  systemctl daemon-reload
+  snap restart microk8s
```
neoaggelos, line 139: is there any situation where these might fail and cause the kubelite process to start crash-looping?

berkayoz: Good point, not sure so let me add ` || true` to make sure we just skip on failures here.

PR #4601, opened 2024-08-05
Do not run addons tests on the host. Only on lxc containers.

ktsakalozos:
#### Summary
Running tests on the host and on lxc containers seems to have unforeseen side-effects. For now we disable testing on the host.

#### Changes
Remove the addon tests that run on the host on strict confinement.

#### Testing
Locally and in our CI

#### Possible Regressions
The most notable change is the test of micro-ceph


bschimke95 (approved):
LGTM

PR #4606, opened 2024-08-05
Disable testing on debian and centos. Images missing.

ktsakalozos:
The lxc images are missing for the spread tests.


bschimke95 (approved):
LGTM

PR #4611, opened 2024-08-06
Update the kubelite patch for 1.31

ktsakalozos:
#### Summary
The kubelite patch we had from 1.28 does not apply cleanly on 1.31.


bschimke95 (approved):
LGTM

PR #4705, opened 2024-10-16
fix: ensure nf_conntrack module loaded for kubelite.

### LINKED ISSUE #4462: Worker nodes fail to start after reboot as nf_conntrack kernel module not loaded
#### Summary
I added a node as a worker to a small cluster that had GPU enabled. It works fine initially, but on rebooting the node the `microk8s.daemon-kubelite.service` fails to start as it is unable to open `/proc/sys/net/netfilter/nf_conntrack_max` :
```
microk8s.daemon-kubelite[2119]: E0314 18:28:26.546586    2119 server.go:537] "Error running ProxyServer" err="open /proc/sys/net/netfilter/nf_conntrack_max: no such file or directory"
``` 

Adding `nf_conntrack` to the end of `/etc/modules-load.d/modules.conf` in the worker node VM works around the issue.

Nodes are all running Ubuntu Server 22.04.4 and microk8s v1.28.7 from snap. They run as VMs in a Proxmox cluster.

Node should come up into Ready status after it is rebooted

#### Reproduction Steps
1. Single node with GPU enabled, but no GPU h/w
2. Add a GPU node as a worker to the non-GPU node
3. Verify all nodes are ready and cluster is functional
4. Reboot the GPU node and wait for it to return to Ready status

#### Introspection Report
Please contact me directly if this is required.

#### Can you suggest a fix?
Ensure the nf_conntrack module is loaded on worker nodes as it is on full nodes.

#### Are you interested in contributing with a fix?
Not sure where to fix this issue properly.

aznashwan:
#### Summary

This patch ensures that the `nf_conntrack` kernel module is loaded before `kubelite` is started as the ProxyServer needs to read some conntrack module-related params from procfs.

Closes #4462 

#### Changes

By explicitly loading `nf_conntrack` before starting `kubelite`, it should ensure the procfs values that the ProxyServer reads are always present on startup.

Previously, although the it would always crashed if the module wasn't loaded, this wasn't that common of an occurrence in practice as there are quite a few ways `nf_conntrack` gets loaded transparently:
* Cilium [automatically loads `iptable_nat`](https://github.com/cilium/cilium/blob/63cd391f93b4e2c865268241d384504348672042/pkg/datapath/iptables/iptables.go#L367-L368) after a small startup delay, whose dependency tree includes `nf_conntrack`
* starting firewalld/ufw/most other firewall services
* setting iptables/nftables rules which imply session tracking

#### Testing

I've written a [trivial eBPF probe](https://gist.github.com/aznashwan/32b8b00f0378e341ea29f80d0a4eca44) to monitor whenever a process loads a kernel module, and validated that:
1) after freshly booting the system (22.04 and 24.04 VMs on Proxmox KVM FWIW), nothing had/was automatically loading the `nf_conntrack` module
2) after installing the updated microk8s snap, the `run-kubelite-with-args` script would load it itself
3) there were no errors/warnings related to values exposed by the conntack module in the kubelite logs
4) ensured a simple test `Deployment` worked as expected

#### Possible Regressions

Only the minute risk that the `nf_conntrack` module (which is *very* widely distributed and used) might not be installed on the host system.

#### Checklist

* [X] The introduced changes are covered by unit and/or integration tests. ***

*** it's technically implicitly covered since the LXD profile used for integration tests includes `nf_nat`, which depends on `nf_conntrack` and thus cause it to be loaded, but there is no "negative test" where we intentionally deny its loading.

### DIFF
--- a/microk8s-resources/wrappers/run-kubelite-with-args
+++ b/microk8s-resources/wrappers/run-kubelite-with-args
@@ -204,6 +204,19 @@ then
   fi
 fi
 
+# kube-proxy reads some values related to the 'nf_conntrack' kernel
+# module from procfs on startup, so we must ensure it is loaded:
+if ! [ -f /proc/sys/net/netfilter/nf_conntrack_max ]
+then
+  if /sbin/modprobe nf_conntrack || modprobe nf_conntrack
+  then
+    echo "Successfully loaded nf_conntrack module."
+  else
+    echo -n "Failed to load nf_conntrack kernel module. "
+    echo "ProxyServer will fail to start until it's loaded."
+  fi
+fi
+
 # on lxc containers do not try to change the conntrack configuration
 # see https://github.com/canonical/microk8s/issues/1438
 if grep -E lxc /proc/1/environ &&


berkayoz (approved):
LGTM

PR #4773, opened 2024-12-04
Disable WatchList feature gate

berkayoz:
Disable WatchList feature gate(enabled by default 1.32+) until k8s-dqlite can properly support it.

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -385,6 +385,13 @@ then
   need_api_restart=true
 fi
 
+# Disable the watchlist feature until k8s-dqlite can handle it
+if ! grep -e "feature-gates=WatchList=false" ${SNAP_DATA}/args/kube-apiserver
+then
+  echo '--feature-gates=WatchList=false' >> ${SNAP_DATA}/args/kube-apiserver
+  need_api_restart=true
+fi
+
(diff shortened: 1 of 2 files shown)


snap/hooks/configure
```diff
@@ -385,6 +385,13 @@ then
   need_api_restart=true
 fi
 
+# Disable the watchlist feature until k8s-dqlite can handle it
+if ! grep -e "feature-gates=WatchList=false" ${SNAP_DATA}/args/kube-apiserver
```
ktsakalozos, line 389: is this enough? What if some one has set `feature-gates=AnotherFeature=True,WatchList=false`? Or what if someone has `WatchList=true` but is running etcd?

I think we should detect the absence of `WatchList` and make sure it is set to `false`.

snap/hooks/configure
```diff
@@ -385,6 +385,13 @@ then
...
+then
+  echo '--feature-gates=WatchList=false' >> ${SNAP_DATA}/args/kube-apiserver
```
ktsakalozos, line 391: What if the user already has `--feature-gates` set? We would be overwriting his options.

ktsakalozos:
LGTM +1 thank you @berkayoz

PR #4781, opened 2024-12-11
Bump go version to 1.23

berkayoz:
Bumps go version to 1.23 to accommodate 1.32 release
Bumps etcd patch version to `3.5.17` to build with 1.23

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -22,7 +22,7 @@ parts:
   build-deps:
     plugin: nil
     build-snaps:
-      - go/1.21/stable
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -22,7 +22,7 @@ parts:
   build-deps:
     plugin: nil
     build-snaps:
-      - go/1.21/stable
+      - go/1.23/stable
```
addyess, line 25: big jump.  Let's go!

addyess (approved):
🍰 🚀 🍨

PR #4791, opened 2024-12-17
Update Images Workflow Improvement

louiseschmidtgen:
## Update Images Workflow Improvement
This PR accomplishes the following:
- replaces @neoaggelos with @berkayoz  as a reviewer.
- adds top-level permissions
- adds 1.31 branches

### DIFF
--- a/.github/workflows/update-images.yml
+++ b/.github/workflows/update-images.yml
@@ -9,7 +9,8 @@ on:
       - "KU-2335/workflow-reviewers" # TODO rm me
 
 permissions:
-  contents: read
+  contents: write
+  pull-requests: write
 
 jobs:
   update:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-images.yml
```diff
@@ -14,11 +18,13 @@ jobs:
           # Latest branches
           - { branch: master, channel: latest/edge }
           # Stable branches
+          - { branch: "1.31", channel: "1.31" }
```
bschimke95, line 21: I guess we should also add 1.32?

louiseschmidtgen: Yes, once we have release branches.

bschimke95: Ah, my mind was stuck in the k8s-snap universe. Ignore my comment...

PR #4819, opened 2025-01-10
kill-host-pods.py: filter pods by node

petrutlucian94:
Starting with k8s 1.32, AuthorizeNodeWithSelectors is enabled by default: https://kubernetes.io/docs/reference/access-authn-authz/node/

If the rbac microk8s addon is enabled, the kube-apiserver will run with "--authorization-mode=RBAC,Node". This means that kublets (system:node:$node) will no longer be allowed to access pods that reside on other nodes.

For this reason, the "kill-host-pods.py" script is now getting access denied errors:

```
  Error from server (Forbidden): pods is forbidden:
  User "system:node:myhostname" cannot list resource "pods" in API group ""
  at the cluster scope: can only list/watch pods with spec.nodeName field selector
```

As suggested by the error message, we'll solve it by filtering pods by the node name.

Fixes: https://github.com/canonical/microk8s/issues/4802

#### Checklist

* [ ] Read the [contributions](https://github.com/canonical/microk8s/blob/master/CONTRIBUTING.md) page.

### DIFF
--- a/microk8s-resources/default-hooks/reconcile.d/10-pods-restart
+++ b/microk8s-resources/default-hooks/reconcile.d/10-pods-restart
@@ -4,7 +4,9 @@
 
 if ! [ -e "${SNAP_DATA}/var/lock/no-cni-reload" ] &&
   [ -e "${SNAP_DATA}/var/lock/snapdata-mounts-need-reload" ]; then
-  if (is_apiserver_ready) && "${SNAP}/scripts/kill-host-pods.py" --with-snap-data-mounts --with-owner -- -A; then
+  if (is_apiserver_ready) && "${SNAP}/scripts/kill-host-pods.py" \
+      --with-snap-data-mounts --with-owner \
+      -- -A --field-selector spec.nodeName=$(hostname) ; then
     rm "${SNAP_DATA}/var/lock/snapdata-mounts-need-reload"
   fi
 fi


berkayoz (commented):
LGTM overall, I'll run some manual tests and merge if all is well

petrutlucian94:
> LGTM overall, I'll run some manual tests and merge if all is well

Thanks! fwiw, I used the following to trigger a pod cleanup:

```
sudo touch /var/snap/microk8s/current/var/lock/snapdata-mounts-need-reload
```

berkayoz (approved):
LGTM

ianroberts:
I just ran into the linked issue on upgrading to 1.32 - this PR is a sensible fix, but I wasn't aware that it is necessary to _manually_ sync `/snap/microk8s/current/default-hooks` over to `/var/snap/microk8s/common/hooks` after upgrading.  If you don't do that then you end up still using the `10-pods-restart` from an earlier revision, and you get the same error message.

`snap/hooks/configure` probably needs to be updated to copy in the new `10-pods-restart` hook on upgrade if required, or else it should be documented somewhere that you have to manually sync any hook changes on every snap refresh.

petrutlucian94:
@ianroberts Thanks for bringing this up. It seems to be by design, the default-hooks are copied over only for fresh installations, probably so that it won't override custom hooks:

https://github.com/canonical/microk8s/blob/636c31336997bd0b2a72c26e063cee8f62dcd4ce/snap/hooks/configure#L223-L227

We have the following options:

* document this
* always copy the default hooks, even if this might override user defined hooks
  * the team suggested to only copy `10-pods-restart` as part of `snap/hooks/configure`
    * I'll open a PR to keep the conversation going

PR #4831, opened 2025-01-21
scripts: Fix kill-host-pods script

HomayoonAlimohammadi:
### Overview
This PR fixes the `kill-host-pods.py` script. Currently the `--with-snap-data-mounts` filter does not work correctly as it returns `False` too early.


HomayoonAlimohammadi:
To be continued in https://github.com/canonical/microk8s/pull/4835

PR #4913, opened 2025-03-02
CI: Distro from env variable in test - 1.30

ktsakalozos:
We would like to switch to use Ubuntu 20.04 in our tests. This PR fixes some tests that had the ubuntu distro hard coded.


louiseschmidtgen (approved):
LGTM, thank you!

PR #4963, opened 2025-03-25
fix: flush the `CNI-HOSTPORT-DNAT` chain on stop

berkayoz:
`microk8s stop` was changed to fully remove containers via ctr(#4755) which results in an old DNAT rule getting left-over that points to the old container id. flushing the chain should fix issues of hostPort connections dropping after a stop & start.

Fixes #4944

PR #4755, opened 2024-11-25
Cleans up containers and shims on microk8s stop and snap removal

claudiubelu:
#### Summary

The containerd shims and the containers are now removed in both classic and strict modes when stopping microk8s (``microk8s stop``) or when removing the snap.

Note that ``ctr container delete`` has no ``--force`` flag.

Adds additional assertion in the upgrade test, ensuring that the Pods / containers have been removed on snap removal.

Adds additional test, ensuring that the Pods / containers have been removed when running ``microk8s stop``.

#### Changes

The Kubernetes-related containers / Pods are now stopped when running ``microk8s stop``. The same containers / Pods are no longer leaked on snap removal.

#### Testing

Added extra assertions and integration test to ensure the new behaviour.

#### Checklist

* [x] Read the [contributions](https://github.com/canonical/microk8s/blob/master/CONTRIBUTING.md) page.
* [x] Submitted the [CLA form](https://ubuntu.com/legal/contributors/agreement), if you are a first time contributor.
* [x] The introduced changes are covered by unit and/or integration tests.

#### Notes

Inspired by: https://github.com/canonical/microk8s/pull/4693
Closes: https://github.com/canonical/microk8s/issues/3969

PR #4966, opened 2025-03-26
[Backport 1.32] fix: flush the `CNI-HOSTPORT-DNAT` chain on stop (#4963)

berkayoz:
`microk8s stop` was changed to fully remove containers via ctr(#4755) which results in an old DNAT rule getting left-over that points to the old container id. flushing the chain should fix issues of hostPort connections dropping after a stop & start.

(cherry picked from commit b170cad76d353570775301fe24e4124511b65108)


HomayoonAlimohammadi (approved):
Thanks a lot @berkayoz! LGTM

PR #5008, opened 2025-04-03
ci: unblock the CI and add a few improvements

petrutlucian94:
* GH actions no longer support 20.04 runners, so we'll switch to 22.04
  * need to apply the well known iptables workaround to address the compatibility issue between LXD and Docker
* we're adding the run-all-tests label which controls the UNDER_TIME_PRESSURE env variable
* to avoid duplicate logic, we'll add a composite action that takes care of the test prerequisites, making them easier to maintain.
* we'll avoid installing python packages globally (no longer allowed by recent Python versions). Instead, we'll use actions/setup-python, which will also cache the packages.
* define job timeouts

### DIFF
--- a/.github/actions/test-prep/action.yaml
+++ b/.github/actions/test-prep/action.yaml
@@ -12,8 +12,6 @@ runs:
       run: |
         set -x
         sudo pip3 install -R ./tests/requirements.txt
-        sudo apt-get -y install open-iscsi
-        sudo systemctl enable iscsid
     # Docker sets iptables rules that interfere with LXD and K8s.
     # https://documentation.ubuntu.com/lxd/en/latest/howto/network_bridge_firewalld/#prevent-connectivity-issues-with-lxd-and-docker
     - name: Apply Docker iptables workaround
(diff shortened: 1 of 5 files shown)


.github/actions/test-prep/action.yaml
```diff
@@ -0,0 +1,24 @@
...
+  steps:
+    - name: Setup Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.10"
+    - name: Install test dependencies
+      shell: bash
+      run: |
+        set -x
+        sudo pip3 install -R ./tests/requirements.txt
```
louiseschmidtgen, line 14: Do we need to pin pytest in this requirements.txt?

louiseschmidtgen (approved):
Nice work, one little comment:

.github/actions/test-prep/action.yaml
petrutlucian94: probably not, I don't think we have specific pytest requirements. as long as all the methods that we use are there, it should be ok.

petrutlucian94:
@louiseschmidtgen turns out that `test-addons.py::TestAddons::test_cis` uses rook and the test fails without `iscsid`. I assume that the test uses the Ceph iSCSI gateway, I'll go ahead and add it back.

PR #5021, opened 2025-04-11
fix: run iptables as sudo on microk8s stop

berkayoz:
Runs `iptables` with sudo and suppress the output.


louiseschmidtgen (approved):
LGTM

PR #5048, opened 2025-04-18
chore: add updated patches for kubernetes v1.33

berkayoz:
adds updated patches for kubernetes v1.33


HomayoonAlimohammadi (approved):
Looks like we're adjusting the patches according to the upstream Kubernetes changes. Nothing else is changed regarding the patches themselves. LGTM, thanks a lot @berkayoz!

PR #5155, opened 2025-07-28
chore: add Kubernetes v1.34.0 patches

rapour:
This PR adds patches needed for Kubernetes v1.34.0


ktsakalozos:
The fix for the failing test will be addressed in a follow-up PR.

PR #5196, opened 2025-08-25
Update the dashboard validator so it passes in both old and new versions

ktsakalozos:
#### Summary
We updated the dashboard addon so it is in its own namespace. The upgrade tests should take this into account.

### DIFF
--- a/tests/validators.py
+++ b/tests/validators.py
@@ -42,9 +42,7 @@ def validate_dns_dashboard():
     attempt = 30
     while attempt > 0:
         try:
-            output = kubectl(
-                f"get --raw /api/v1/namespaces/{ns}/services/https:{service}/proxy/"
-            )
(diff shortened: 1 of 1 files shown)


tests/validators.py
```diff
@@ -23,21 +23,26 @@ def validate_dns_dashboard():
     Validate the dashboard addon by trying to access the kubernetes dashboard.
     The dashboard will return an HTML indicating that it is up and running.
     """
-    ns = "kubernetes-dashboard"
-    components = ["api", "auth", "metrics-scraper", "web"]
-    app_names = [f"kubernetes-dashboard-{app}" for app in components]
-    app_names.append("kong")
+    service = "kubernetes-dashboard:"
```
rapour, line 26: Where do we set the service port in case the if statement doesn't check?

PR #5199, opened 2025-08-25
chore: update base to core22

rapour:
Following changes are introduced:
- Update `base` to `core22`
- Update python version to `3.10`
- Update env names based on the [official documentation](https://documentation.ubuntu.com/snapcraft/stable/how-to/change-bases/change-from-core20-to-core22/)

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -16,7 +16,7 @@ environment:
   REAL_PATH: $PATH
   REAL_LD_LIBRARY_PATH: $LD_LIBRARY_PATH
   REAL_PYTHONPATH: $PYTHONPATH
-  SNAPCRAFT_ARCH_TRIPLET: $SNAPCRAFT_ARCH_TRIPLET
+  SNAPCRAFT_ARCH_TRIPLET: $CRAFT_ARCH_TRIPLET
 
 parts:
   build-deps:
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -92,31 +100,34 @@ parts:
     after: [build-deps]
     plugin: nil
     source: build-scripts/components/helm
-    override-build: $SNAPCRAFT_PROJECT_DIR/build-scripts/build-component.sh helm
+    override-build: $CRAFT_PROJECT_DIR/build-scripts/build-component.sh helm
```
berkayoz, line 103: Do we need `build-attributes: [enable-patchelf]` here for helm as well?

snap/snapcraft.yaml
```diff
@@ -202,92 +214,94 @@ parts:
...
       - python3-pkg-resources
-      - python3.8-minimal
+      - python3.10-minimal
       - python3-openssl
       - python3-requests
       - python3-click
       - python3-dateutil
     stage:
-      - -usr/lib/python3.9
+      - -usr/lib/python3.11
```
berkayoz, line 304: I can't recall the intention of adding excluding these from staging, do we have `usr/lib/python3.11` in core22?

berkayoz (commented):
LGTM overall, left some comments. We should check our jenkins repo as well to update any `core20` references if there are any.

snap/snapcraft.yaml
rapour: I was under the impression that helm is built statically and CGO_ENABLED is set to 0 so we don't need to patch the elf headers for it. Do you think we need to look into it further?

snap/snapcraft.yaml
rapour: I double checked by building and installing the snap locally and the directory for 3.11 was there but I just followed the previous pattern. Shall we remove the line?

snap/snapcraft.yaml
berkayoz: We should be fine, thanks for double checking.

berkayoz (approved):
LGTM

PR #5315, opened 2025-11-25
backport: update security scanner

ethandcosta:
The old security scanner process used in 1.31 and prior fails because the process through which it uploads SARIF results is deprecated. This PR updates the CI to use the security scanning process used in main.


ethandcosta:
Ignore the failing checks related to community/core addons, these are being addressed in parallel PRs.

PR #5342, opened 2025-12-16
fix: add wait tolerance for unit test

ethandcosta:
This PR fixes a race condition in test_simple.py where the test fails because it tries to check for coredns before microk8s is finished getting ready.


HomayoonAlimohammadi:
The failing tests will pass after https://github.com/canonical/microk8s/pull/5339

louiseschmidtgen:
@copilot rebase

PR #5366, opened 2026-01-08
tests: check CoreDNS via pod state instead of host process

Santa0337:
#### Summary
The stop/start test was checking for a `coredns` host process using `psutil`.
In modern MicroK8s, CoreDNS runs as a containerized pod and is not visible as a
host process even when DNS is healthy. This caused the test to fail even though
CoreDNS was running correctly.

#### Changes
- Added a helper to verify CoreDNS availability via Kubernetes pod state.
- Updated the stop/start test to use Kubernetes pod status instead of checking
  for a host-level `coredns` process.

#### Testing
- Ran `pytest tests/test-simple.py` locally.
- Verified that the test now passes when CoreDNS is running as a pod.
- Confirmed CoreDNS status using `microk8s kubectl get pods -n kube-system`.

#### Possible Regressions
This change should not introduce regressions. The test now relies on Kubernetes
API state, which is the correct and stable source of truth for CoreDNS health.

#### Checklist
* [x] Read the contributions page.
* [x] Submitted the CLA form (first-time contributor).
* [x] The introduced changes are covered by tests.

#### Notes
This change improves test reliability across environments where CoreDNS runs
inside containerd and is not visible as a host process due to snap confinement.

### DIFF
--- a/tests/test-simple.py
+++ b/tests/test-simple.py
@@ -2,7 +2,6 @@ import subprocess
 import time
 import requests
 import os.path
-
 import utils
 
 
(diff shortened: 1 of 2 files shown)


Santa0337 (commented):
@lazzarello, @ktsakalozos could you make a code review.

PR #5501, opened 2026-05-19
fix(strict): remove duplicate kubernetes-support sub-profile causing AppArmor failure

louiseschmidtgen:
## Problem

The strict snap fails to install on Ubuntu 24.04 (including GitHub Actions runners) with:

```
cannot setup profiles for snap "microk8s": cannot load apparmor profiles: exit status 1
apparmor_parser output:
Multiple definitions for hat systemd_run in profile (null) exist, bailing out.
```

This blocks all CI tests for `latest/edge/strict` in the [microk8s-core-addons repo](https://github.com/canonical/microk8s-core-addons/actions/runs/25422501572/job/76690142267).

## Root Cause

`daemon-containerd` and `daemon-apiserver-kicker` each plug both:
- `k8s-journald` → `kubernetes-support` interface ([flavor: `autobind-unix`](https://github.com/canonical/snapd/blame/master/interfaces/builtin/kubernetes_support.go#L336))
- `kubernetes-support` → `kubernetes-support` interface (default flavor)

Both connections are of the **same interface type** (`kubernetes-support`). snapd generates a `^systemd_run` child profile (hat) for each connection. On older AppArmor parsers these duplicate hats were silently merged, but AppArmor 4.x (shipped with Ubuntu 24.04) rejects them outright.

## Fix

Remove `k8s-journald` from the two affected apps. The bare `kubernetes-support` plug (default flavor) is a superset that already includes the `autobind-unix` permissions, so no functionality is lost.

## Testing

Tested manually in an LXD container running Ubuntu 24.04 (Noble).

### Reproducing the bug (before fix)

```console
$ snap install microk8s --channel=latest/edge/strict
error: cannot perform the following tasks:
- Setup snap "microk8s" (8894) security profiles for auto-connections (cannot setup profiles for snap "microk8s": cannot load apparmor profiles: exit status 1
apparmor_parser output:
Multiple definitions for hat systemd_run in profile (null) exist,bailing out.
Multiple definitions for hat systemd_run in profile (null) exist,bailing out.
...
)
```

### Verifying the fix

Patched the downloaded `latest/edge/strict` snap by removing `k8s-journald` from `daemon-containerd` and `daemon-apiserver-kicker` in `meta/snap.yaml`, then repacked with `snap pack`:

```console
$ snap install /root/microk8s-fixed.snap --dangerous
microk8s v1.36.0 installed

$ snap connect microk8s:docker-privileged
$ snap connect microk8s:kubernetes-support
[...]

PR #5502, opened 2026-05-19
fix(strict): assign k8s-kubeproxy to daemon-apiserver-proxy to prevent AppArmor 4.x duplicate hat

louiseschmidtgen:
## Problem

On Ubuntu 24.04+ with AppArmor 4.x, connecting the `k8s-kubeproxy` interface fails with:

```
Multiple definitions for hat systemd_run in profile (null) exist, bailing out.
```

## Root Cause

The `k8s-kubeproxy` plug (snapcraft.yaml)
```
  k8s-kubeproxy:
    interface: kubernetes-support
    flavor: kubeproxy
```
is defined in the top-level `plugs:` section but never explicitly assigned to any app. When a plug is unassigned, snapd implicitly assigns it to **all** apps. This means every app that already has the `kubernetes-support` plug (which generates a `^systemd_run` hat) also gets `k8s-kubeproxy` (which generates its own identical `^systemd_run` hat) — resulting in duplicate hat definitions that AppArmor 4.x rejects.

Older AppArmor parsers (< 4.x) silently merged duplicate hats, masking this bug.

## Fix

Explicitly assign `k8s-kubeproxy` to `daemon-apiserver-proxy`, which does **not** have the `kubernetes-support` plug and therefore has no conflicting hat. This prevents the implicit assignment to all apps.

## Testing

Tested on Ubuntu 24.04 LXD container (AppArmor 4.x):

```
$ snap install microk8s_fixed.snap --dangerous
microk8s v1.36.0 installed

$ snap connect microk8s:kubernetes-support && echo OK
kubernetes-support OK

$ snap connect microk8s:k8s-journald && echo OK
k8s-journald OK

$ snap connect microk8s:k8s-kubelet && echo OK
k8s-kubelet OK

$ snap connect microk8s:k8s-kubeproxy && echo OK
k8s-kubeproxy OK
```

All four `kubernetes-support` interface variants connect successfully.

## Related

- Supersedes #5501 (which only fixed `k8s-journald` duplicates but missed `k8s-kubeproxy`)
- Fixes the AppArmor failure blocking CI on microk8s-core-addons

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -423,6 +423,7 @@ apps:
       - network-bind
       - network-observe
       - network-control
+      - k8s-kubeproxy
   daemon-cluster-agent:
     command: run-cluster-agent-with-args
     daemon: simple


ktsakalozos (approved):
LGTM +1

PR #5537, opened 2026-06-06
fix: add keyUsage extension to CA certificates (#4864)

### LINKED ISSUE #4864: root CA certificate does not include key usage extension
#### Summary

The root CA certificate that is generated does not have the keyUsage defined (see https://github.com/canonical/microk8s/blob/master/microk8s-resources/actions/common/utils.sh#L689)

an example CA cert extension has:

        X509v3 extensions:
            X509v3 Subject Key Identifier: 
                0E:EF:10:1C:40:F6:85:87:76:23:A4:40:C7:D7:73:41:AB:F4:9E:A8
            X509v3 Authority Key Identifier: 
                0E:EF:10:1C:40:F6:85:87:76:23:A4:40:C7:D7:73:41:AB:F4:9E:A8
            X509v3 Basic Constraints: critical
                CA:TRUE

Python 3.13 now has:

Changed in version 3.13: The context now uses [VERIFY_X509_PARTIAL_CHAIN](https://docs.python.org/3/library/ssl.html#ssl.VERIFY_X509_PARTIAL_CHAIN) and [VERIFY_X509_STRICT](https://docs.python.org/3/library/ssl.html#ssl.VERIFY_X509_STRICT) in its default verify flags.

An example error, when running kopf via python 3.13, sees:

[2025-02-11 21:16:16,609] kopf._core.reactor.o [ERROR   ] Request attempt #9/9 failed; escalating: GET https://10.152.183.1:443/api -> ClientConnectorCertificateError(ConnectionKey(host='10.152.183.1', port=443, is_ssl=True, ssl=True, proxy=None, proxy_auth=None, proxy_headers_hash=None), SSLCertVerificationError(1,
 '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: CA cert does not include key usage extension (_ssl.c:1028)'))

In order to workaround this issue, the strict setting must be removed in order to establish an SSL connection to microk8s with the CA certificate, which is not ideal (there could be multiple packages, for example, kopf, kubernetes client, etc), and less secure.

Generate a CA certificate with keyUsage defined.

Is there another way to update the openssl.cnf file and its defaults? I see in:

https://github.com/canonical/microk8s/blob/master/microk8s-resources/wrappers/openssl.wrapper#L9C1-L10C1

and it refers to 
export OPENSSL_CONF="${SNAP}/etc/ssl/openssl.cnf"

however, I don't believe this is a file that users can update/modify.

ktsakalozos-canonical:
## Summary

- Add `keyUsage`, `basicConstraints`, `subjectKeyIdentifier`, and `authorityKeyIdentifier` extensions to the generated CA certificates (`ca.crt` and `front-proxy-ca.crt`)
- Fixes Python 3.13+ TLS verification failures caused by `VERIFY_X509_STRICT` requiring the `keyUsage` extension on CA certs

## Problem

Python 3.13 now enables `VERIFY_X509_STRICT` by default in its SSL context. This causes any Python-based Kubernetes client (e.g., `kopf`, `lightkube`, the official `kubernetes` package) to fail with:

```
SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: CA cert does not include key usage extension
```

The root CA certificates generated by microk8s did not include any X509v3 extensions.

## Fix

Both CA cert generation commands now include:

| Extension | Value |
|-----------|-------|
| `basicConstraints` | `critical, CA:TRUE` |
| `keyUsage` | `critical, keyCertSign, cRLSign` |
| `subjectKeyIdentifier` | `hash` |
| `authorityKeyIdentifier` | `keyid:always, issuer` |

Uses OpenSSL's `-addext` flag (available since OpenSSL 1.1.1).

## Note for existing clusters

This only affects newly generated certificates (the code skips generation if the cert file already exists). Existing clusters will need to regenerate their CA certs, e.g. via `microk8s refresh-certs --cert ca.crt`, to pick up the fix.

Fixes #4864

### DIFF
--- a/.github/actions/test-prep/action.yaml
+++ b/.github/actions/test-prep/action.yaml
@@ -11,7 +11,8 @@ runs:
       shell: bash
       run: |
         set -x
-        pip3 install -r ./tests/requirements.txt
+        sudo apt-get remove -y python3-yaml
+        sudo pip3 install -r ./tests/requirements.txt
     # Docker sets iptables rules that interfere with LXD and K8s.
     # https://documentation.ubuntu.com/lxd/en/latest/howto/network_bridge_firewalld/#prevent-connectivity-issues-with-lxd-and-docker
     - name: Apply Docker iptables workaround
--- a/.github/actions/test-prep/action.yaml
+++ b/.github/actions/test-prep/action.yaml
@@ -11,7 +11,7 @@ runs:
       shell: bash
       run: |
         set -x
-        sudo pip3 install -r ./tests/requirements.txt
+        pip3 install -r ./tests/requirements.txt
     # Docker sets iptables rules that interfere with LXD and K8s.
     # https://documentation.ubuntu.com/lxd/en/latest/howto/network_bridge_firewalld/#prevent-connectivity-issues-with-lxd-and-docker
     - name: Apply Docker iptables workaround
--- a/microk8s-resources/actions/common/utils.sh
+++ b/microk8s-resources/actions/common/utils.sh
@@ -690,12 +690,26 @@ produce_certs() {
 
     # Generate apiserver CA
     if ! [ -f ${SNAP_DATA}/certs/ca.crt ]; then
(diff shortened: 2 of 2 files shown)


ethandcosta (approved):
LGTM

Issue #5543, opened 2026-06-09
Can't deploy metallb - apparmor issue


wallyworld:
#### Summary
To my recollection this was working a few weeks ago when I last tried it.
Maybe I'm holding it wrong? Or maybe there's a legitimate issue?
It seems there's an apparmor permission problem.

NB: the exact same steps work with the classic microk8s snap.

I installed microk8s 1.36-strict and attempted to deploy metallb

` sudo microk8s kubectl apply -f "metallb.io_configurationstates.yaml"`

Where the yaml file is
```
---
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  annotations:
    controller-gen.kubebuilder.io/version: v0.19.0
  name: configurationstates.metallb.io
spec:
  group: metallb.io
  names:
    kind: ConfigurationState
    listKind: ConfigurationStateList
    plural: configurationstates
    singular: configurationstate
  scope: Namespaced
  versions:
  - additionalPrinterColumns:
    - jsonPath: .status.result
      name: Result
      type: string
    - jsonPath: .status.errorSummary
      name: ErrorSummary
      type: string
    - jsonPath: .metadata.creationTimestamp
      name: Age
      type: date
    name: v1beta1
    schema:
      openAPIV3Schema:
        description: |-
          ConfigurationState is a status-only CRD that reports configuration validation results from MetalLB components.
          Labels:
            - metallb.io/component-type: "controller" or "speaker"
            - metallb.io/node-name: node name (only for speaker)
        properties:
          apiVersion:
            description: |-
              APIVersion defines the versioned schema of this representation of an object.
              Servers should convert recognized schemas to the latest internal value, and
              may reject unrecognized values.
              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
            type: string
          kind:
            description: |-
              Kind is a string value representing the REST resource this object represents.
              Servers may infer this from the endpoint the client submits requests to.
              Cannot be updated.
              In CamelCase.
              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
            type: string
          metadata:
            type: object
          status:
            description: ConfigurationStateStatus defines the observed state of ConfigurationState.
            properties:
              conditions:
                description: Conditions contains the status conditions from the reconcilers
                  running in this component.
                items:
                  description: Condition contains details for one aspect of the current
                    state of this API Resource.
                  properties:
[...]

PR #5552, opened 2026-06-18
fix(runc): apply AppArmor profile after finalizeNamespace in standard_init (strict)

ktsakalozos-canonical:
## Summary
Reorders the strict-mode runc **v1.4.2** patch `0003-standard_init_linux-change-AppArmor-profile-as-late-` so the AppArmor profile change (and `NoNewPrivileges`) happen **after** `finalizeNamespace` runs `setuid(2)`, mirroring the ordering already used in `setns_init_linux`.

## Background
This fixes the second AppArmor denial seen when deploying workloads such as metallb in strict mode on AppArmor 4.x hosts (e.g. Plucky 25.10):

```
profile="snap.microk8s.daemon-containerd" comm="runc:[2:INIT]"
requested_mask="receive" denied_mask="receive" signal=rtmin+1
peer="cri-containerd.apparmor.d"
```

## Root cause
The strict patches switch the container's AppArmor profile **immediately** (`aa_change_profile`) instead of on exec (`aa_change_onexec`) — this is required because the kernel forbids gaining a profile on `execve` once `NO_NEW_PRIVS` is set, and Kubernetes pods with `allowPrivilegeEscalation: false` set NNP.

In `standard_init_linux.go` (container **creation**, `runc:[2:INIT]`), patch `0003` applied the profile right after `syncParentReady`, i.e. **before** `finalizeNamespace` calls `setuid(2)`. That relabels only the calling thread to `cri-containerd.apparmor.d`, while the Go runtime's sibling threads stay under `snap.microk8s.daemon-containerd`. The subsequent `setuid(2)` triggers glibc's NPTL setxid broadcast (`SIGRTMIN+1`) **across the two profiles**, which AppArmor 4.x denies → pod creation fails.

`setns_init_linux.go` (container **exec**) already does `finalizeNamespace` → `ApplyProfile` → NNP, which is why `kubectl exec` works but pod creation does not. This change makes `standard_init` follow the same proven ordering.

## Why Docker's containerd is unaffected
The Docker snap uses stock upstream runc with `aa_change_onexec`, which defers the transition to `execve`; all threads share one profile during the `setuid` handshake, so the cross-profile signal never occurs. Only microk8s's immediate-relabel patches create the split, so this is fixed here rather than in snapd's `docker-support` interface.

## Change
- `build-scripts/components/runc/strict-patches/v1.4.2/0003-...patch`: move the `ApplyProfile` + `NNP` block from after `syncParentReady` to after `finalizeNamespace`/`pdeath.Restore()`. NNP still follows the profile change (the kernel forbids switching profile once NNP is set).

Only the currently-built version (v1.4.2) is touched.

## Verification
- All three strict patches `git am` cleanly on a fresh `v1.4.2` checkout.
- `go build ./libcontainer/` succeeds; `gofmt` clean.
- Pending: strict snap build + metallb repro on an AppArmor 4.x host to confirm no `rtmin+1` denials on either `snap.microk8s.daemon-containerd` or `cri-containerd.apparmor.d`, and that `kubectl exec` still works.

## Related
Complements the `containerd-profile` signal rule change in #5550 / #5551, which is retained as defense-in-depth on the send side. Refs #5543.

### DIFF
--- a/.github/workflows/build-snap.yml
+++ b/.github/workflows/build-snap.yml
@@ -214,6 +214,10 @@ jobs:
   upload_sarifs_matrix:
     needs: security-scan
     runs-on: ubuntu-latest
+    permissions:
+      contents: read          # for actions/checkout
+      actions: read           # for download-artifact (same-run Artifacts API)
+      security-events: write  # for codeql-action/upload-sarif
     strategy:
       fail-fast: true
       matrix:
--- a/.github/workflows/build-snap.yml
+++ b/.github/workflows/build-snap.yml
@@ -215,9 +215,9 @@ jobs:
     needs: security-scan
     runs-on: ubuntu-latest
     permissions:
-      contents: read          # for actions/checkout
-      actions: read           # for download-artifact (same-run Artifacts API)
-      security-events: write  # for codeql-action/upload-sarif
+      contents: read # for actions/checkout
+      actions: read # for download-artifact (same-run Artifacts API)
+      security-events: write # for codeql-action/upload-sarif
     strategy:
       fail-fast: true
       matrix:
--- a/build-scripts/components/runc/strict-patches/v1.4.2/0003-standard_init_linux-change-AppArmor-profile-as-late-.patch
+++ b/build-scripts/components/runc/strict-patches/v1.4.2/0003-standard_init_linux-change-AppArmor-profile-as-late-.patch
@@ -1,15 +1,29 @@
-From 5007ab516cb619db9e75218f9739c555762e26da Mon Sep 17 00:00:00 2001
+From 8b06d0fcf51fbaee145b8108c83a03b6dd8cdeb6 Mon Sep 17 00:00:00 2001
 From: Alberto Mardegan <mardy@users.sourceforge.net>
-Date: Wed, 6 May 2026 14:44:12 +0200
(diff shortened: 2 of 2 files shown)


louiseschmidtgen (approved):
LGTM

Context around the approval:

The problem: With our strict-snap patches switching the AppArmor profile immediately (aa_change_profile) instead of on-exec, calling ApplyProfile before setuid splits the Go runtime's sibling threads across two profiles. The kernel only changes the profile on the current thread ([domain.c#L1433](https://github.com/torvalds/linux/blob/master/security/apparmor/domain.c#L1433) — it doesn't broadcast to siblings. So when glibc's NPTL does the setuid(2) setxid broadcast (SIGRTMIN+1) across all threads, that signal crosses two profiles and AppArmor 4.x denies it → pod creation fails (e.g. metallb with allowPrivilegeEscalation: false).

The fix: Do the setuid broadcast first, while all threads still share one profile, then apply the locked-down profile + NNP to the current thread. The post-switch split is harmless because we have nothing left to broadcast to the siblings.

Why we can't just match upstream: Upstream defers the switch to exec, so the ordering never bites them. In a strict snap we need the profile active earlier than that, so this reordering is the best we can do.

PR #5557, opened 2026-06-26
fix(runc): apply AppArmor profile after finalizeNamespace in standard_init

ktsakalozos-canonical:
## Summary

Master counterpart of the runc init-order fix being landed on `strict`. Cherry-picks the same three commits onto `master` (which also builds runc v1.4.2).

- **fix(runc): apply AppArmor profile after finalizeNamespace in standard_init** — Reorders the strict-mode runc v1.4.2 patch so the AppArmor profile change (and NoNewPrivileges) happen after `finalizeNamespace` runs `setuid(2)`, mirroring `setns_init_linux`. Because the strict patches switch the profile immediately (`aa_change_profile`) instead of on exec, relabelling before `setuid` left the Go runtime's sibling threads under the old profile; glibc's NPTL setxid broadcast (SIGRTMIN+1) then crossed two AppArmor profiles and was denied by AppArmor 4.x, breaking pod creation for workloads with `allowPrivilegeEscalation: false` (e.g. metallb).
- **ci: grant security-events write permission for SARIF upload** — Adds a job-scoped `permissions` block to `upload_sarifs_matrix` so `codeql-action/upload-sarif` can upload (fixes "Resource not accessible by integration").
- **ci: normalize permission comment formatting in build-snap workflow** — Whitespace-only cleanup of the comments added above.

## Notes
- The runc patch file (`strict-patches/v1.4.2/0003-...`) is byte-identical to the version landing on `strict`.
- A `1.35-strict` counterpart is **not** included here: that branch builds runc **v1.3.3** (no `v1.4.2` patch dir), so the fix must be hand-ported to `strict-patches/v1.3.3/0003-...` separately.

### DIFF
--- a/.github/workflows/build-snap.yml
+++ b/.github/workflows/build-snap.yml
@@ -210,9 +210,9 @@ jobs:
     needs: security-scan
     runs-on: ubuntu-latest
     permissions:
-      contents: read          # for actions/checkout
-      actions: read           # for download-artifact (same-run Artifacts API)
-      security-events: write  # for codeql-action/upload-sarif
+      contents: read # for actions/checkout
+      actions: read # for download-artifact (same-run Artifacts API)
+      security-events: write # for codeql-action/upload-sarif
     strategy:
       fail-fast: true
       matrix:
--- a/.github/workflows/build-snap.yml
+++ b/.github/workflows/build-snap.yml
@@ -209,6 +209,10 @@ jobs:
   upload_sarifs_matrix:
     needs: security-scan
     runs-on: ubuntu-latest
+    permissions:
+      contents: read          # for actions/checkout
+      actions: read           # for download-artifact (same-run Artifacts API)
+      security-events: write  # for codeql-action/upload-sarif
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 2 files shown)


louiseschmidtgen (approved):
LGTM

Context around the approval:

The problem: With our strict-snap patches switching the AppArmor profile immediately (aa_change_profile) instead of on-exec, calling ApplyProfile before setuid splits the Go runtime's sibling threads across two profiles. The kernel only changes the profile on the current thread ([domain.c#L1433](https://github.com/torvalds/linux/blob/master/security/apparmor/domain.c#L1433) — it doesn't broadcast to siblings. So when glibc's NPTL does the setuid(2) setxid broadcast (SIGRTMIN+1) across all threads, that signal crosses two profiles and AppArmor 4.x denies it → pod creation fails (e.g. metallb with allowPrivilegeEscalation: false).

The fix: Do the setuid broadcast first, while all threads still share one profile, then apply the locked-down profile + NNP to the current thread. The post-switch split is harmless because we have nothing left to broadcast to the siblings.

Why we can't just match upstream: Upstream defers the switch to exec, so the ordering never bites them. In a strict snap we need the profile active earlier than that, so this reordering is the best we can do.

